Issue #286, opened 2016-02-23
Executor Refactor


mpeck:
The Great Executor Refactor
======================
The executor is in need of a refactor. Currently everything is pretty much crammed into a single module which makes it difficult to maintain. In order to improve maintainability and open up the possibility for future improvements we have some suggestions for a refactor.

Proposal - move templating out of the executor
----------------------------------------------------------------
Templating should be done independently of the executor. Specifically we are thinking about moving templating up to the initializer level. The initializer is a named process that starts executor processes. If we move templating up to this level, and perhaps rename the initializer to something more appropriate(pipeline_coordinator?), it will not only make things a bit cleaner, it gives us a place to more clearly implement things like error templating. It also makes it easier to eventually introduce parallel command execution and recursive alias expansion, because the executor can simple return json.

Proposal - move redirects out of the executor
------------------------------------------------------------
In the same vein as the templating move. By moving redirects out of the we create a further separation of responsibilities for command execution.

Proposal - break the executor into two parts
-------------------------------------------------------------
The real meat of the executor exists in the `prepare_or_finish` function. Over time the function has grown and become difficult to follow. A major issue revolves around the fact that we iterate over multiple lists. At the top level we have a list of invocations, the pipeline. We iterate over the list of invocations executing each command in turn. Dependent on the type of command, they can be executed once or multiple times. And we further iterate over command input and output lists based on that execution type.

We can simplify this by breaking the executor into two distinct components: The pipeline executor (iterates over the invocation list), and the command executor (managers how commands are executed). The pipeline executor's main purpose is to maintain the top level list of invocations, iterate over them and spin up a new command executor to run each invocation. The command executor is responsible for identifying the type of command to be executed and properly executing said command. 

This would allow us to do some interesting things with command execution, namely things like parallel execution. It greatly simplifies the whole input output buffer bit. And it also allows us to do actual recursive expansion of aliases. Currently we just expand the alias and prepend the resulting pipeline to the invocation list. Using this method we can just spin up another pipeline executor as a child of the command executor. When the pipeline executor returns json, we just continue processing as normal. We can easily track how deep we go down the alias expansion rabbit hole and add configurable limits. Which should protect us from things like cyclic references.

I anticipate this issue spawning multiple additional issues. Also any discussion or feedback is welcome.

kevsmith:
There are a couple of things that would help me consider these changes:

1. Description of the desired end state wrt responsibilities of each component
[...]

PR #290, opened 2016-02-23
Command aliases

### LINKED ISSUE #260: Command alias executor wiring
References #200 

The executor should understand how to deal with command aliases. Command aliases should be expanded lazily as the pipeline is processed. To prevent infinite recursion, ie when an alias references itself directly or indirectly, a recursion depth limit will be configurable.

mpeck:
This enables command aliases.

resolves #260 
dependent on https://github.com/operable/piper/pull/8
dependent on https://github.com/operable/piper/pull/9


lib/cog/command/command_resolver.ex
```diff
@@ -0,0 +1,56 @@
...
+  alias Cog.Repo
+  alias Cog.Queries.Command
+  alias Cog.Queries.Alias
+  alias Piper.Command.SemanticError
+
+  def find_bundle(name) when is_binary(name) do
+    if String.contains?(name, ":") do
+      :identity
+    else
+      case Repo.all(Command.bundle_for(name)) do
```
kevsmith: Since bundle names trump alias names do we prevent users from creating aliases which conflict with installed bundles? What happens when a user creates an alias and later a bundle with a conflicting name is installed?

lib/cog/command/command_resolver.ex
```diff
@@ -0,0 +1,56 @@
...
+        bundle_names ->
+          SemanticError.new(name, {:ambiguous_command, bundle_names})
+      end
+    end
+  end
+  def find_bundle(name) do
+    SemanticError.new("#{inspect name}", :no_command)
+  end
+
+  defp is_alias?("user:" <> name) do
```
kevsmith: I think

`defp get_alias_type(String.t()) :: "site" | "user" | nil`

is a bit easier to understand.

lib/cog/command/command_resolver.ex
christophermaier: Maybe we do something like when an unqualified command name is not authoritative (i.e., "You typed `foo`, but we don't know if you meant `foo:foo` or `bar:foo`")

Perhaps we need a way to unambiguously note that something is an alias? (`alias:foo`, though that prevents there ever being a bundle named `alias`...)

kevsmith: If memory serves, returning a `SemanticError` from the old `BundleResolver` with a reason of `:ambiguous_command` would generate such an error message. Based on what I've read so far I believe we preserve the behavior.

As a reasonably helpful and not entirely difficult first step we could send a warning message to the user whenever they execute an alias which shadows a bundle.

kevsmith: If memory serves, returning a `SemanticError` from the old `BundleResolver` with a reason of `:ambiguous_command` would generate such an error message. Based on what I've read so far I believe we preserve the behavior.

As a reasonably helpful and not entirely difficult first step we could send a warning message to the user whenever they execute an alias which shadows a bundle.

lib/cog/command/pipeline/executor.ex
```diff
@@ -215,7 +232,38 @@ defmodule Cog.Command.Pipeline.Executor do
...
+    do: expand_alias(current_bound, state)
+  def maybe_expand(:timeout, state),
+    do: {:next_state, :get_options, state, 0}
+
+  defp expand_alias(current_bound, state) do
+    case AliasExpander.expand(current_bound.command) do
+      {:ok, %Ast.Pipeline{invocations: invocations}} ->
+        cond do
+          state.current_depth + 1 > state.max_depth ->
+            Helpers.send_reply("Error expanding alias. I'm trying to can... but I just can't.", state.request, state.mq_conn)
```
kevsmith: Can we have a friendly but less joke-y error message here?

lib/cog/command/pipeline/executor.ex
```diff
@@ -215,7 +232,38 @@ defmodule Cog.Command.Pipeline.Executor do
...
+            fail_pipeline(state, :alias_expansion_error, "Error expanding alias. Too much recursion.")
+          true ->
+            [new_current|rest] = invocations
+            # We insert a marker here to determine how deep we have gone with expansion
+            rest = rest ++ [%Marker{alias_name: current_bound.command}]
```
kevsmith: I'm having difficulty following the logic here. Why do we need the marker on the `remaining` stack?

lib/cog/command/pipeline_parser.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Cog.Command.PipelineParser do
+  alias Piper.Command.Parser
```
kevsmith: Might it be cleaner to add a `return_pipelines` option to `Piper.Command.Parser.parse/2`?

mix.lock
```diff
@@ -41,7 +41,7 @@
   "phoenix_html": {:hex, :phoenix_html, "2.5.0"},
   "phoenix_live_reload": {:hex, :phoenix_live_reload, "1.0.3"},
   "piper": {:git, "https://github.com/operable/piper.git", "a3f4a069c88c0241a3dcaf94b4495fbc5fddaaa2", [tag: "0.2"]},
-  "plug": {:hex, :plug, "1.1.1"},
+  "plug": {:hex, :plug, "1.1.2"},
```
kevsmith: ?

lib/cog/command/pipeline/executor.ex
mpeck: oops, I forgot to change that.

lib/cog/command/pipeline/executor.ex
mpeck: I add a marker to determine how deep we have gone with alias expansion. When we expand an alias we add the marker to the stack before the new invocation list and increase the current_depth by 1. When we come across the marker later during execution, we drop it and decrease the current_depth by 1. If our current_depth ever exceeds our max_depth we shutdown and return an error to the user.

mix.lock
mpeck: I think this happened when I rebased and ran `mix deps.get`.

lib/cog/command/pipeline/executor.ex
kevsmith: iiuc, isn't it possible to arrive at the wrong count given the way we're embedding the marker in `remaining`? I don't have a concrete example yet as I'm still trying to wrap my head around the flow of execution.

kevsmith: Is the marker used to count the number of expansion passes required for a single alias instance? Or is it meant to help detect alias cycles?

mpeck: It's meant to detect alias cycles. It's a pretty naive check, but I thought it would prevent users from accidentally or maliciously creating a cyclic reference.

mpeck: I don't think we will arrive at an incorrect count though, unless I'm missing something. We add a marker for each alias expansion and increase the count, the marker is removed in `prepare_or_finish` and we decrease the count.

mpeck: Another note: I implemented this before we had our discussion on refactoring the executor to an interpreter so I was still imagining us expanding aliases and executing them as tiny pipelines. I was attempting to match the behavior there. I thought this behavior was still relevant though. Since it solves the issue of infinite expansion of aliases regardless of the position of the alias in the command string. Especially since we haven't implemented pipeline timeouts yet.

kevsmith: 1. As I think I've said previously we can (and should) defer cycle detection until the executor has been refactored per #286. For now, a warning in Cog's docs about cyclical alias references should be sufficient.
1. Limiting the number of expansion passes used to expand a single alias is orthogonal to detecting alias reference cycles. The conceptual separation should be reflected in the code as well. If I'm reading the code correctly -- and I might not be -- it appears that we're co-mingling the two concepts via the marker.

Suggestions:

1. Remove alias cycle detection logic.
1. Add logic to `Cog.Command.AliasExpander` to discover when an alias expands to a pipeline with a command in the first position. It should accumulate intermediate pipelines, if any, it finds along the way while being careful to preserve pipeline order.
1. Once the expander can identify a successful alias expansion we can add a counter and limit the number of times it loops.

lib/cog/command/command_resolver.ex
mpeck: What's the best path to accomplish that? The only thing I can think of is querying for aliases on every command invocation. Is there a more efficient way?

lib/cog/command/pipeline/executor.ex
```diff
@@ -153,12 +158,11 @@ defmodule Cog.Command.Pipeline.Executor do
   to `remaining` in the state.
   """
   def parse(:timeout, state) do
-    case Parser.scan_and_parse(state.request["text"], command_resolver: &BundleResolver.find_bundle/1) do
-      {:ok, %Ast.Invocation{}=invocation} ->
-        {:next_state, :lookup_redirects, %{state | pipeline: %Ast.Pipeline{invocations: [invocation]}}, 0}
+    case Parser.scan_and_parse(state.request["text"], command_resolver: &CommandResolver.find_bundle/1, return_pipeline: true) do
       {:ok, %Ast.Pipeline{}=pipeline} ->
+        maybe_warn_on_ambiguous_aliases(pipeline, state)
```
mpeck: @kevsmith @christophermaier Any thoughts on this? If a command is executed with an ambiguous alias we send a warning to the user. It's a bit odd because we query for aliases on every command invocation, and it won't trigger if the user passes the fully qualified name for the alias. implementation is on L#809 and in the `Lib.Command.AmbiguousAliasChecker` module.

lib/cog/command/pipeline/executor.ex
```diff
@@ -785,4 +806,15 @@ defmodule Cog.Command.Pipeline.Executor do
     |> Enum.sort
   end
 
+  defp maybe_warn_on_ambiguous_aliases(%Ast.Pipeline{invocations: invocations}, state) do
+    Enum.map(invocations, fn(invocation) ->
+      case AmbiguousAliasChecker.ambiguous_command?(invocation) do
+        {true, alias_name} ->
+          Helpers.send_reply("Warning: The alias '#{alias_name}' and the command '#{invocation.command}' are ambiguous. Ignoring '#{alias_name}' and executing '#{invocation.command}'. Consider renaming your alias.", state.request, state.mq_conn)
```
christophermaier: You'd need to fail the pipeline here, correct?

mpeck: I don't know. Should we fail or just tell the user that the behavior may not be what they expect?

christophermaier: We fail with a `SemanticError` if an unqualified-command name is given, which is the same kind of ambiguity. We should always fail in such a case.

mpeck: ok, I think that would actually make this more efficient. I can fail in the CommandResolver and here.

lib/cog/command/ambiguous_alias_checker.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Cog.Command.AmbiguousAliasChecker do
+  alias Cog.Repo
+  alias Cog.Queries.Alias
+  alias Cog.Models.Command
+  alias Piper.Command.Ast
+
+  def ambiguous_command?(%Ast.Invocation{command: "site:" <> _}),
+    do: false
+  def ambiguous_command?(%Ast.Invocation{command: "user:" <> _}),
```
christophermaier: If we are treating `site` and `user` effectively as "bundles" for the purposes of alias and command resolution, then we should prevent "real bundles" named `site` and `user` from ever being created.

(Come to think of it, we should have already done that with `site` since we have a permission namespace with that name.)

christophermaier: You could do this by adding a custom validation function to the flow in `Cog.Models.Bundle.changeset/2`, similar to what I did to enforce the fact that the `operable` bundle should always be enabled, for example.

mpeck: I like that. Could that warrant another PR though? That seems like that belongs in alias management rather than execution.

lib/cog/command/ambiguous_alias_checker.ex
```diff
@@ -0,0 +1,26 @@
...
+    do: false
+  def ambiguous_command?(%Ast.Invocation{command: command_name}) do
+    {_bundle, command_name} = Command.split_name(command_name)
+
+    case Repo.one(Alias.user_alias_by_name(command_name)) do
```
christophermaier: `Repo.one` doesn't return an empty list, it returns `nil`

kevsmith:
Assuming current build is green I'm :+1: to merge.

PR #294, opened 2016-02-24
Fix Cog's Dockerfile so the container can be built

imbriaco:
* Updates dependencies in `mix.exs` to make sure they all use HTTPS instead of SSH to fetch.
* Uses operable/docker-base, simplifying the Dockerfile considerably

This solves the same problem as #284, but we'll take this iteration since it uses the operable/docker-base that wasn't publicly available when @sgerrand investigated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,48 +1,16 @@
-FROM debian:jessie
+FROM operable/docker-base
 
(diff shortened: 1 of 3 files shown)


sgerrand:
:gift_heart:

PR #295, opened 2016-02-24
Enforce unique bundle names in database

christophermaier:
Not having this was a complete oversight!

Changed some tests to both use the Bundle changeset, and not conflict
with the already-installed `operable` bundle.


kevsmith:
D'oh! :+1:

PR #296, opened 2016-02-24
Commands no longer have versions

christophermaier:
Bundles will have versions soon, but commands are losing them now. There
should be no user-facing effects on this change, as we weren't really
using command versions anyway; they're vestigial from an earlier design.

See also https://github.com/operable/spanner/pull/41, but that is actually independent of this change, and not necessary for this to be merged.

### DIFF
--- a/lib/cog/bundle/install.ex
+++ b/lib/cog/bundle/install.ex
@@ -15,7 +15,7 @@ defmodule Cog.Bundle.Install do
 
   require Logger
 
-  @command_attrs ["name", "version", "options", "enforcing",
+  @command_attrs ["name", "options", "enforcing",
                   "calling_convention", "execution",
                   "documentation"]
 
@@ -82,7 +82,6 @@ defmodule Cog.Bundle.Install do
   #                              "enforcing" => true,
   #                              "calling_convention" => "bound",
   #                              "execution" => "multiple",
-  #                              "version" => "0.0.1",
   #                              "options": [%{"name" => "instance-id", "type" => "string", "required" => true}],
   #                              "module" => "Cog.Commands.EC2Tag"})
   defp create_command(%Bundle{}=bundle, command_spec) do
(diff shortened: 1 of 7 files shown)


lib/cog/bundle/install.ex
```diff
@@ -15,7 +15,7 @@ defmodule Cog.Bundle.Install do
 
   require Logger
 
-  @command_attrs ["name", "version", "options", "enforcing",
+  @command_attrs ["name", "options", "enforcing",
```
kevsmith: Idle curiosity: I wonder if there's a way to get these off the command model directly....

christophermaier: Actually, I think there is... lemme go poke around a bit.

kevsmith:
:+1:

Does `scripts/setup.sh` or our startup scripts run migrations on existing installs? /cc @imbriaco

lib/cog/bundle/install.ex
christophermaier:
```elixir
iex(1)> Cog.Models.Command.__schema__(:fields)
[:id, :name, :documentation, :enforcing, :calling_convention, :execution, :bundle_id]
iex(2)> Cog.Models.Command.__schema__(:associations)
[:bundle, :rules, :options]
```
We'd need to filter out the two IDs from the fields, and know which of the associations we care about. Seems like we'd need some additional annotations in order to do this more generally. Maybe we can leave this as is for now; if there are other instances where this pattern pops up we could reconsider then.

imbriaco:
None of our current scripts run migrations, except for the AMI builder. The work I've been doing with docker-compose _does_ automatically run migrations, but I'm generally afraid of doing that in an automated way. 

I do have some ideas that I'll put into another issue about how we might handle this generally, but in the mean time we should remember to put this in release notes when we cut a new release and maybe call out in the commit message that a migration will be required.

kevsmith:
@christophermaier call out the migration requirement either in commit message here or on the parent issue. Once that's done I'm :+1: on merging.

christophermaier:
@kevsmith I'll amend the commit message and then merge; thanks!

christophermaier:
@imbriaco Also, when cutting a new release, we can take a look at the files that have changed since the last release; if anything is in `priv/repo/migrations`, we know a migration is needed.

Issue #255, opened 2016-02-17
Extract a "pluck" command


christophermaier:
Currently, the `filter` command performs both filtering and projection / "plucking". These should be separate commands.

The new "pluck" command (think Rails and ActiveRecord [see, e.g., http://gavinmiller.io/2013/getting-to-know-pluck-and-select/]) effectively performs the `--return` option functionality from `filter`. That is, `pluck foo` would return inputs with all but their `foo` keys removed (e.g., `{"foo": "blah"}`), `pluck foo bar` would return just the `foo` and `bar` keys (e.g., `{"foo": "blah", "bar": "blah"}`), and so on. (Could instead call it like `pluck --fields=foo,bar`, perhaps)

`pluck` should be a non-enforcing, multiple execution function with "all" binding semantics.

Removing the `--return` functionality from the `filter` command should be part of this work, as well.

lhaskins:
In a conversation about filter, there was some discussion about how this command should be executed. It should essentially perform the same action that the usage of the `return` option would perform in the previous filter command before the completion of Issue #256.

In the same way that filter has the capability to traverse a path, pluck should be able to do the same. Initially, the first pass was to do the equivalent of return adding in the ability to handle JSON paths. 

It should be noted that the fields must be escaped when handling paths that are quoted. For example, in order to access the following: `-fields=foo,"bar.baz"` the quotes must be escaped to look like the following: `-fields="foo,\"bar.baz\""`.

lhaskins:
Given the following command: ```seed '[{"foo":{"bar":{"baz":"stuff"}, "one": 1}, "qux": "one"}, {"foo": {"bar":{"baz":"me"}}, "qux": "two"}]' | pluck --paths="qux,'foo.bar.baz'"```

Should the output be collapsed such that it would look like:
```
[{
  "qux": "one",
  "foo.bar.baz": "stuff"
},
{
  "qux": "two",
  "foo.bar.baz": "me"
}]
```
or
```
{
  "qux": "one",
  "foo": {
    "one": 1,
    "bar": {
      "baz": "stuff"
    }
  }
},
{
  "qux": "two",
  "foo": {
    "bar": {
      "baz": "me"
    }
  }
}
```
Or a third option where there is a ```collapse``` flag that collapses the output into a single level with the default set to display the expanded map

lhaskins:
Closing since changes in piper/executor may make this command unnecessary.

Issue #251, opened 2016-02-16
Investigate Piper's typing of options


christophermaier:
It appears that Piper isn't enforcing the declared types of option values, causing commands to manage this on their own.

If we could be certain that the data we receive in a command was the proper type, we could simplify a lot of the commands.

(If we could do the same with arguments, that would be useful as well, but would require changes to lots of other places as well, and would be much larger in scope.)

imbriaco:
When we fix this we should make sure that it's valid to send integer and boolean values to options that specify that they take strings, rather than forcing users to explicitly quote them. Since we know that integers and booleans are both safe to cast to strings we can do so before we create the CommandRequest.

mpeck:
dependent on #340

Issue #248, opened 2016-02-16
Create minimal test harness for commands


christophermaier:
Ideally, something that we can execute independently of the executor. This would make testing return values a lot easier, at the very least, as they wouldn't be filtered through the executor's various transformations.

PR #298, opened 2016-02-25
Filter command revamp

### LINKED ISSUE #256: Filter command should take a "--path" option instead of "--field"
We currently do not deal with nested keys yet, so the addition of a `--path` option is needed. This would allow specifying a key path to nested objects with a dot notation, e.g., `--path=foo.bar.baz`, which could be used to target `"hello world"` in the object `{"foo": {"bar": {"baz": "hello world"}}}`, for instance.

It should also handle array indexing as well, e.g. `--path=foo.bar[0]` would target `"hello"` in this object:

```json
{
  "foo": {
      "bar": ["hello", "world"]
  }
}
```

Since dots are legal in JSON field names, we should also allow this by quoting such fields, so `--path=foo."bar.baz".quux` would target "hello" in this object:

```json
  {
      "foo": {
          "bar.baz": {
              "quux": "hello"
          }
      }
  }
```
(This may or may not require some tweaking of Piper, though; that needs investigation)

This would subsume the current `--field` option (a field is just a path of length 1, after all), so we'll need to remove the `--field` option.

Also, `filter` currently accepts a `--matches` option without a corresponding `--field` option. This should be disallowed; `--matches` only makes sense with `--field` (well, with `--path` once this feature is complete)

Just like `--field`, `--path` can continue to be specified alone; in this case, we let the input object through if it has data at the specified path.

Specifying no options (as in, e.g., `something | filter | something-else`) is allowed, but is effectively a no-op)

lhaskins:
The filter command should accept a path instead of just a field. The path should be separated by `.` while allowing for sections of the path to be quoted. It should be noted that the path needs to be in quotes with inner quotes being escaped.

The filter command works like the unix find command. where the path specifies the key in the JSON data and the matches option denotes the value that the user is looking for.

It is now considered an error to use this command specifying a `matches` option without a `path` option. The user needs to state the key that one is to search when looking for the matched value.

There were some additional tests added around the filter command with paths and matches

Fixes #256


test/integration/command_test.exs
```diff
@@ -110,7 +110,22 @@ defmodule Integration.CommandTest do
...
 
   test "running a command in a pipeline with nil output", %{user: user} do
-    response = send_message(user, ~s(@bot: seed '[{"a": "1", "b": "2"}, {"a": "3"}]' | filter --field="b" | echo $a))
+    response = send_message(user, ~s(@bot: seed '[{"a": "1", "b": "2"}, {"a": "3"}]' | filter --path=b | echo $a))
     assert response["data"]["response"] == "1"
   end
+
+  test "reading the path to filter a certain path", %{user: user} do
+    response = send_message(user, ~s(@bot: seed '[{"foo":{"bar":{"baz":"stuff"}}}, {"foo": {"bar":{"baz":"me"}}}]' | operable:filter --path="foo.bar"))
+    assert response["data"]["response"] == "{\n  \"bar\": {\n    \"baz\": \"stuff\"\n  }\n}\n{\n  \"bar\": {\n    \"baz\": \"me\"\n  }\n}"
```
christophermaier: This isn't the correct behavior. `filter --path=...` does not _extract_ the value found at the end of the path, it allows the whole object through. If the object has nothing at that path, it does not make it through the filter.

christophermaier: Also, some code I was experimenting with on an unpushed branch might come in useful for dealing with the JSON strings we get back:

```elixir
  #
  # Example:
  #
  # lookbehind.
  #
  # Now you have `n` problems.
```

Might be worthwhile to add that into a test helper module and take advantage of it here, at least until we change how we test commands in #248

lib/cog/commands/filter.ex
```diff
@@ -26,29 +35,55 @@ defmodule Cog.Commands.Filter do
...
+  defp maybe_filter(item, %{"path" => path}) do
+    build_path(String.split(path, "."), [], [])
+    |> fetch(item)
+  end
+  defp maybe_filter(_item, %{"matches" => _matches}),
+    do: %{error: "Please include a `path` to match with"}
+  defp maybe_filter(item, _),
+    do: item
+
+  defp fetch([], item), do: item
```
christophermaier: You can simplify a lot (all?) of this by using `Kernel.get_in/2`

lib/cog/commands/filter.ex
```diff
@@ -57,6 +92,29 @@ defmodule Cog.Commands.Filter do
   defp maybe_pluck(item, _),
     do: item
 
+  defp build_path([], [], path), do: path
+  defp build_path([key|remaining], acc, path) do
```
christophermaier: Any way to split the path using a regular expression instead? This code is a bit opaque.

lib/cog/commands/filter.ex
```diff
@@ -26,29 +35,55 @@ defmodule Cog.Commands.Filter do
...
+
+  defp fetch([], item, matches) do
+    match_str = Regex.compile!(matches)
+    match = Regex.match?(match_str, to_string(item))
+    fetch_match(match, item)
+  end
+  defp fetch(_, nil, _), do: nil
+  defp fetch([key|path], item, matches) do
+    result = with {:ok, value} <- Access.fetch(item, key),
+      regex = compile_regex(matches),
```
christophermaier: What happens if the regular expression given with `--matches` doesn't actually compile?

test/integration/command_test.exs
lhaskins: So to try this again to make sure that we are on the same page: 
Input:
```seed '[{"foo":{"bar":{"baz":"stuff"}}}, {"foo": {"bar":{"baz":"me"}}}]' | operable:filter --path="foo.bar"```
Expected output:
```[{"foo":{"bar":{"baz":"stuff"}}}, {"foo": {"bar":{"baz":"me"}}}]```

Input:
```seed '[{"foo":{"bar":{"baz":"stuff"}}}, {"foo": {"bar":{"baz":"me"}}}]' | operable:filter --path="foo.bar" --matches="me"```
Expected Output:
```{"foo": {"bar":{"baz":"me"}}}```

Input:
``` seed '[{"foo":{"bar.qux":{"baz":"stuff"}}}, {"foo": {"bar":{"baz":"me"}}}]' | operable:filter --path="foo.\"bar.qux\".baz" ```
Expected Output:
``` {"foo":{"bar.qux":{"baz":"stuff"}}} ```

Comment (lhaskins) [inline test/integration/command_test.exs]: Yeah, that simplifies this even more. Thanks!

File: lib/cog/commands/filter.ex [hunk 5]
```diff
@@ -26,29 +35,55 @@ defmodule Cog.Commands.Filter do
...
+        fetch_path(true, path, item, key)
       :error ->
         nil
     end
+    fetch(path, result)
   end
-  defp maybe_filter(item, _),
-    do: item
+
+  defp fetch([], item, matches) do
```
christophermaier: I'm finding it very difficult to trace through this section of the code; I think it's trying to do too much at once, and ends up obscuring things.

Conceptually, at least, we want to see if anything exists at the specified path, and if it does (and we've been given a regular expression) see if the value matches. The fact that we're treating values as strings for the purpose of matching might warrant a dedicated function in order to highlight this fact, and to make it easier to modify the behavior if / when we decide to provide data-type-specific matching (e.g., specifying numerical ranges or cutoffs, for example, which are challenging to do correctly with regular expressions).

test/integration/command_test.exs
christophermaier: Yes, the input / output pairs you specified just now are correct. 

Conceptually the `filter` command is just deciding whether or not to allow a given input map through to the rest of the pipeline; it should not transform any individual input map in any way (and this is why the `--return` option is being removed in favor of a `pluck` command in #255)

lib/cog/commands/filter.ex
lhaskins: The command returns with ```%Regex.CompileError{message: "regex could not be compiled"}```

I can change this to return :error and return a nice error message to the user.

christophermaier: I recommend checking out the implementation of the `rules` and `permissions` command for a useful processing pattern that makes managing errors easier.

lhaskins: So I've added in some `translate_error` functionality, but in the specific case of string compilation errors for this command, I have not been able to force the error (except by explicit code changes). 

I have been able to generate an error when entering a non-string value for matches. As a result, I return an error if it is deemed that the `matches` option is not a string. 

From my understanding, the issue #251 will fix this sort of problem for all commands.

lib/cog/commands/filter.ex
lhaskins: I cleaned this up some. Let me know if it's not enough.

lib/cog/commands/filter.ex
christophermaier: It's not just about translating errors into strings, or even just about catching bad input types (though that is part of it). The `rules` and `permissions` commands are implemented in such a way that there is a single flow of execution through the command regardless of the specifics of the invocations. It's perhaps a bit verbose, but it makes it very easy to trace what happens; I've found it very useful for ensuring that all the various error scenarios are caught (and don't cause crashes, e.g.)

lhaskins: @christophermaier: Rewrote the command using the validate |> execute |> format flow. It's definitely easier to follow and read through.

imbriaco:
It looks like this removes the `--returns` option. Does that mean that filter will always return the entire object that contains a match now?

lhaskins:
Yes, the idea is to use the `pluck` command listed in https://github.com/operable/cog/issues/255 to perform the "return" portion of what this command used to do. The filter command will only return what matches the specified path and match.

PR #299, opened 2016-02-25
Adapter refactor

vanstee:
This PR does a few things to clean up the adapters:

* All functions used by modules external to adapters themselves now use functions defined in the adapter behaviour.
* All emqtt communication is handled by the adapter process which is now a GenServer like module that can be `use`d by any adapter implementation.
* Some of the adapter API was changed to be more versatile for different chat services (for instance the `send_message` function takes an entire room map rather than just an id to support direct messages in HipChat)


lib/cog.ex
```diff
@@ -32,7 +32,8 @@ defmodule Cog do
      worker(Cog.TemplateCache, []),
      worker(Carrier.CredentialManager, []),
      supervisor(Cog.Relay.RelaySup, []),
-     supervisor(Cog.Command.CommandSup, [])] ++ adapter.describe_tree()
+     supervisor(Cog.Command.CommandSup, []),
+     supervisor(Module.concat(adapter, "Supervisor"), [])]
```
kevsmith: I think it might be helpful to add some code to verify the supervisor module exists and log an error message if we can't find it. This won't be the case for the HipChat and Slack adapters but I could see this happening with a user-provided adapter. Something like `Code.ensure_loaded/1` should do the trick.

vanstee: This is a good idea. On it.

kevsmith:
Modulo above comment :+1: to merge

PR #310, opened 2016-02-26
Database is sole source of truth for bundle status 

christophermaier:
Previously, `Cog.Relay.Relays` was tracking not only which relays were
running a given bundle, but also the enabled/disabled status of those
bundles. Now, it tracks only which relays run which bundles, and knows
nothing of the bundles' status.

Instead, the executor now factors in the bundle status when
determining which relay to dispatch to. If the bundle is disabled, we
can now return a more accurate error message to the user (previously,
it was confusing to say that no relays were running a bundle if some
relays actually _were_, but the bundle was just disabled).

To ease the database burden for these status checks, a new bundle
cache is introduced. It is similar to other caches we have, but
implemented to use generic functions as much as possible. Hopefully
this can be the basis of a future generic cache behaviour that we can
use elsewhere.

### DIFF
--- a/lib/cog/command/bundle_cache.ex
+++ b/lib/cog/command/bundle_cache.ex
@@ -38,6 +38,14 @@ defmodule Cog.Command.BundleCache do
     end
   end
 
+  @doc """
+  Removes any entry for `bundle_name` in the cache. The next time
+  information for the bundle is requested, it will be retrieved fresh
+  from the database.
+  """
+  def purge(bundle_name),
+    do: GenServer.call(__MODULE__, {:purge, bundle_name})
+
   def init(_) do
     new_ets_cache_table(@ets_table)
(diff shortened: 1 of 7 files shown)


lib/cog/command/bundle_cache.ex
```diff
@@ -0,0 +1,163 @@
...
+  end
+  def handle_call({:purge, bundle_name}, _caller, state) do
+    purge(state.table, bundle_name)
+    {:reply, :ok, state}
+  end
+
+  defp bool_to_status(true),  do: :enabled
+  defp bool_to_status(false), do: :disabled
+
+  ########################################################################
```
mpeck: :+1: on standardizing the caches.

mpeck:
:+1: looks reasonable to me

PR #312, opened 2016-02-26
Enable MQTT over SSL

kevsmith:
This commit simplifies and unifies Cog's message bus configuration. The
tuple forest which is emqttd's config has been banished to priv/config/[common,plain,ssl]_mqtt.exs. In its place are the new :message_bus config stanzas on lines 12 and 18 of config/config.exs.

A new module, Cog.BusDriver, has been added to manage loading emqttd's config, starting and stopping emqttd, and reporting any errors which might occur. Doing it this way allows us to hide the complexity of configuring emqttd from the user without sacrificing configurability since any changes made to the private config files in priv/config will take effect when Cog is restarted.

emqttd has also been moved from an application dependency to an included application in mix.exs. Doing so indicates Cog is responsible for starting emqttd to OTP and allows us to intercept and rewrite its configuration as stated above.

Also, static emqttd config has been removed from config/test.exs. We were using it to control emqttd's listening port which can be more easily and obviously done via the `COG_MQTT_PORT` env var.


lib/cog/bus_driver.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Cog.BusDriver do
+
+  require Logger
+  use GenServer
+
```
mpeck: Some comments explaining what bus driver does might be helpful.

mpeck:
:+1:

lib/cog/bus_driver.ex
```diff
@@ -0,0 +1,62 @@
...
+    bindings = [mqtt_addr: mqtt_host, mqtt_port: mqtt_port]
+    case load_private_config("common_mqtt") do
+      {:ok, _} ->
+        if cert_file != nil and key_file != nil do
+          Logger.info("Message bus configured for SSL")
+          load_private_config("ssl_mqtt", [cert: String.to_char_list(cert_file),
+                                           key: String.to_char_list(key_file),
+                                           mqtt_type: :mqtts] ++ bindings)
+        else
+          Logger.info("Message bus configured for plain TCP")
```
christophermaier: I wonder if this should be a `warn` instead?

PR #315, opened 2016-02-26
Properly kill relay after failed tests

vanstee:
We now kill relay from an `on_exit` hook so processes don't stick around in-between test runs.

### DIFF
--- a/test/integration/relay_test.exs
+++ b/test/integration/relay_test.exs
@@ -15,7 +15,7 @@ defmodule Integration.RelayTest do
 
     checkout_relay
     build_relay
-    port = start_relay
+    os_pid = start_relay
     wait_for_relay(conn)
 
     checkout_mist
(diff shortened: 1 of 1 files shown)


test/integration/relay_test.exs
```diff
@@ -57,11 +59,12 @@ defmodule Integration.RelayTest do
...
   def start_relay do
-    Port.open({:spawn, "iex -S mix"}, cd: "../cog_relay", env: [{'COG_MQTT_PORT', '1884'}])
+    port = Port.open({:spawn, "iex -S mix"}, cd: "../cog_relay", env: [{'COG_MQTT_PORT', '1884'}])
+    {:os_pid, os_pid} = Port.info(port, :os_pid)
+    os_pid
   end
 
-  def stop_relay(port) do
-    {:os_pid, os_pid} = Port.info(port, :os_pid)
+  def stop_relay(os_pid) do
```
mpeck: Is there a reason that all of these helper functions are public?

mpeck:
:+1: Looks good, just had the one question.

test/integration/relay_test.exs
vanstee: Ah no I'll fix that.

Issue #10, opened 2016-01-08
Bundle activation/deactivation via cogctl


kevsmith:
Activate/deactivate bundle on one or more relays. Deactivation is a non-destructive act resulting in a bundle which is unavailable but files are still installed on all Relay hosts. Think of it as taking a bundle out of rotation or pressing "pause" on a bundle.

Example:

```
# One relay
cogctl bundle deactivate --relay=BFC92057-B4AB-43B0-9E4C-D044AEDDCD78 --bundle=BFC92057-B4AB-43B0-9E4C-D044ABDDCD32

# All relays
cogctl bundle deactivate --relay=all --bundle=BFC92057-B4AB-43B0-9E4C-D044ABDDCD32
```

PR #317, opened 2016-02-29
Support Slack emojis as aliases

kevsmith:
This commit adds support for Slack emojis as command aliases. It incorporates parser changes in Piper which allow alias names like `:shipit:` to be parsed and recognized as potential command or alias names.

The following changes were made to Cog:

- Slight tweak to `CommandResolver` to recognize names starting with `:`
  as potential commands or aliases and look them up.
- Tweak to `Models.Command.split_name/1` to correctly split names like
  `user::pager:` or `somebundle::thumbsup:` into bundle and command
  parts.
- Added integration test to exercise emoji alias execution

Requires Piper PR [#10](https://github.com/operable/piper/pull/10)

### DIFF
--- a/lib/cog/commands/alias/helpers.ex
+++ b/lib/cog/commands/alias/helpers.ex
@@ -67,7 +67,7 @@ defmodule Cog.Commands.Alias.Helpers do
   def error({:alias_not_found, alias}),
     do: "I can't find '#{alias}'. Please try again"
   def error(:bad_pattern),
-    do: "Bad pattern. Sorry, but that is an invalid pattern only letters numbers and the following special characters are allowed: *, -, _"
+    do: "Invalid alias name. Only emoji, letters, numbers, and the following special characters are allowed: *, -, _"
   def error(:too_many_wildcards),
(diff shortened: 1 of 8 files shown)


imbriaco:
Pushed a couple of minor tweaks to make `alias:ls` allow emoji characters in the pattern that it ingests. Other than that, LGTM.

lib/cog/commands/alias/helpers.ex
```diff
@@ -67,7 +67,7 @@ defmodule Cog.Commands.Alias.Helpers do
   def error({:alias_not_found, alias}),
     do: "I can't find '#{alias}'. Please try again"
   def error(:bad_pattern),
-    do: "Bad pattern. Sorry, but that is an invalid pattern only letters numbers and the following special characters are allowed: *, -, _"
+    do: "Invalid alias name. Only emoji, letters, numbers, and the following special characters are allowed: *, -, _"
```
mpeck: You can actually pass a simple pattern to `ls` using '*' as a wildcard. To me this message implies that you need to pass the entire alias name.

mpeck:
:+1:

PR #320, opened 2016-02-29
add a special db_error message when an alias name is already in use

### LINKED ISSUE #313: new alias creation error messaging
The error message "Whoops! An error occurred. name has already been taken" which occurs when a user attempts to create an alias with an in use name should read "Whoops! An error occurred. The alias name is already in use."

mpeck:
Return a clearer and better formatted error message when the name already in use error occurs during alias creation.

resolves #313

### DIFF
--- a/lib/cog/commands/alias/helpers.ex
+++ b/lib/cog/commands/alias/helpers.ex
@@ -101,7 +101,13 @@ defmodule Cog.Commands.Alias.Helpers do
     do: EctoJson.render(data)
 
   # Special errors that come when there are issues with the database.
-  defp db_errors(errors),
-    do: Enum.map_join(errors, "\n", fn({key, message}) -> "#{key} #{message}" end)
+  defp db_errors(errors) do
+    Enum.map_join(errors, "\n", fn
+      ({:name, "has already been taken"}) ->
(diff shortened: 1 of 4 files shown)


lib/cog/commands/alias/helpers.ex
```diff
@@ -101,7 +101,13 @@ defmodule Cog.Commands.Alias.Helpers do
     do: EctoJson.render(data)
 
   # Special errors that come when there are issues with the database.
-  defp db_errors(errors),
-    do: Enum.map_join(errors, "\n", fn({key, message}) -> "#{key} #{message}" end)
+  defp db_errors(errors) do
+    Enum.map_join(errors, "\n", fn
+      ({:name, "has already been taken"}) ->
```
christophermaier: Might be better to fix this in the model, instead: [unique_constraint/3](https://hexdocs.pm/ecto/2.0.0-beta.1/Ecto.Changeset.html#unique_constraint/3) (see the `:message` option)

mpeck: TIL: `:message`

It's a bit weird since we have to put the message on both models. But I think I agree with you that moving it to the model is best.

PR #322, opened 2016-03-01
Don't require git SSH setup for scripted install

GregMefford:
GitHub HTTPS URLs don't require you to add your SSH key to your GitHub account prior to cloning.
This is probably a much easier user-experience for someone trying to build Cog, particularly on a machine that doesn't have a key-pair generated yet at all.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -13,7 +13,7 @@ REPOS="cog relay cogctl relayctl"
 COG_REPO="https://github.com/operable/cog.git"
 RELAY_REPO="https://github.com/operable/relay.git"
 COGCTL_REPO="https://github.com/operable/cogctl.git"
-RELAYCTL_REPO="https://github.com/operable/relayctl"
+RELAYCTL_REPO="https://github.com/operable/relayctl.git"
 
 # All the executables needed to clone and
 # build Cog & Relay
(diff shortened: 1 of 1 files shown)


GregMefford:
So, making this change locally, I was able to clone the `cog` repo, but the build immediately fails because it can't clone the `spanner` repo from GitHub because of the same issue. I guess if it's going to be a big overhaul, we should talk about whether it's worthwhile to make things easier, or if we just say you have to have GitHub credentials set up and clone using SSH.

kevsmith:
@GregMefford Thanks for the contribution! Eradicating SSH urls has been a work in progress, tbh. We thought we had them all removed about a week ago but they keep creeping back in.

I'll take a look at the issues you ran into with Spanner and get those fixed up before I merge your PR.

GregMefford:
This seems to be working fine now, along with the change in 373183502ab0b14f0c796aaa080c1364e1886b94. Thanks!

GregMefford:
Actually, it doesn't _quite_ work as-is, because the latest `setup.sh` script checks out the `0.2` tag, which doesn't have this change merged yet, so it still tries to go check out the old version of cog with the git SSH deps.

If I manually change my `setup.sh` script to `RELEASE_TAG="master"`, it seems to work as far as I can tell (haven't actually joined it to a Slack channel quite yet).

kevsmith:
Deps now consistently use public URLs.

Closing.

PR #329, opened 2016-03-02
The `pluck` command

### LINKED ISSUE #255: Extract a "pluck" command

lhaskins:
This PR adds the pluck command to Cog.

The pluck command takes the specified fields and returns those fields and only those fields for each item in the input list.

Example:
```seed '[{"foo": 1, "bar": 2, "baz": 3}, {"foo": 4, "bar": 5, "baz": 6} ]' | pluck --fields=bar```
returns:
```[{"bar": 2}, {"bar": 5}]```

It should be noted, that this command currently does not traverse a JSON path to pluck items.

Also, a small refactor of is_ok was moved to a general Cog utility file for use within this command.

Fixes #255

### CODE REVIEW

File: lib/cog/commands/pluck.ex [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+  defp validate_options(%__MODULE__{req: %{options: _}}=state),
+    do: %{state | fields: []}
+
+  defp contains_field(item, field, acc) do
+    result = Map.fetch(item, field)
+    |> CogHelpers.is_ok?
+    result and acc
+  end
+
+  defp validate_inputs(%__MODULE__{req: %{cog_env: item}}=state),
```
christophermaier: When writing a command using this processing pipeline style, it helps to put the functions in the order in which they are called during processing to aid readability (so `validate_inputs/1` should be above `validate_options/1`)

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,82 @@
...
+    do: %{state | input: item}
+
+  defp execute(%{errors: [_|_]}=state), do: state
+  defp execute(%__MODULE__{fields: [], input: item}=state),
+    do: %{state | output: item}
+  defp execute(%__MODULE__{fields: fields, input: item}=state) when is_map(item),
```
christophermaier: How is `input` ever _not_ going to be a map?

lib/cog/commands/pluck.ex
christophermaier: Also, this function currently doesn't really do much; you could just pull out the `cog_env` directly on Line 36.

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  defp validate(req) do
+    %__MODULE__{req: req}
+    |> validate_inputs
+    |> validate_options
+  end
+
+  defp validate_options(%__MODULE__{req: %{options: %{"fields" => opt_fields}}, input: item}=state) do
+    fields = String.split(opt_fields, ",")
```
christophermaier: What happens if `fields` isn't a string?

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,82 @@
...
+    case Enum.reduce(fields, true, &contains_field(item, &1, &2)) do
```
christophermaier: This would be clearer:
```elixir
if Enum.all?(fields, &Map.has_key(item, &1)) do
  # etc
end
```

christophermaier: Though I'm not 100% sold on the fact that it's an error for the item not to have the key. I could be persuaded either way; probably worth a discussion.

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,82 @@
...
+
+  defp validate_inputs(%__MODULE__{req: %{cog_env: item}}=state),
+    do: %{state | input: item}
+
+  defp execute(%{errors: [_|_]}=state), do: state
+  defp execute(%__MODULE__{fields: [], input: item}=state),
+    do: %{state | output: item}
+  defp execute(%__MODULE__{fields: fields, input: item}=state) when is_map(item),
+    do: %{state | output: Map.take(item, fields)}
+  defp execute(%__MODULE__{fields: _, input: item}=state),
```
christophermaier: What scenario does this function head describe?

lib/cog/util/command_helpers.ex
```diff
@@ -0,0 +1,5 @@
+defmodule Cog.Command.Helpers do
+
+  def add_errors(input, error_or_errors),
```
christophermaier: If you're going to do this, this module needs documentation about what's going on, what the assumptions are, etc. As is, it's not clear.

christophermaier:
Do you have thoughts on how to support paths here?

test/integration/commands/pluck_test.exs
```diff
@@ -0,0 +1,35 @@
...
+  use Cog.AdapterCase, adapter: "test"
+
+  setup do
+    user = user("lucky", first_name: "Shamaus", last_name: "McLucky")
+    |> with_chat_handle_for("test")
+
+    {:ok, %{user: user}}
+  end
+
+  test "with no field option", %{user: user} do
```
christophermaier: So, I have this code lying around from some previous work that ended up getting tabled, and I never got around to extracting it. These functions would be good to add to a command test utility module and would help clean up these tests immensely.

```elixir
 defp seed(input, pipeline) when is_list(input) or is_map(input),
    do: "@bot: seed '#{Poison.encode!(input)}' | #{pipeline}"

  #
  # Example:
  #
  # lookbehind.
  #
  # Now you have `n` problems.

```

lib/cog/commands/pluck.ex
lhaskins: This handles the situation where we are not totally sure what could be coming out from the pipeline. If the result is `nil` for example, we should just pass that on.

lib/cog/commands/pluck.ex
lhaskins: The option type is set to string. I can add another header to error if it's not a string.

christophermaier: I believe there is a bug in the parser (which will probably be fixed soon) that doesn't take that into account. If you can confirm that it doesn't crash with non-string input, this code is fine (and is ultimately where we want to be), but if it does, please add code that guards against that.

lib/cog/commands/pluck.ex
lhaskins: If I list fields that I'd like to pluck and one or more are missing, I'd like to know that. Or if I've misspelled a field name. Otherwise, it's possible to silently fail.

lib/cog/commands/pluck.ex
christophermaier: If `nil` is the only case we care about, then that should be explicitly part of the pattern matching here.

lhaskins:
I started to handle paths with this and realized that things were headed in the same direction as filter. 

When traversing paths in the input's JSON, the fields would be a list of field lists where each field list represented a path in the json output. Then for each item in the field option, we get the results using `get_in` and put them in into a map.

I want to add the path retrieval capability after the executor rewrite is complete and then also add templating to pluck.

lib/cog/commands/pluck.ex
christophermaier: Fair argument. In that case, the names of all fields that are not present should be part of the error.

christophermaier:
It would be similar to what's going on with `filter`, and that's intentional; `filter` and `pluck` would be used quite a bit together.

I'm not clear how that would need to wait on the executor rewrite, though... couldn't you do something like `--paths=foo,bar.baz,blah`?

I'm also not clear on what kind of templating would be used here.

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,82 @@
...
+    fields = String.split(opt_fields, ",")
+    case Enum.reduce(fields, true, &contains_field(item, &1, &2)) do
+      true -> %{state | fields: fields}
+      false -> CommandHelpers.add_errors(state, :invalid_field)
+    end
+  end
+  defp validate_options(%__MODULE__{req: %{options: _}}=state),
+    do: %{state | fields: []}
+
+  defp contains_field(item, field, acc) do
```
kevsmith: `Map.has_key?`, no?

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,82 @@
...
+    result = Map.fetch(item, field)
+    |> CogHelpers.is_ok?
+    result and acc
+  end
+
+  defp validate_inputs(%__MODULE__{req: %{cog_env: item}}=state),
+    do: %{state | input: item}
+
+  defp execute(%{errors: [_|_]}=state), do: state
```
kevsmith: When does this function head match? I'm not seeing where we construct a map with an `:errors` key.

lib/cog/relay/relays.ex
```diff
@@ -130,7 +131,7 @@ defmodule Cog.Relay.Relays do
     {success_bundles, failed_bundles} = announcement
     |> Map.fetch!("bundles")
     |> Enum.map(&lookup_or_install/1)
-    |> Enum.partition(&Util.is_ok?/1)
+    |> Enum.partition(&Helpers.is_ok?/1)
```
kevsmith: What's the motivation here? `Util` and `Helpers` intuitively feel like equivalent names.

kevsmith:
We've fallen back into retrofitting design onto code via PR review. If a plan had been sketched out and reviewed in #255 before writing code I suspect PR review would've been much shorter.

lib/cog/commands/pluck.ex
lhaskins: No, it's not the only case, it's an example. Another example (though admittedly I don't know why anyone would do this), someone could sort some data then pass the data to pluck. Depending on the data it is possible that the input to pluck can be an integer or a string.

lib/cog/commands/pluck.ex
lhaskins: `CommandHelpers.add_errors` updates the map with the `:errors` key.

christophermaier: I think the confusion stems from the fact that the pattern here is `%{errors: ...}` rather than `%__MODULE__{errors: ...}`

lib/cog/relay/relays.ex
lhaskins: The is_ok function was located in Relay.Util. I moved it out into Cog.Helpers - it's a pretty generic function that's useful throughout and not just in the Relay code.

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,112 @@
...
+      > { "qux": "one", "foo.bar": {"baz": "stuff"} }
+        { "qux": "two", "foo.bar": {"baz": "me"} }
+  """
+
+  option "fields", type: "string", required: false
+
+  defstruct req: nil, fields: nil, input: nil, output: nil, errors: []
+
+  def handle_message(req, state) do
+    case req |> validate |> execute |> format do
```
mpeck: Someone recently pointed `with` out to me, and it seems like a pretty useful function for cases like this. You could write this similarly to:
```
...
results = with {:ok, options} <- validate_options(req.options),
                      {:ok, data} <- execute(options, req.args),
                        do: format(data)
case results do
  {:ok, response} ->
    {:reply, req.reply_to, response, state}
  {:error, err} ->
    {:error, req.reply_to, err, state}
end
...
```

Then each of the functions in your flow can just return `{:ok, result}` or `{:error, error}`. `with` will bail on the first error. So you abandon execution at the first sign of trouble. I think it would alleviate the need to pass so much state around. You also don't have to worry about errors from previous functions in functions at the end of the flow. I think it may simplify stuff greatly.

I used a similar approach when I wrote the `which` command. https://github.com/operable/cog/blob/master/lib/cog/commands/which.ex#L31-L42

lib/cog/util/command_helpers.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Cog.Command.Helpers do
```
mpeck: There is actually already a module called `Cog.Command.Helpers`. I created when I did the which command. On another note, I generally like to have module names match their location in the file system. For example, `Cog.Command.Helpers` would live at `lib/cog/command/helpers.ex`.

test/integration/commands/pluck_test.exs
mpeck: I like the suggestion @christophermaier. Another approach that I've taken with some of the tests is to generate the map that I expect from the command's output and render the commands template with it. https://github.com/operable/cog/blob/master/test/integration/commands/which_test.exs#L12-L20

christophermaier: @mpeck I think trying to decouple the tests of the bare command from the templates is the way we ultimately want to go. Otherwise you need to create special "test" templates, which (by definition) only ever get used in the tests. We really shouldn't have test templates.

Having a separate little suite of tests to ensure that all our templates work, given some pre-defined input, would be a good thing to have, too.

mpeck: I agree @christophermaier. Decoupling the testing from the templates is a much better route. I only mention this here because I had such a hard time separating templating from command output when I was writing the alias and which command tests.

lib/cog/commands/pluck.ex
lhaskins: The idea with this flow is to catch all of the errors and then return all the errors found during the course of trying to execute the command. This helps let the user know all of the issues with the command execution as opposed to only the first one.

lib/cog/util/command_helpers.ex
lhaskins: I'll move the function to that module. Thanks!

lib/cog/commands/pluck.ex
mpeck: I get the idea, but if I'm understanding what I'm reading correctly a failure in validate would cause the flow to bypass execution anyway. So you are most likely only ever going to have a single error returned to the user. And if that wasn't the case, often times an error early on will ripple through the rest of flow causing more errors down the line. And don't get me wrong, I'm not saying what you did is wrong, but in my opinion it keeps things a lot simpler when you fail quickly.

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,115 @@
...
+  use Spanner.GenCommand.Base, bundle: Cog.embedded_bundle, enforcing: false, calling_convention: :all
+  alias Cog.Commands.Helpers, as: CommandHelpers
+
+  @moduledoc """
+  Returns the specified fields from a set of inputs.
+  Be sure to add quotes when traversing a JSON path.
+
+  ## Example
+
+      @bot #{Cog.embedded_bundle}:rules --list --for-command="#{Cog.embedded_bundle}:permissions" | #{Cog.embedded_bundle}:pluck --fields=id
```
vanstee: I know this might not be as future proof, but the `Cog.embedded_bundle` bit makes it a little bit harder to read in the raw docs.

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,115 @@
...
+      > [ {"foo": {"bar.qux": {"baz": "stuff"} } }, {"foo": {"bar": {"baz": "me"} } } ]
+      @bot #{Cog.embedded_bundle}:seed '[{"foo":{"bar":{"baz":"stuff"}}}, {"foo": {"bar":{"baz":"me"}}}]' | #{Cog.embedded_bundle}:pluck --fields="foo.bar"
+      > { "foo.bar": {"baz": "stuff"} }
+        { "foo.bar": {"baz": "me"} }
+      @bot #{Cog.embedded_bundle}:seed '[{"foo":{"bar":{"baz":"stuff"}}, "qux": "one"}, {"foo": {"bar":{"baz":"me"}}, "qux": "two"}]' | #{Cog.embedded_bundle}:pluck --fields="qux,\"foo.bar\""
+      > { "qux": "one", "foo.bar": {"baz": "stuff"} }
+        { "qux": "two", "foo.bar": {"baz": "me"} }
+  """
+
+  option "fields", type: "string", required: false
```
vanstee: I think you can use the `list` type here, which splits a string on `,` for you.

lib/cog/commands/pluck.ex
```diff
@@ -0,0 +1,115 @@
...
+    end
+  end
+
+  defp validate(%{cog_env: item}=req) do
+    %__MODULE__{req: req, input: item}
+    |> validate_options
+  end
+
+  defp validate_options(%__MODULE__{req: %{options: %{"fields" => opt_fields}}}=state) when not is_binary(opt_fields),
+    do: CommandHelpers.add_errors(state, {:not_string, opt_fields})
```
vanstee: Since the `fields` option is currently defined a as a string, is there a way that this command could receive a non-string value? Seems like something we might want to enforce before it gets to the command.

lhaskins:
Closing since changes in piper/executor may make this command unnecessary.

PR #330, opened 2016-03-02
IRC adapter

### LINKED ISSUE #307: Implement IRC adapter
Now that we've refactored the adapter behavior a bit, it's time to put it to the test with a new adapter. Folks have been asking for an IRC adapter and it's significantly different from our Slack and HipChat adapters so it should be a good way to check if the adapter behavior has the right API.

vanstee:
The IRC adapter works. You can set options just as you'd expect which allows cog to connect to your IRC channel (currently only 1 channel is supported). Mentioned commands and spoken commands are supported, as well as user private message redirection.

Part of this PR was to take a fresh look at the new adapter refactor and see if it is well designed. I think the refactor has improved the interface but I think there are a few extra things we can add to make it easier to write adapters in the future:

* Remove `lookup_room` calls from the executor that pass along a string. This makes for some function definitions. We should instead implement it by having each adapter identify a string as a room or a user (used for redirection) and then having the executor call `lookup_room(name: room_name)` or `lookup_direct_room(name: user_name)`.
* Add some code to the base adapter module that handles converting a message with a mentioned command or spoken command into a raw command (stripping the bot's handle or the command prefix). Something like a `parse_command` function that returns `{true, raw_command}` or `false` should do for now.

One area that I would like some help with is writing an integration test. For now I figured we could go ahead and merge this, but it'd be nice to go back and write an integration test against a real IRC server. The main problem was finding a hosted IRC service. We could standup our own IRC server on each build node but that seemed like headache to manage.

Closes #307

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -34,6 +34,12 @@ config :cog, Cog.Adapters.HipChat,
   api_token: System.get_env("HIPCHAT_API_TOKEN"),
   mention_name: System.get_env("HIPCHAT_MENTION_NAME")
 
+config :cog, Cog.Adapters.IRC,
+  host: System.get_env("IRC_HOST"),
+  port: System.get_env("IRC_PORT"),
+  channel: System.get_env("IRC_CHANNEL"),
+  nick: System.get_env("IRC_NICK")
+
 # ========================================================================
 # Commands, Bundles, and Services
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -39,6 +39,9 @@ config :cog, Cog.Adapters.IRC,
   port: System.get_env("IRC_PORT"),
   channel: System.get_env("IRC_CHANNEL"),
   nick: System.get_env("IRC_NICK"),
+  user: System.get_env("IRC_USER"),
+  name: System.get_env("IRC_NAME"),
+  password: System.get_env("IRC_PASSWORD"),
   use_ssl: System.get_env("IRC_USE_SSL") || true
 
 # ========================================================================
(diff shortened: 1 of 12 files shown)


mpeck:
:+1:

varac:
Hej, i'm confused - this looks like IRC *is* supported, https://cog-book.operable.io/#_introducing_cog says it isn't.
https://operable.io/cog_vs_friends.html also says IRC is not supported.

Could you please share some light here ? And update the docs if it's really supported ?

vanstee:
@varac We used to have an IRC adapter, as shown here. But, to better support rendering templates for each chat provider, we rewrote the adapter interface along with the Slack and HipChat adapters. However, due to limited use we still haven't reimplemented the IRC adapter. If you'd like to give it a shot, let me know and I can point you in the right direction.

varac:
@vanstee: Thats sad, leaving cog to only work with proprietary backends. And sorry no, I'm not the one to do it. I'd just like to use cog together with a free-software based adapter. Please consider re-adding it, and give it a while until ppl actually adopt it.

kevsmith:
@varac Please open a ticket so we can keep track of your request. Commenting on a 7+ months old closed pull request almost guarantees it will fall off our radar.

kevsmith:
@varac I get the impression from your comment that you're frustrated and/or disappointed with our decision to stop developing a IRC adapter. Keep in mind our team is small and so we can't build everything we want just yet. Any assistance we get, even if it's just registering interest in a feature with a ticket, really does help.

varac:
@kevsmith ok, thx, I'll do.

PR #332, opened 2016-03-03
VM sanity check

### LINKED ISSUE #331: Verify Erlang VM settings at launch
We have been bitten multiple times by problems with the Erlang VM not being configured as we need for Cog. We should investigate the possibility of adding some code that runs at startup to verify that the VM is running with dirty schedulers enabled. If we don't see dirty schedulers enabled, we should check to see if SMP is enabled.

We should abort with a useful error message that tells the user what to do to fix things.

kevsmith:
Fixes #331

### DIFF
--- a/lib/cog.ex
+++ b/lib/cog.ex
@@ -78,6 +78,7 @@ defmodule Cog do
     end
     unless smp_status == :ok and ds_status == :ok do
       Logger.error("Application start aborted.")
+      Logger.flush()
       :init.stop()
     end
   end
--- a/lib/cog.ex
+++ b/lib/cog.ex
@@ -89,8 +89,8 @@ defmodule Cog do
 SMP support disabled.
 SMP support can be enabled via one of the following:
 
-1. Add '--erl "-smp enable"' to the Elixir args in Cog's launch script.
-2. Add '-smp enable" to the $ERL_FLAGS environment variable.
+  1. Add '--erl "-smp enable"' to the Elixir args in Cog's launch script.
+  2. Add '-smp enable" to the $ERL_FLAGS environment variable.
 """}
     else
       {:ok, "SMP support enabled."}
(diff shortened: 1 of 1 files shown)


lib/cog.ex
```diff
@@ -61,4 +62,45 @@ defmodule Cog do
...
+    unless smp_status == :ok and ds_status == :ok do
+      Logger.error("Application start aborted.")
+      Logger.flush()
+      :init.stop()
+    end
+  end
+
+  defp verify_smp() do
+    if :erlang.system_info(:schedulers_online) < 2 do
+      {:error, "SMP support disabled. Add '-smp enable' to $ERL_FLAGS and restart Cog."}
```
vanstee: Since we're not using `ERL_FLAGS` to enable this when run with `iex` we could add that info in here as well.

kevsmith: What info?

vanstee: Setting `iex --erl "-smp enable"` rather than setting `ERL_FLAGS`.

lib/cog.ex
```diff
@@ -61,4 +62,45 @@ defmodule Cog do
...
+      Logger.info(smp_message)
+    else
+      Logger.error(smp_message)
+    end
+    if ds_status == :ok do
+      Logger.info(ds_message)
+    else
+      Logger.error(ds_message)
+    end
+    unless smp_status == :ok and ds_status == :ok do
```
vanstee: `unless` + `and` is always kinda confusing for me. Mind changing this to `if smp_status == :error or ds_status == :error`?

lib/cog.ex
```diff
@@ -5,6 +5,7 @@ defmodule Cog do
   import Supervisor.Spec, warn: false
 
   def start(_type, _args) do
+    sanity_check()
```
vanstee: A better name for this might be `verify_erlang_options`.

vanstee:
LGTM

PR #333, opened 2016-03-03
Set ERL_FLAGS to use SMP and dirty schedulers

lhaskins:
This sets the environment variable ERL_FLAGS="-smp enable". This allows Cog and Relay to execute using SMP and dirty schedulers.


mpeck:
Looks good to me @lhaskins :+1:

PR #336, opened 2016-03-04
Adjust alias search algorithm

christophermaier:
See commit messages for details. Much of this is preparatory for the executor rewrite. Work will be done to make this more efficient, but that's all tied up in the rewrite.


kevsmith:
:+1:

Nice cleanup!

PR #337, opened 2016-03-04
Which

### LINKED ISSUE #334: Promote `alias which` command to a top-level command
Currently `operable:alias which <ALIAS>` tells a user whether the given alias is a site alias or their own user alias. It would be useful to move this to the top level (i.e., an `operable:which` command) that could operate on aliases as well as commands.

For instance, if there was a site alias that shadowed an installed command (similar to how many developers have shell aliases like `alias ls="ls -CF"`), a `which` command could tell a user this, without requiring them to first know it was an alias.

mpeck:
This implements a top level which command. Passing a string that resolves to a command will return `command - <bundle>:<command name>`. An alias will return `alias - <user | site>:<alias name> -> "command string"`. Aliases take precedence over commands. So if an alias shadows a command the alias will be returned instead of the command. I'm not sure what the planned precedence will be in the new executor, so we may need to swap this.

resolves #334


christophermaier:
Looks good overall. I think we need a way to consolidate the alias / command resolution logic that now lives here and in the executor, though (see https://github.com/operable/cog/pull/336). I'm not sure if now is quite the best time to do this, given the major surgery going on in the executor. Can you make an issue to revisit this unification as soon as the executor rewrite is done?

mpeck:
@christophermaier are we good to merge this guy?

christophermaier:
@mpeck Indeed; :shipit:

PR #355, opened 2016-03-08
Remove some unused command-related queries

christophermaier:
`Cog.Queries.Command.options_for/1` wasn't used.

`Cog.Queries.Command.rules_for/1` was only used in a test, which is
now rewritten.

### DIFF
--- a/test/cog/rule_ingestion_test.exs
+++ b/test/cog/rule_ingestion_test.exs
@@ -1,6 +1,5 @@
 defmodule RuleIngestionTest do
   use Cog.ModelCase
-  alias Cog.Queries
   alias Piper.Permissions.Parser
 
   test "a valid rule gets parsed and saved" do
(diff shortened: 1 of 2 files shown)


test/cog/rule_ingestion_test.exs
```diff
@@ -77,7 +76,8 @@ defmodule RuleIngestionTest do
     permission("s3:delete")
     inserted_rule = rule("when command is cog:s3 with option[op] == 'delete' must have s3:delete")
 
-    %Cog.Models.Rule{parse_tree: retrieved_tree} = Queries.Command.rules_for(command) |> Cog.Repo.one!
+    command = Repo.preload(command, :rules)
```
vanstee: If you're only going to be working with rules after this, using an `assoc` query might be more intention revealing.

```
rules = Repo.all(assoc(command, :rules))
```

vanstee: Also since it looks like you only intend to have a single rule returned you could use `Repo.one!` which would be even nicer.

vanstee:
Looks good. Just a minor suggestion but :+1: to merge either way.

test/cog/rule_ingestion_test.exs
christophermaier: @vanstee Oh, nice... TIL :+1:

vanstee:
LGTM

PR #361, opened 2016-03-09
Use a config schema for validating the slack config

vanstee:


lhaskins:
I'm assuming that each adapter will now have it's own api.ex, config.ex, rtm_connector.ex (or some equivalent - maybe nice to rename this to connector or connection), supervisor.ex, and formatter.ex

Otherwise, LGTM.

lib/cog/adapters/slack/config.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Cog.Adapters.Slack.Config do
+  use Cog.Adapters.Config,
+    key: Cog.Adapters.Slack,
+    schema: [api:
+              [{:token, [:required], :api_token},
+               {:cache_ttl, [:integer], :api_cache_ttl}]]
+end
```
kevsmith: Is `cache_ttl` required? Does `token` have a type?

vanstee: `cache_ttl` is not required since we have a default and `token` is a string which we don't have a coercion for. I think the config stuff for the adapter might be able to be rewritten with the json schema things once that's figured out.

mpeck:
@vanstee is this something we are planning on merging? The json schema bits are in place for bundle configs, but it isn't really generalized yet. I don't think it would take much to make it more generally applicable, but I don't think that's something we will be able to do if we want to release tomorrow.

vanstee:
Oh whoops I forgot about this. Yes I plan to merge this today along with https://github.com/operable/cog/pull/381

PR #364, opened 2016-03-09
Add setup instructions to README

jsteiner:


imbriaco:
I love the idea of showing people how to get up and run developing quickly. These steps have a couple of implicit assumptions that I think may trip people up which we should document before merging, though.

1. They assume that Postgres is installed and available and that the current user can create users and databases.
2. They assume chat provider environment variables are available, otherwise Cog will crash the first time you `make run`.

jsteiner:
@imbriaco updated per your suggestion

PR #368, opened 2016-03-10
Issue #362: Make first_name and last_name optional fields

### LINKED ISSUE #362: Optional first and last names for users
When creating a user via the API, the user's first name and last name are not required fields. We should be able to create a user without these fields.

lhaskins:
The first and last names in the User model should be optional fields. We allow the database to accept null values for first and last names.

Fixes #362


lhaskins:
The test failure is a separate issue unrelated to this PR.

Issue #229, opened 2016-02-12
Unknown users should be able to interact with "safe" commands


kevsmith:
Users unknown to Cog -- users without Cog accounts -- should be able to interact with safe commands like `echo`. Allowing this type of interaction simplifies set up and speeds end-to-end sanity tests.

Reported by @ktheory

mpeck:
I think this may have some bearing to #286

vanstee:
@kevsmith Do you think COG_ALLOW_SELF_REGISTRATION is an ok solution to this for now, or is this still needed.

PR #369, opened 2016-03-10
Add a basic glossary outline

drapergeek:
This starts the basic of a glossary in the repo. This is something we've talked about a bit while moving forward in development and I wanted to get a structure started so that we could start adding to it.

Some of this information is already in the Wiki so you're probably asking yourself "Why would we move this into the repo?" I'm glad you asked!

1. Moving it into the repo allows it to be versioned alongside the code. As the code changes and new ideas and concepts are added and removed, this file can change with the code. It may fall out of date at periods, that inevitable but if properly maintained, you can look back at a previous release and understand the concepts that are particular to that release that may have changed in a newer release.

2. Wikis do not have any way of reviewing feedback. If I have something I'd like to add to the Wiki and I add it, it may be 100% wrong and no one realizes that. Moving this to the repo allows me to get feedback before it's merged.

3. Wikis do not allow external contributions (at least not that I'm aware of). I'd like to be able to update this without having access to the repo. That's not possible with a Wiki. Moving this also makes it easy for the community to get involved with updating it.

I haven't filled in much as my understanding is still kind of vague but I wanted something to start adding PRs against.

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -1,0 +1,22 @@
+# Cog Glossary
+
+## Bundle
+## Command
+### Subcommand
+## Group
+## Permission
+  * A way to define what commands can be accessed
+  * Come with bundles
+  * Can be created custom in the `site:` namespace
+
+### Examples:
+  * `operable:manage_commands`
+  * `ec2:delete`
+
+## Relay
+## Role
+  * Group of permissions
+  * Can be attached to a group
+## Rule
+## Tag
+## User


GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
+# Cog Glossary
+
+## Bundle
```
lhaskins: Bundle - a package containing one or more commands and the attributes (rules, permissions, possible options, documentation, templates, etc.) that belong to each command. A bundle can contain all the code for the command execution. https://github.com/operable/cog/wiki/Commands-and-Bundles#bundles

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+## Command
```
lhaskins: Command - executable; directive that performs a certain task. https://github.com/operable/cog/wiki/Commands-and-Bundles#commands

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Examples:
+  * `operable:manage_commands`
+  * `ec2:delete`
+
+## Relay
+## Role
+  * Group of permissions
+  * Can be attached to a group
+## Rule
```
lhaskins: Rule - governs the execution of a command.  https://github.com/operable/cog/wiki/A-Tour-Through-Cog's-Authorization-System#authorization-rules

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
+# Cog Glossary
+
+## Bundle
+## Command
+### Subcommand
+## Group
```
lhaskins: Group - the classification of a number of persons within an organization; Examples: operations, engineering, sales. It is possible to have groups within groups. Example: the Product and Devops groups can belong to the engineering group. https://github.com/operable/cog/wiki/A-Tour-Through-Cog%27s-Authorization-System#groups

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+## Permission
+  * A way to define what commands can be accessed
+  * Come with bundles
```
lhaskins: command specific permissions come with bundles.

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+  * Can be created custom in the `site:` namespace
+
+### Examples:
+  * `operable:manage_commands`
+  * `ec2:delete`
```
lhaskins: Another example: `site:deploy-prod`

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Relay
+## Role
+  * Group of permissions
```
lhaskins: Role is more about what a group or user would be performing. This is useful especially when you have a role that spans multiple groups or multiple roles within a single group. It allows for flexibility depending on how an organization is set up. For example the `management` role can spread across multiple groups; while within the operations group you can have `admin` and `support` roles. A group or a user can have a role. Then you can have permissions that are based on the role.

imbriaco: Language is tricky for these, so I'm going to be super picky. As far as glossary definitions are concerned:

- Roles are **named collections of permissions**.
- Roles may be **granted** to **users** or **groups**.

I know that's super picky.

There are a lot of reasons why people might want to define roles, but I think should avoid talking about them those reasons in the glossary. We should definitely talk about those sorts of things in a best practices document for Roles at some point, though.

lhaskins: That's fair. My hope was to bring more clarity. "Group of permissions" seemed pretty vague and I didn'r want to use other glossary terms in the definition.

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+  * Can be attached to a group
+## Rule
+## Tag
+## User
```
lhaskins: User - people executing Cog commands. Currently, permissions must be attached to a Cog user in some manner(whether directly or via groups or roles). A Cog user can have multiple accounts in different chat adapters that are associated with the Cog user.  https://github.com/operable/cog/wiki/A-Tour-Through-Cog%27s-Authorization-System#users (This definition may change slightly if/when we implement Issue #229)

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
+# Cog Glossary
+
+## Bundle
+## Command
+### Subcommand
```
lhaskins: Subcommand - tasks performed under a specified rule. You typically find a subcommand when you have a command that can perform different tasks. For example, the `rules` command can perform a `list`, `create`, or `delete` task. These are the subcommands.

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+## Permission
+  * A way to define what commands can be accessed
+  * Come with bundles
+  * Can be created custom in the `site:` namespace
+
+### Examples:
+  * `operable:manage_commands`
+  * `ec2:delete`
+
+## Relay
```
lhaskins: Relay - the execution environment for Cog commands (non-embedded bundled) https://github.com/operable/cog/wiki/The-Architecture-of-the-Cog-System#relay

GLOSSARY.md
```diff
@@ -0,0 +1,22 @@
...
+## Role
+  * Group of permissions
+  * Can be attached to a group
+## Rule
+## Tag
```
lhaskins: Tag - **Not currently implemented** The idea is for Relay to have a name, but can have multiple tags associated with it. This tag can be used when defining rules for certain bundled commands. Tags will be immutable. This reduces the complexity of needing to update rules with certain tags when a tag is changed for a certain relay or group of relays.

imbriaco: Tag - A named collection of Relays. Tags are used to determine which relays bundles are available on. Additionally, tags may be used within rules to signify that matching command invocations should be executed on a relay associated with that tag. (Future Feature)

GLOSSARY.md
imbriaco: Got it. I'd prefer to keep the glossary specifically focused on definitions (ie: "named collection of permissions") and avoid going into our intent.

PR #381, opened 2016-03-15
Check HipChat config when starting the HipChat Adapater

vanstee:
We now validate the config and then also try to make an authenticated request when starting the HipChat.API process. To do that (and avoid having to cache the config) we converted the HipChat.API module into a GenServer so we have a good place to check the config and crash the entire Cog application as well as hold onto the config between API calls.

### DIFF
--- a/lib/cog/adapters/hipchat/api.ex
+++ b/lib/cog/adapters/hipchat/api.ex
@@ -145,18 +145,18 @@ defmodule Cog.Adapters.HipChat.API do
     end
   end
 
-  def include_headers(options, config) do
+  defp include_headers(options, config) do
     headers = request_headers(config)
     Keyword.update(options, :headers, headers, &(headers ++ &1))
   end
 
-  def request_headers(%{token: token}) do
+  defp request_headers(%{token: token}) do
     ["Accept":        "application/json",
      "Content-Type":  "application/json",
      "Authorization": "Bearer #{token}"]
   end
 
-  def encode_body(options) do
+  defp encode_body(options) do
     case Keyword.has_key?(options, :body) do
       true ->
         Keyword.update!(options, :body, &Poison.encode!/1)
@@ -165,15 +165,15 @@ defmodule Cog.Adapters.HipChat.API do
     end
   end
 
-  def normalize_user(%{"id" => id, "mention_name" => handle, "name" => name}) do
+  defp normalize_user(%{"id" => id, "mention_name" => handle, "name" => name}) do
     %{id: id, handle: handle, name: name}
   end
 
-  def normalize_room(%{"id" => id, "name" => name}) do
+  defp normalize_room(%{"id" => id, "name" => name}) do
     %{id: id, name: name}
   end
 
-  def normalize_message(%{"message" => message}) do
+  defp normalize_message(%{"message" => message}) do
     message
   end
 end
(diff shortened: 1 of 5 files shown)


kevsmith:
What does the output look like if the HipChat adapter is misconfigured?

vanstee:
Here's an example of config with missing `mention_name` and `token`. I do wish it was a bit more clear for folks new to erlang/elixir, but it's not too difficult to figure out.

```
2016-03-15T13:09:16.0510  (Cog.Adapters.Config:115) [error] Required key :mention_name was not provided
2016-03-15T13:09:16.0510  (Cog.Adapters.Config:115) [error] Required key :token was not provided
2016-03-15T13:09:16.0523  () [info] Application cog exited: Cog.start(:normal, []) returned an error: shutdown: failed to start child: Cog.Adapters.HipChat.Supervisor
    ** (EXIT) an exception was raised:
        ** (RuntimeError) Invalid config found for Cog.Adapters.HipChat
            (cog) lib/cog/adapters/config.ex:116: Cog.Adapters.Config.fetch_config!/2
            (cog) lib/cog/adapters/hipchat/supervisor.ex:10: Cog.Adapters.HipChat.Supervisor.init/1
            (stdlib) supervisor.erl:272: :supervisor.init/1
            (stdlib) gen_server.erl:328: :gen_server.init_it/6
            (stdlib) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```

lib/cog/adapters/hipchat/api.ex
```diff
@@ -1,153 +1,219 @@
...
+    {:ok, %{"direct" => user_id}}
+  end
 
-    user = Enum.find(members, fn(user) ->
-      Map.get(user, "name") == user_name
-    end)
+  def lookup_room_user(room_name, user_name) do
+    with {:ok, room} <- lookup_room(name: room_name),
+         {:ok, users} <- lookup_users() do
+      user = Enum.find(users, &match?(%{name: ^user_name}, &1))
```
mpeck: Why don't you use `lookup_user/1` here?

lib/cog/adapters/hipchat/api.ex
```diff
@@ -1,153 +1,219 @@
...
+    end
+
+    {:reply, result, state}
   end
 
-  defp normalize_last_message(%{"items" => [_sent_message, item|_]}),
-    do: item["message"]
-  defp normalize_last_message(%{"items" => _}),
-    do: nil
+  def handle_call({:retrieve_last_message, %{id: room_id}, not_before}, _from, state) do
```
mpeck: What do we actually use this function for?

mpeck:
This looks nice. Seems a lot cleaner. :+1:

lib/cog/adapters/hipchat/api.ex
vanstee: This is used in the integration tests. I could probably only include it if `MIX_ENV=test` but I sorta just wanted to replicate the API we already had when converting this over.

lib/cog/adapters/hipchat/api.ex
vanstee: Not sure, this was just pulled from the old codebase. Let me try that out before I merge this.

PR #388, opened 2016-03-15
Start Endpoint process last in supervision tree

vanstee:
The application isn't ready to receive HTTP requests until we've
connected to the database and we've installed the embedded bundle, which
includes creating permissions necessary for bootstrapping. So, Ecto.Repo
and all bundle related processes need to have started before the
Endpoint process.

### DIFF
--- a/lib/cog.ex
+++ b/lib/cog.ex
@@ -29,20 +29,20 @@ defmodule Cog do
   def site_namespace, do: "site"
 
   defp build_children(:dev, nochat, _) when nochat != nil do
-    [supervisor(Cog.Endpoint, []),
-     worker(Cog.Repo, []),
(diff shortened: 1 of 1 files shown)


lib/cog.ex
```diff
@@ -29,20 +29,20 @@ defmodule Cog do
   def site_namespace, do: "site"
 
   defp build_children(:dev, nochat, _) when nochat != nil do
```
kevsmith: Didn't we nuke the `NOCHAT` env var a while back? If so we can drop this whole function head.

vanstee: We talked about it but I don't think we ever circled back on it. I can nuke all that in a second PR.

kevsmith:
:+1: to merge

PR #389, opened 2016-03-15
Enable spoken commands via env var

### LINKED ISSUE #379: Make enable_spoken_commands configurable via an env var
Since some folks who are running cog in docker want to enable this, we should make this configurable via an env var, so they don't have to modify the `config/config.exs` file. Our docker compose setup should also include this env var.

To verify this works, running `ENABLE_SPOKEN_COMMANDS=true docker-compose up` should boot docker with support for spoken commands.

kevsmith:
Fixes #379


vanstee:
:+1:

PR #390, opened 2016-03-15
New and improved executor

### LINKED ISSUE #286: Executor Refactor

christophermaier:
Uses the new parser and simplifies A LOT of what the executor does.

Introduces a new concept of an "execution plan" for pipeline stages,
in an attempt to decouple from the parser AST (though that's not quite
complete just yet).

Lots more remains to be done, but this takes care of a huge chunk of
the work.

Closes #286


lib/cog/command/option_interpreter.ex
```diff
@@ -1,20 +1,20 @@
...
+  @truthy_values ["true", "t", "y", "yes", "on"]
 
-  def initialize(%Ast.Invocation{}=command, raw) do
-    case CommandCache.fetch_options(command, :prepared) do
-      :not_found ->
-        :not_found
-      {:ok, defs} ->
-        case interpret(defs, raw, [], %{}) do
+  def initialize(%Ast.Invocation{args: args, meta: command_model}=invocation) do
+    case Enum.reduce(command_model.options, %{}, &(prepare_option(&1, &2))) do
```
vanstee: No need to wrap `prepare_option` in an anonymous function. Could just use `&prepare_option/2`

lib/cog/command/command_resolver.ex
```diff
@@ -1,54 +1,73 @@
...
-    case Repo.all(AliasQuery.user_alias_by_name(user, name)) do
+  defp find_one(query) do
+    case Repo.one(query) do
+      nil -> :not_found
+      result -> result
+    end
+  end
+
+  defp lookup(bare_command_name) do
+    case Repo.all(Queries.Command.by_any_name(bare_command_name))do
```
vanstee: Missing a space before `do`

lib/cog/command/option_interpreter.ex
```diff
@@ -1,20 +1,20 @@
 defmodule Cog.Command.OptionInterpreter do
   require Logger
-  alias Cog.Command.CommandCache
+  alias Cog.Models.CommandOption
   alias Piper.Command.Ast
 
-  @truthy_values ["1", "t", "true", "y", "yes", "on"]
+  require Logger
```
vanstee: Dup of line 2

lib/cog/command/pipeline/executor.ex
```diff
@@ -149,36 +117,51 @@ defmodule Cog.Command.Pipeline.Executor do
...
-        {:next_state, :lookup_redirects, %{state | pipeline: pipeline}, 0}
+        # Extract and unwrap redirect information here; keeps AST
+        # details from creeping around, and eliminates the need to
+        # keep the pipeline around
+        redirect_destinations = case pipeline.redirect_to do
+                                  nil ->
+                                    []
+                                  %Ast.Redirect{targets: targets} ->
+                                    Enum.map(targets, &(&1.value))
+                                end
```
kevsmith: Helper function on `Ast.Pipeline` would prevent the details of `Ast.Redirect` from leaking in here.

```
redirect_destinations = Ast.Pipeline.redirect_targets(pipeline) # Returns empty list or list of strings
```

lib/cog/command/option_interpreter.ex
vanstee: Might be a good place to use `with`

```
defs = Enum.reduce(command_model.options, %{}, &(prepare_option(&1, &2)))

with {:ok, options, args} <- interpret(defs, args, [], %{}),
     :ok <- check_required_options(defs, options) do
  options = set_defaults(invocation, options)
  {:ok, options, args}
end
```

Also not sure if `defs` can ever not be a map so that first case might be unnecessary.

lib/cog/command/option_interpreter.ex
```diff
@@ -25,127 +25,123 @@ defmodule Cog.Command.OptionInterpreter do
...
-    {:ok, state, Enum.reverse(args)}
-  end
-  defp interpret(defs, [%Ast.Option{flag: flag, value: nil}=opt|t], args, state) do
-    case Map.get(defs, flag) do
+  defp interpret(_defs, nil, true_args, validated_options),
+    do: {:ok, validated_options, true_args}
+  defp interpret(_defs, [], true_args, validated_options),
+    do: {:ok, validated_options, Enum.reverse(true_args)}
+  defp interpret(defs, [%Ast.Option{name: %Ast.String{value: name}, value: nil}=opt|t], true_args, validated_options) do
+    case Map.get(defs, name) do
```
vanstee: If you're checking for the key rather than checking if it's non-nil, `Map.fetch` is probably more explicit, since it returns an ok/error tuple.

lib/cog/command/option_interpreter.ex
```diff
@@ -25,127 +25,123 @@ defmodule Cog.Command.OptionInterpreter do
...
           error ->
             error
         end
     end
   end
-  defp interpret(defs, [%Ast.Option{flag: flag, value: value}=opt|t], args, state) do
-    case Map.get(defs, flag) do
+  defp interpret(defs, [%Ast.Option{name: %Ast.String{value: name},
+                                    value: value}=opt|t], true_args, validated_options) do
+    case Map.get(defs, name) do
```
vanstee: Same comment about `Map.fetch` from above.

lib/cog/command/option_interpreter.ex
```diff
@@ -25,127 +25,123 @@ defmodule Cog.Command.OptionInterpreter do
...
+    do: {:ok, validated_options, true_args}
+  defp interpret(_defs, [], true_args, validated_options),
+    do: {:ok, validated_options, Enum.reverse(true_args)}
+  defp interpret(defs, [%Ast.Option{name: %Ast.String{value: name}, value: nil}=opt|t], true_args, validated_options) do
+    case Map.get(defs, name) do
       nil ->
-        Logger.debug("Skipping unknown flag #{flag}")
-        interpret(defs, t, args, state)
+        Logger.debug("Skipping unknown name #{name}")
+        interpret(defs, t, true_args, validated_options)
```
vanstee: Might be nice to pull this outer `case` statement + error case into it's own function since it's duplicated below in the next func def as well.

lib/cog/command/option_interpreter.ex
```diff
@@ -25,127 +25,123 @@ defmodule Cog.Command.OptionInterpreter do
...
+    defs = command.meta.options
+    Enum.reduce(defs, options, &(set_default_value(&1.name, &1.option_type.name, &2)))
   end
 
+  defp set_default_value(name, "incr", options) do
+    if Map.has_key?(options, name) == false do
+      Map.put(options, name, 0)
+    else
+      options
+    end
```
vanstee: You can replace this `if else` statement with `Map.put_new(options, name, 0)`

lib/cog/command/pipeline/binder.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Cog.Command.Pipeline.Binder do
+  alias Piper.Command.Bind.Scope
+  alias Piper.Command.Bindable
+
+  def bind(unbound_invocation, context) when is_map(context) do
+    scope = Scope.from_map(context)
+    try do
```
vanstee: Seems weird that `Bindable.resolve` can either return an ok/error tuple OR raise an error. Maybe we should change that API to have both `resolve` and `resolve!` with only the latter raising exceptions.

lib/cog/command/pipeline/binder.ex
```diff
@@ -0,0 +1,25 @@
...
+      case Bindable.resolve(unbound_invocation, scope) do
```
vanstee: Another good place to use `with`.

lib/cog/command/pipeline/executor.ex
```diff
@@ -56,69 +29,64 @@ defmodule Cog.Command.Pipeline.Executor do
...
 
   use Adz
-  use Cog.Util.Debug
 
-  def start_link(request) do
-    :gen_fsm.start_link(__MODULE__, [request], [])
-  end
+  # Provide an empty map for initial binding. The `nil` is for a
+  # template, and just to maintain the proper "shape" of the data
+  @seed_context [{%{}, nil}]
```
vanstee: :+1: I like how this simplifies the code a bit.

lib/cog/command/pipeline/executor.ex
```diff
@@ -198,169 +181,107 @@ defmodule Cog.Command.Pipeline.Executor do
...
+
+    case Cog.Command.Pipeline.Planner.plan(current_invocation, context, permissions) do
+      {:ok, plans} ->
+        {:next_state, :execute_plan, %{state |
+                                       output: [],
+                                       current_plan: nil,
+                                       plans: plans,
+                                       invocations: remaining}, 0}
       {:error, {:not_found, var}} ->
         Helpers.send_reply("I can't find the variable '$#{var}'.", state.request, state.mq_conn)
```
kevsmith: Outside the scope of this PR but we should settle on Cog's voice. Some errors are written in a formal somewhat detached tone whilst others use informal and familiar phrasing.

lib/cog/command/pipeline/executor.ex
```diff
@@ -149,36 +117,51 @@ defmodule Cog.Command.Pipeline.Executor do
...
-  `parse` -> {:stop, :shutdown} | {:next_state, :bind}
-
-  After connecting the request string is parsed and an invocation
-  (`%Piper.Ast.Invocation{}`) or a pipeline(`%Piper.Ast.Pipeline{}`) is added
-  to `remaining` in the state.
-  """
   def parse(:timeout, state) do
-    case Parser.scan_and_parse(state.request["text"], command_resolver: &CommandResolver.find_bundle(&1, state.user), return_pipeline: true) do
+    options = %ParserOptions{resolver: CommandResolver.command_resolver_fn(state.user)}
+    case Parser.scan_and_parse(state.request["text"], options) do
```
vanstee: Do we have any plans to add a struct for request? Seems weird that we're indexing into strings in a lot of places.

lib/cog/command/pipeline/executor.ex
```diff
@@ -198,169 +181,107 @@ defmodule Cog.Command.Pipeline.Executor do
...
+  def plan_next_invocation(:timeout, %__MODULE__{invocations: [], output: output, destinations: destinations}=state) do
+    # No more invocations, but we've got data; run it through
+    # templating and send it out to all the destinations
+    command = state.current_plan.command
+    adapter = state.request["adapter"]
 
-  @doc """
-  `maybe_enforce` -> {:next_state, :maybe_collect_command} | {:next_state, :check_permission}
+    message = Enum.map_join(output, "\n", fn({data,template}) ->
+      render_template(command, adapter, template, data)
```
vanstee: Does this mean that we're now rendering a template per map in the final output list?

lib/cog/command/pipeline/executor.ex
```diff
@@ -198,169 +181,107 @@ defmodule Cog.Command.Pipeline.Executor do
...
-  If the command is enforced then the permission check is skipped and the command
-  is executed.
-  """
-  def maybe_enforce(:timeout, %__MODULE__{current_bound: current_bound}=state) do
-    if enforce?(current_bound) do
-      {:next_state, :check_permission, state, 0}
-    else
-      {:next_state, :maybe_collect_command, state, 0}
-    end
+    Enum.each(destinations, &publish_response(response_fn(message, adapter), &1, state))
```
vanstee: Wish elixir had some better functional utilities to avoid stuff like the `response_fn` thing.

lib/cog/command/pipeline/executor.ex
kevsmith: Yes. We should also look into unifying the different types of requests and responses. We should open issues for both.

lib/cog/commands/filter.ex
```diff
@@ -55,9 +55,9 @@ defmodule Cog.Commands.Filter do
     do: state
 
   defp validate_matches(state, matches) do
-    case String.valid?(matches) do
-      true -> %{state | match: compile_regex(matches)}
-      false -> add_errors(state, :bad_match)
+    case Regex.compile(matches) do
```
vanstee: Are we compiling the regex twice now? I know this also happens in `compile_regex` after some manipulating.

lib/cog/commands/table.ex
```diff
@@ -20,7 +20,7 @@ defmodule Cog.Commands.Table do
   @cell_padding "  "
 
   def handle_message(req, state) do
-    %{"fields" => headers} = hd(req.options)
+    %{"fields" => headers} =req.options
```
vanstee: Missing a space after the `=`

lib/cog/queries/command.ex
```diff
@@ -43,6 +54,16 @@ defmodule Cog.Queries.Command do
   end
 
   @doc """
+  Retrieve all information about a command
+  """
+  def complete_command(bundle_name, command_name) do
+    named(bundle_name, command_name)
+    |> with_bundle
+    |> with_rules
+    |> with_options
```
vanstee: You can probably nuke these `with_...` funcs and replace it with `|> Repo.preload([:bundle, :rules, [options: :option_type]])` since I think all those queries are run separately whether or not they're part of the query or not.

test/integration/commands/alias_test.exs
```diff
@@ -294,19 +294,13 @@ defmodule Integration.Commands.AliasTest do
   end
 
   test "list all aliases with no matching aliases", %{user: user} do
-    alias_list = []
-
     response = send_message(user, "@bot: operable:alias ls \"my-*\"")
-    expected_response = Helpers.render_template("alias-ls", Enum.sort(alias_list))
-    assert response["data"]["response"] == expected_response
+    assert response["data"]["response"] == "Pipeline executed successfully, but no output was returned"
```
vanstee: We probably need to standardize on ending responses in a period or not.

vanstee:
The only two important comments were the ones about creating a request struct and whether or not we're rendering a template per item in the output list. Other than that this looks great. Seriously good work @christophermaier.

lib/cog/command/pipeline/executor.ex
kevsmith: Template per final output list item. Makes sense to me as it'll be uniform with how we're presenting inputs to commands.

kevsmith:
:100: @christophermaier !

If we can all agree the new template rendering logic is correct I'm :+1: to merge. Let's table the request/response struct to a new ticket.

lib/cog/command/pipeline/executor.ex
vanstee: I think that's fine. It is a little limiting since we can no longer rendering something like:

```
These are all the ec2 instances in the us-east-1 region:

i-al8f3a
i-1ab34b
i-9b90ea
```

But that should be fine in most cases I guess.

mpeck: I remember some conversation about commands being aware of where they are in the list of inputs during execution. So commands could collect the results and output a template like you suggest @vanstee at the end of execution.

vanstee: How does that work if the command doesn't render the template? If there's a way to do that while also preserving the list of json data we output, I'm all for it.

lib/cog/queries/command.ex
christophermaier: I don't think so... they build up a single query, without any interaction with the database at all. `Repo.preload` relies on being passed a model struct (or structs) so you can preload _after_ you've already got an instance of something.

Using the query preload, however, allows you to have composable queries.

lib/cog/command/command_resolver.ex
```diff
@@ -1,54 +1,73 @@
...
+          {:command, {c.bundle.name, c.name, c}}
+        %UserCommandAlias{pipeline: pipeline} ->
+          {:pipeline, pipeline}
+        %SiteCommandAlias{pipeline: pipeline} ->
+          {:pipeline, pipeline}
+        :not_found ->
+          :not_found
+        {:ambiguous, names} ->
+          {:ambiguous, names}
+      end
```
lhaskins: I'm not following how the anonymous func is getting the namespace and command name from the inputs passed. Maybe more comments about how this works.

lib/cog/command/pipeline/executor.ex
mpeck: I honestly don't know. I just seem to remember it coming up at some point. I imagined a scenario where a command would be invoked with some additional metadata; `:first`, `:last`. The command would then process it's input, append the results to it's collected outputs and return. If processing returns no results, for commands like filter, then the command just returns whatever outputs, if any, it's collected so far. When an input with `:first ` or `:last` is encountered we know to reset the output list. It would result in commands maintaining some state, but I don't think that would be terrible. And now that I think about it, we should be able to hide that bit from developers pretty easily if we need to.

Take what I say with a grain of salt though. I hadn't really thought about it. I'm totally making this up as I go based something that I may imagined us talking about :)

christophermaier: As I was working through the code, I thought our rendering was more complicated than this. I was surprised as anyone to see that this is what it boils down to.

The example you provide @vanstee implies that all that data is in a single map (which I think is what the mustache templates also seem to require). In the case that we have a `once` command at the end of a pipeline, if it consolidates everything into a single map and has a template that can handle that, it would render properly (`table` works like this, currently).

That said, this works for everything that we currently have tests for; if there are edge cases that are being missed, we'll need to add some tests first and then reason back to a good approach.

lib/cog/command/command_resolver.ex
mpeck: Piper invokes the anonymous function with the namespace and command name @lhaskins. If that helps at all?

lib/cog/command/pipeline/executor.ex
imbriaco: Let's say there's a pipeline that looks conceptually like this:
```
ec2-list --region=-us-east-1 | ec2-terminate --region=us-east-1 $instance_id
```

The `ec2-list` stage returns:

```
[
  { "instance_id": "i-al8f3a" },
  { "instance_id": "i-1ab34b" },
  { "instance_id": "i-9b90ea" }
]
```

The first two executions of `ec2-terminate` get run normally, but know that they are not processing the final input. In this case, rather than returning a real result to the executor, they store their result in a state store that the final execution can access and return an empty result which will not get rendered.

For demonstration purposes let's assume the following:

```
Execution: ec2-terminate --region=us-east-1 i-al8f3a
Last: false
Stored Result: { "status": "terminated", "instance_id": "i-al8f3a" }
Returned Result: []
```

```
Execution: ec2-terminate --region=us-east-1 i-1ab34b
Last: false
Stored Result: { "status": "terminated", "instance_id": "i-1ab34b" }
Returned Result: []
```

The final execution is called exactly the same way, but has an additional piece of information in the environment to tell it that it's the last item in the stage.

```
Execution: ec2-terminate --region=us-east-1 i-1ab34b
Last: true
```

The command would perform its normal work, generating the following result:

```
Result: { "status": "terminated", "instance_id": "i-1ab34b" }
```

It could then retrieve the results from the previous executions in the stage and could return the following, which could get rendered by a single template:

```
[
  { "status": "terminated", "instance_id": "i-al8f3a", "region": "us-east-1" },
  { "status": "terminated", "instance_id": "i-1ab34b", "region": "us-east-1" },
  { "status": "termianted", "instance_id": "i-9b90ea", "region": "us-east-1" }
]
```

christophermaier: @mpeck I think some of what you mention is in the works for the upcoming removal of the distinction between `once` and `multiple`, but there's gonna be more changes involved with that.

vanstee: @imbriaco Ah so the executor has it's accumulated output set to something like this before rendering the template?

```
[
  [
    { "status": "terminated", "instance_id": "i-al8f3a", "region": "us-east-1" },
    { "status": "terminated", "instance_id": "i-1ab34b", "region": "us-east-1" },
    { "status": "termianted", "instance_id": "i-9b90ea", "region": "us-east-1" }
  ]
]
```

Which means the template only gets rendered once with the entire output.

That's alright I guess, but wouldn't that break our nice "lists of json" API that we have everywhere? How would appending a call to `table` on the end of that pipeline work since it's expecting a list of json, not a list in a list full of json? (eg: `ec2-list --region=-us-east-1 | ec2-terminate --region=us-east-1 $instance_id | table --fields "status,instance_id"`).

I might just be missing something. @christophermaier maybe we can go over this in a bit more detail later today.

lib/cog/command/command_resolver.ex
christophermaier: @lhaskins `command_resolver_fn` returns a function that closes over the passed in user. This function is passed into the parser and is used as a callback to lookup the relevant information; the parser determines the bundle name and command name to be passed in based on the parsed text.

lib/cog/queries/command.ex
vanstee: Ah you're right. I have gone a similar route as you have (adding composable functions for preloads), but given the number of combinations you end up needing to write functions for it got a little out of hand and I just opted to keep the preloads near where I end up using that data, which ends up being in a pipelined `Repo.all |> Repo.preload` call. I also like that you can easily combine all preloads into a single list when using `Repo.preload`, but this seems totally fine for now.

lib/cog/command/pipeline/executor.ex
christophermaier: If a command returns a list of items, that list effectively gets flattened into the accumulated output list; it's just going to be a list of maps, not a list of maps-or-lists.

Adding a table call to the end of a pipeline will continue to work as usual, because it will be given this entire list of accumulated output and will turn all of that into a single output map.

mpeck: @christophermaier yeah, I was thinking those bits were coming later. I think I'm saying the same thing as @imbriaco. Maybe just not quite as well :)

lib/cog/command/command_resolver.ex
lhaskins: Ah! gotcha - a callback function that gets executed in Piper. Thanks!

Issue #275, opened 2016-02-20
Support parsing smart quotes


vanstee:

vanstee:
When telling the bot to execute a command, we should support parsing OS X's smart quotes as regular ascii quotes. Right now we error out, but we should just replace each instance of a smart quote with the ascii version.

PR #404, opened 2016-03-16
Replace smart quotes with ascii quotes

vanstee:
Closes https://github.com/operable/cog/issues/275


mpeck:
Do single quotes get smart quoted? And does that matter?

vanstee:
Oh good call. Let me add another line for that.

mpeck:
:+1:

PR #408, opened 2016-03-17
send a room map to send_message instead of just the id

### LINKED ISSUE #405: Editing a previous command to resend a command to the bot is broken

mpeck:
Cog was silently failing when a user tried to edit a command. The issue centered around us sending the channel id to send_message instead of a room map.

resolves #405

### DIFF
--- a/lib/cog/adapters/slack/rtm_connector.ex
+++ b/lib/cog/adapters/slack/rtm_connector.ex
@@ -108,7 +108,10 @@ defmodule Cog.Adapters.Slack.RTMConnector do
                               fn() -> handle_command(room, user_id, text, state) end}
                          end
     unless message == nil do
-      Slack.API.send_message(channel, message)
+      {:ok, room} = Slack.API.lookup_room(id: channel)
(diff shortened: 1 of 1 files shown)


lib/cog/adapters/slack/rtm_connector.ex
```diff
@@ -108,7 +108,10 @@ defmodule Cog.Adapters.Slack.RTMConnector do
                               fn() -> handle_command(room, user_id, text, state) end}
                          end
     unless message == nil do
-      Slack.API.send_message(channel, message)
+      {:ok, room} = Slack.API.lookup_room(id: channel)
```
christophermaier: It looks like `Slack.API.send_message/2` only requires an `id` key in the map, so you could skip the lookup and just send in `%{"id" => channel}` as `room`.

mpeck: I thought about that. But then I thought it would be odd to sometimes have the name and sometimes not have it.

PR #409, opened 2016-03-17
Peck/edited commands fix

### LINKED ISSUE #405: Editing a previous command to resend a command to the bot is broken

mpeck:
Cog was silently failing when a user tried to edit a command. The issue centered around us sending the channel id to send_message instead of a room map.

resolves #405

### DIFF
--- a/lib/cog/adapters/slack/api.ex
+++ b/lib/cog/adapters/slack/api.ex
@@ -15,8 +15,8 @@ defmodule Cog.Adapters.Slack.API do
     GenServer.start_link(__MODULE__, config, name: @server_name)
   end
 
-  def send_message(room, message) when is_binary(message) do
-    GenServer.call(@server_name, {:send_message, room, message}, :infinity)
+  def send_message(%{"id" => id}, message) when is_binary(message) do
+    GenServer.call(@server_name, {:send_message, id, message}, :infinity)
   end
 
   def lookup_user(key) when is_tuple(hd(key)) do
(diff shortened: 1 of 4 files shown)


kevsmith:
:+1:

lib/cog/adapters/slack/rtm_connector.ex
```diff
@@ -108,7 +108,10 @@ defmodule Cog.Adapters.Slack.RTMConnector do
                               fn() -> handle_command(room, user_id, text, state) end}
                          end
     unless message == nil do
-      Slack.API.send_message(channel, message)
+      {:ok, room} = Slack.API.lookup_room(id: channel)
+      room = %{"id" => room.id,
+               "name" => room.name}
+      Slack.API.send_message(room, message)
```
christophermaier: I think a better thing would be to return send a map with just the "id" key, and add a guard to `Slack.API.send_message` indicating that all it cares about is that key.

vanstee: Actually it'd be really nice if `send_message` just took the room map, but I think in the executor we deserialize json into a map with string keys so that's why we need to do that weird conversion on line 112.

christophermaier: @vanstee I meant that `Slack.API.send_message/2` just cares about the ID, which we already have, so doing a lookup is just busy-work here.

vanstee: Oh good point. :+1:

lib/cog/adapters/slack/api.ex
```diff
@@ -15,8 +15,8 @@ defmodule Cog.Adapters.Slack.API do
     GenServer.start_link(__MODULE__, config, name: @server_name)
   end
 
-  def send_message(room, message) when is_binary(message) do
-    GenServer.call(@server_name, {:send_message, room, message}, :infinity)
+  def send_message(%{"id" => id}, message) when is_binary(message) do
+    GenServer.call(@server_name, {:send_message, id, message}, :infinity)
```
mpeck: @christophermaier is this what you were thinking?

christophermaier: :+1:

PR #412, opened 2016-03-17
Decouple room lookup and membership check

### LINKED ISSUE #410: Cog crashes on mentions of unknown rooms (Slack)
Cog has difficulty when it encounters names of rooms it hasn't joined. More specifically, the Slack adapter can crash whenever a room Cog hasn't joined is used or mentioned. Further it doesn't matter if the room mention occurs in a bot command or not.

After some investigation it turns out the central problem is Cog's Slack adapter conflates room lookup with room membership. This is old logic dating back to Cog's very early days before output redirects were implemented and, so, has been Not Quite Right for a while now.

Fixing this means the following:

- [x] Break out room membership checks as a first class adapter API function named `is_room_writable?`
- [x] Modify the Slack and HipChat adapters to implement this new function
- [x] Remove all room membership checking logic from other places in the Slack adapter
- [x] Teach the pipeline executor to use this API when preparing pipeline redirects

kevsmith:
This commit separates room membership checks from room lookups. This is
achieved by adding a new public API function to the `Cog.Adapter`
behavior: `room_writeable?/1` which returns true when the bot is
permitted to send messages to a room, false when it's not, and an error
tuple for all other cases. All adapters except Slack always return true.

The Slack adapter returns true if the bot is a member of the room
otherwise false (or error if the API call fails). This allows us to
catch the case where a user redirects output to room(s) where their Cog
instance is not present.

The executor uses this check to detect invalid redirects.

Fixes #410

### DIFF
--- a/lib/cog/adapters/slack.ex
+++ b/lib/cog/adapters/slack.ex
@@ -16,15 +16,10 @@ defmodule Cog.Adapters.Slack do
 
   def room_writeable?(opts) do
     case lookup_room(opts) do
-      {:ok, room} ->
-        case Slack.RTMConnector.assigned_userid() do
-          {:ok, userid} ->
-            is_member?(room, userid)
-          error ->
-            error
-        end
-      error ->
-        error
+      {:ok, %{is_member: is_member}} ->
+        is_member
+      _ ->
+        true
     end
   end
 
@@ -40,7 +35,4 @@ defmodule Cog.Adapters.Slack do
     "Slack"
   end
 
-  defp is_member?(%{members: members}, userid),
-    do: Enum.member?(members, userid)
-  defp is_member?(_room, _userid), do: true
 end
--- a/lib/cog/adapters/slack.ex
+++ b/lib/cog/adapters/slack.ex
@@ -14,6 +14,20 @@ defmodule Cog.Adapters.Slack do
     Slack.API.lookup_direct_room(opts)
   end
 
+  def room_writeable?(opts) do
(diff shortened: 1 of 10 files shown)


lib/cog/adapters/slack.ex
```diff
@@ -14,6 +14,21 @@ defmodule Cog.Adapters.Slack do
     Slack.API.lookup_direct_room(opts)
   end
 
+  def room_writeable?(opts) do
+    case lookup_room(opts) do
+      {:ok, room} ->
+        case Slack.RTMConnector.assigned_userid() do
+          {:ok, userid} ->
+            IO.puts "\n\n\n#{inspect room}\n\n\n"
```
christophermaier: Leftover debug logging?

kevsmith: Argh. Fixing.

lib/cog/adapters/slack.ex
```diff
@@ -14,6 +14,20 @@ defmodule Cog.Adapters.Slack do
...
+      {:ok, room} ->
+        case Slack.RTMConnector.assigned_userid() do
+          {:ok, userid} ->
+            is_member?(room, userid)
+          error ->
+            error
+        end
+      error ->
+        error
+    end
```
vanstee: `with` is perfect for stuff like this:

```
with {:ok, room} <- lookup_room(opts),
     {:ok, user_id} <- Slack.RTMConnector.assigned_userid(),
     do: is_member?(room, user_id)
```

PR #424, opened 2016-03-18
Add ExCoveralls for test coverage reporting

christophermaier:
Run `make coverage` to run tests, producing a coverage report. An
overview will be printed to the terminal, but open
`cover/excoveralls.html` for a more detailed report.

Does not (yet?) post anything to coveralls.io.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ erl_crash.dump
 *.pem
 /data
 Mnesia.*
+/cover
--- a/Makefile
+++ b/Makefile
@@ -56,4 +56,9 @@ test-watch: export MIX_ENV = test
 test-watch: reset-db
 	mix test.watch $(TEST)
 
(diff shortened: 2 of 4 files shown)


vanstee:
We might not have the best coverage percentage yet, but it might be nice to have buildkite check the change in coverage and fail if it goes below some threshold.

imbriaco:
I'm :-1: on arbitrarily failing the build right now until we get more experience with these tools.

christophermaier:
Just to clarify, this is not wiring up anything to CI builds right now. Just adding it to be a source of information that devs can use to drive their work for now.

mpeck:
I think it's a great idea to get some of these tools integrated into our dev process. It's probably time that we start thinking more about these things. I'm in agreement with @imbriaco though regarding the ci bits though. :+1:

PR #425, opened 2016-03-18
Add credo for linting / style checking

christophermaier:
This just makes [credo](https://github.com/rrrene/credo) available for
developers to use locally via `mix credo`.

Over time, as we get more experience with the tool, we can
collectively converge around the exact checks and parameters that
define the "Operable Style". At that point (and _not before_!) we can
wire these checks into our CI system to fail the build if we write
code that violates our expectations.


mpeck:
I'm adamantly :+1: on this. Really think it's time we start exploring these tools.

Issue #353, opened 2016-03-08
Track chat users by internal ID, not (only) chat handle


christophermaier:
Chat platforms often give users a stable internal ID as well as a public "handle", which can change. When available, we should use the internal ID to resolve a Cog user instead of just the handle.

Without this, users that change their handle would need to make a corresponding change in Cog in order to continue using Cog. Additionally, another user could change their handle to the old handle of the first user and thus "inherit" their permissions.

mpeck:
Looks like this might be a duplicate of #305

vanstee:
Comment from @imbriaco in duplicate issue:

There was a good conversation in cogbot chat tonight (thanks @ptinsley!) regarding handles and the safety of using them as a primary identifier to map chat provider users to cog users. I think there is a big enough concern that we should consider moving from handles to provider-specific user IDs. In cases where there may not be a user ID (ahem, IRC) we will have to be creative and carefully document the tradeoffs we make.

The primary problem with handles is that they can be changed by users. In the case of HipChat, for example, it's particularly troublesome since two different users can have what appears to us via XMPP to be exactly the same handle as long as they're not in the same room at the same time.

I think we can keep using handles for provisioning users, we just look them up at handle create time and convert to permanent IDs. There's a short but acceptable race condition there.

vanstee:
Slack and HipChat both have ids we can use, so IRC will be a little more interesting.

Here's a pan for some tiny PRs to get this worked out:

* [ ] Add provider_user_id column (not null) to chat_handles table
* [ ] Include provider_user_id field when creating chat-handles via the api. The call will remain the same, but we'll ask the adapter for the provider_user_id before creating the record
* [ ] When looking up the rules for the user, force the call to be made with the id rather than the handle (for IRC the id will be the handle so those calls will still work).
* [ ] Ensure that HipChat id's are unique to HipChat and not just per room. If not, append room info to the provider_user_id.

vanstee:
PR: https://github.com/operable/cog/pull/426

PR #426, opened 2016-03-18
Use provider user id to lookup cog user

vanstee:
We needed to lookup a user and their permissions with the internal id their chat provider users, since handles can be changed (or swaped) leading to someone stealing old permissions. Here's the break down of what changed:

* [x] Add chat_provider_user_id column (not null) to chat_handles table
* [x] Include chat_provider_user_id field when creating chat-handles via the api. The call will remain the same, but we'll ask the adapter for the chat_provider_user_id before creating the record
* [x] When looking up the rules for the user, force the call to be made with the id rather than the handle (for IRC and test adapters the id will be the handle so those calls will still work).
* [x] Add `lookup_user` callback to adapters
* [x] Decrease ttl on slack caches
* [x] Rework alias commands to lookup user with provider_user_id

Closes https://github.com/operable/cog/issues/353

### DIFF
--- a/lib/cog/command/user_permissions_cache.ex
+++ b/lib/cog/command/user_permissions_cache.ex
@@ -83,16 +83,38 @@ defmodule Cog.Command.UserPermissionsCache do
 
   defp fetch_and_cache({adapter, username}=key, state) do
     Logger.info("Cache miss for #{inspect key}")
-    case Repo.one(Queries.User.for_handle(username, adapter)) do
-      nil ->
-        {:error, :not_found}
-      user ->
-        perms = Cog.Models.User.all_permissions(user)
-        value = {user, perms}
-        expiry = Cog.Time.now() + state.ttl
-        :ets.insert(@ets_table, {key, value, expiry})
-        {:ok, value}
+
+    with {:ok, external_user} <- fetch_user_from_adapter(adapter, username),
+         {:ok, internal_user} <- fetch_user_from_database(adapter, external_user.id) do
+      perms = Cog.Models.User.all_permissions(internal_user)
+      value = {internal_user, perms}
+      expiry = Cog.Time.now() + state.ttl
+      :ets.insert(@ets_table, {key, value, expiry})
+      {:ok, value}
     end
   end
 
+  def fetch_user_from_adapter(adapter, handle) do
+    user = with {:ok, adapter_module} <- Cog.adapter_module(adapter),
+                do: adapter_module.lookup_user(handle: handle)
+
+    case user do
+      {:ok, user} ->
+        {:ok, user}
+      {:error, _} ->
+        {:error, :not_found}
+    end
+  end
+
+  def fetch_user_from_database(adapter, chat_provider_user_id) do
+    user = Queries.User.for_chat_provider_user_id(chat_provider_user_id, adapter)
+    |> Repo.one
+
+    case user do
+      nil ->
+        {:error, :not_found}
+      user ->
+        {:ok, user}
+    end
+  end
 end
--- a/lib/cog/queries/user.ex
+++ b/lib/cog/queries/user.ex
@@ -9,6 +9,14 @@ defmodule Cog.Queries.User do
     preload: [:chat_handles]
   end
 
+  def for_chat_provider_user_id(chat_provider_user_id, provider) when is_binary(provider) do
+    from u in User,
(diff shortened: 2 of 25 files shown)


lib/cog/queries/user.ex
```diff
@@ -9,6 +9,14 @@ defmodule Cog.Queries.User do
     preload: [:chat_handles]
   end
 
+  def for_chat_provider_user_id(chat_provider_user_id, provider) when is_binary(provider) do
```
christophermaier: We've now got 3 queries in this module that use the same from and join clauses (as well as some of the where and preloads, too!) Any way we can consolidate / extract the common bits and use composable queries?

vanstee: Yeah good point. On it.

christophermaier: :+1: :confetti_ball:

imbriaco:
Summarizing a conversation in #dev on Cog Slack:

* Once this lands, `%ChatHandle.handle` is effectively just a cache for the user's symbolic handle at the chat provider. 
* `%ChatHandle.chat_provider_user_id` is used for access enforcement. 
* This PR can land as is, but a follow-up issue noting that `%ChatHandle.handle` may get out of sync with the provider needs to be created. We'll discuss how or if we want to deal with the case when a user changes their handle at the provider in that follow-up issue.

lib/cog/queries/user.ex
vanstee: Just realized we're still using handles to lookup aliases so another :+1: for this suggestion.

lib/cog/command/user_permissions_cache.ex
```diff
@@ -83,16 +83,38 @@ defmodule Cog.Command.UserPermissionsCache do
...
+    user = with {:ok, adapter_module} <- Cog.adapter_module(adapter),
+                do: adapter_module.lookup_user(handle: handle)
+
+    case user do
+      {:ok, user} ->
+        {:ok, user}
+      {:error, _} ->
+        {:error, :not_found}
+    end
+  end
```
kevsmith: At first blush I find this function hard to read. Maybe I need more time to fully warm up to `with`. I think the explicit flow `case` statements display are more obvious:

```elixir
case Cog.adapter_module(adapter) do
  {:ok, mod} ->
    case mod.lookup_user(handle: handle) do
      {:error, _} ->
        {:error, :not_found}
      user ->
        user
    end
  error ->
    error
end
```

vanstee: The reason I used with was to avoid duplicating the match that swaps out the error for the more generic `{:error, :not_found}`.

The code would be: 

```
case Cog.adapter_module(adapter) do
  {:ok, mod} ->
    case mod.lookup_user(handle: handle) do
      {:ok, user} ->
        {:ok, user}
      {:error, _} ->
        {:error, :not_found}
    end
  error ->
    {:error, _} ->
      {:error, :not_found}
end
```

It is a tiny bit more clear, but I think the `with` version is more maintainable.

vanstee: Also I guess with the `with` version, you can quickly look to see what the possible return values would be: `{:ok, user}` or `{:error, :not_found}`.

kevsmith:
I'm seeing two warnings from this branch when I run `mix ecto.migrate`:

```
_build/dev/lib/cog/priv/repo/migrations/20160317183129_add_chat_provider_user_id_to_chat_handles.exs:5: warning: unused alias ChatHandle
_build/dev/lib/cog/priv/repo/migrations/20160317183129_add_chat_provider_user_id_to_chat_handles.exs:4: warning: unused alias Repo
```

We should clean these up before merging.

vanstee:
@kevsmith Good catch with the warnings in the migration. Fixing.

lib/cog/command/user_permissions_cache.ex
kevsmith: Fair enough. I need more time with `with` :)

kevsmith:
LGTM modulo the warnings mentioned above. Agree w/@imbriaco wrt ticket to track handling out-of-sync handle names.

vanstee:
This kept getting bigger and bigger. Gonna break out two issues once this is merged:

* Figure out how to start using `ChatHandle.handle` as a cache rather than live data
* Lookup the user in a single place when a request first comes in and roll it into the request object that is sent to commands

PR #429, opened 2016-03-18
Peck/command template rendering errors

### LINKED ISSUE #397: Handle command template rendering errors
The pipeline executor will crash if rendering a template throws an error. Instead it should log the error, along with contextual information about the error (pipeline id, requesting user, template and bundle which triggered the error), and return a human-friendly error message to the user.

mpeck:
Rescue's template rendering crashes and returns a message to the user.

resolves #397


lib/cog/command/pipeline/executor.ex
```diff
@@ -221,12 +221,15 @@ defmodule Cog.Command.Pipeline.Executor do
...
 
-    message = Enum.map_join(output, "\n", fn({data,template}) ->
-      render_template(command, adapter, template, data)
-    end)
-
-    Enum.each(destinations, &publish_response(response_fn(message, adapter), &1, state))
-    {:stop, :shutdown, state}
+    case render_templates(output, command, adapter) do
+      {:error, {error, template, adapter}} ->
+        msg = "There was an error rendering the template '#{template}' for the adapter '#{adapter}'."
```
christophermaier: Useful to include the rendering error in the message back to the user?

lib/cog/command/pipeline/executor.ex
```diff
@@ -387,6 +390,25 @@ defmodule Cog.Command.Pipeline.Executor do
   defp default_template(_),
     do: "raw"
 
+  defp render_templates(command_output, command, adapter) do
+    rendered = Enum.reduce_while(command_output, [], fn({data, template}, acc) ->
+      try do
+        {:cont, [render_template(command, adapter, template, data) | acc]}
```
christophermaier: Just a stylistic thing, but I'd probably separate these out to make it a little easier to read:

```elixir
rendered = render_template(command, adapter, template, data)
{:cont, [rendered|acc]}
```

christophermaier:
:+1: aside from a few small comments

test/integration/command_test.exs
```diff
@@ -13,6 +13,11 @@ defmodule Integration.CommandTest do
     assert response["data"]["response"] == "req-opt response"
   end
 
+  test "running a command with a bad template", %{user: user} do
+    response = send_message(user, "@bot: operable:bad-template")
+    assert response["data"]["response"] == "@vanstee Whoops! An error occurred. There was an error rendering the template 'badtemplate' for the adapter 'test': %Protocol.UndefinedError{description: nil, protocol: String.Chars, value: %{\"bad\" => %{\"foo\" => \"bar\"}}}"
```
christophermaier: Any way we can produce a more informative error message than a `Protocol.UndefinedError`?

kevsmith: Looks like a missing `inspect` when we're building the error.

kevsmith:
LGTM modulo above comment re: `inspect`.

test/integration/command_test.exs
mpeck: That's actually why I left it out the first time. I wasn't sure how useful it would be. We get the `Protocol.UndefinedError` when we try to serialize a map. That's actually the only thing I could find that causes us to crash when we render a template. I could match on `Protocol.Undefined` and return something like, "Error serializing value: #{value}". And then have a default case that just inspects the error. Thoughts?

mpeck: @kevsmith, we are inspecting when we build the error. That's just the error that bubbles up from FuManchu. https://github.com/operable/cog/pull/429/files#diff-aa3cd5dc48d292cc28d14bc42612addeR226

christophermaier: @vanstee any thoughts on this?

vanstee: This is undefined behavior as far as the mustache spec is concerned, since we're trying to interpolate a map. Ideally, we would just raise an error like `FuManchu.UnserializableContextError` with the value of context and some line number details and such. Does that sound reasonable?

mpeck: Since that will require a change to FuManchu, can we close this for now, or should we wait?

vanstee: Seems like it should be a separate followup issue. :+1:

PR #434, opened 2016-03-18
Match correct atom for variable binding errors

### LINKED ISSUE #431: Fix bug during error when binding a variable in a once command invocation
We fail to print the error message we have for binding variables during a once command. Here's the stacktrace:

```
 {%Protocol.UndefinedError{description: nil, protocol: String.Chars,
   value: {:missing_key, "timestamp"}},
  [{String.Chars, :impl_for!, 1, [file: 'lib/string/chars.ex', line: 3]},
   {String.Chars, :to_string, 1, [file: 'lib/string/chars.ex', line: 17]},
   {Cog.Command.Pipeline.Executor, :plan_next_invocation, 2,
    [file: 'lib/cog/command/pipeline/executor.ex', line: 205]},
   {:gen_fsm, :handle_msg, 7, [file: 'gen_fsm.erl', line: 518]},
   {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 240]}]}}
```

Probably just need to add an `inspect` call around `msg`.

vanstee:
Closes #431

### DIFF
--- a/lib/cog/command/pipeline/executor.ex
+++ b/lib/cog/command/pipeline/executor.ex
@@ -188,7 +188,7 @@ defmodule Cog.Command.Pipeline.Executor do
                                        current_plan: nil,
                                        plans: plans,
                                        invocations: remaining}, 0}
-      {:error, {:not_found, var}} ->
+      {:error, {:missing_key, var}} ->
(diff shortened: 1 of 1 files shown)


christophermaier:
:+1:

Issue #371, opened 2016-03-11
cogctl should warn when it can't find .cogctl


kevsmith:

kevsmith:
`cogctl` should error out with an informative message when a user attempts to execute an action which requires authorization and no possible authorization method is found. Authorization methods available to the user are `$HOME/.cogctl`, the CLI flags `-u/--user` and `--pw`, and/or the `--profile` CLI flag.

**Example:**

Bob has just installed `cogctl` on his laptop. He doesn't have a `$HOME/.cogctl` file. He runs `cogctl bundles` and sees the following error message on stderr:

```
User credentials missing. Credentials can be given on the command
 line via the --user and --pw flags or set in $HOME/.cogctl. See 
https://github.com/operable/cog/wiki/blahblah for more information.
```

PR #435, opened 2016-03-18
Token controller handles missing or nil parameters

christophermaier:
Previously, the controller would crash if either the username or password were not provided, or were nil.

Related to #371


imbriaco:
This looks good to me.

kevsmith:
:+1:

PR #436, opened 2016-03-19
Require @ and # for Slack redirect destinations

### LINKED ISSUE #210: Require '@' and '#' sigils for Slack redirects
When a user or channel is explicitly mentioned in a Slack message (e.g. `@user`, `#channel`), the text we receive from Slack translates these into IDs. For instance, instead of seeing `"@user"`, we would see `"<@U12345678>"`, and similar for channels. If a non-existent user or channel is given, Slack performs no such translation. Thus, if we see, e.g., a `"@does_not_exist"` string in a redirect position, we immediately know that there is no such user and that the redirect is invalid.

We currently allow redirects to leave off the sigils, in which case we attempt to resolve an appropriate Slack ID manually. First, we treat the string as though it were the name of a channel; if no such channel exists, we try it as a user name; if that fails, we error out.

This is problematic for several reasons:
1) It's ambiguous; does the user intend `"foo"` to refer to `@foo` or `#foo` (or did they accidentally hit enter early when trying to type `"foobar"`?
2) It could potentially conflict with our custom redirect targets `me` and `here`; If your Slack has any users or channels named "me" or "here", their owners are going to be very unhappy with all the spam.
3) It adds a fair bit of complexity to our Slack API module to perform the extra resolutions, not to mention additional latency due to the additional API requests.
4) It's not at all clear that resolving redirects without the `@` or `#` is even something that's likely to be used in real life, given that Slack itself requires the sigils for mentions already; that and autocomplete practically guarantee that users will always supply the sigil anyway.

Given this, we should simplify our lives and require explicit `@` and `#` sigils for all user and channel mentions in redirects for Slack.

christophermaier:
Removes the resolution code we had in place if a user redirected to
`foo`, as opposed to either `@foo` or `#foo`; now, you must be explicit.

Extracted the redirection error message out to a helper function to make
it a bit easier to reason about.

Fixes #210

### DIFF
--- a/lib/cog/adapters/slack/api.ex
+++ b/lib/cog/adapters/slack/api.ex
@@ -49,27 +49,13 @@ defmodule Cog.Adapters.Slack.API do
   NOTE: Only called (in this form) from the executor for finding redirects.
   """
   def lookup_room("@" <> user_name) do
-    {:ok, user} = lookup_user(handle: user_name)
-    lookup_direct_room(user_id: user.id)
-  end
-
-  def lookup_room("#" <> channel_name) do
-    lookup_room(name: channel_name)
-  end
-
-  # We do some snazzy recovery by not actually requiring the
-  # end-user to preface user or room names with "@" or "#",
-  # respectively (we can do this because we only accept room or
-  # user names in specific places in pipeline invocations).
-  #
-  # First, we try the string as a room name, and if that fails,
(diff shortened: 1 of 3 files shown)


kevsmith:
How difficult would it be to extend this to HipChat? If more than an hour or two let's open a ticket and track it separately. If not, I think it'd be good to fold it in to the PR.

Otherwise, :+1:

PR #437, opened 2016-03-19
Dialyzer!

christophermaier:
It looks like it works now, without littering the output with warnings
about struct field accesses!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ erl_crash.dump
 /data
 Mnesia.*
 /cover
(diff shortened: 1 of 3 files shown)


kevsmith:
How does one run dialyzer via mix?

christophermaier:
```
mix dialyzer.plt
mix dialyzer
```

vanstee:
:+1:

christophermaier:
Withdrawing for the time being, pending some more configuration investigations.

PR #441, opened 2016-03-21
Use consistent errors param

### LINKED ISSUE #442: Cogctl should use updates of CogAPI
The CogAPI is being updated and cogctl should be using more recent versions of the API. 

Consistent error handling needs to take place in order to accomodate the CogAPI changes.

lhaskins:
Before this change there was an inconsistent struct usage for errors. Some changesets used `:errors`, while others used `:error`. This PR changes things such that Cog uses `:errors`, which is consistent with the Ecto.changeset.

Fixes #442

### DIFF
--- a/web/views/bootstrap_view.ex
+++ b/web/views/bootstrap_view.ex
@@ -6,7 +6,7 @@ defmodule Cog.V1.BootstrapView do
   end
 
   def render("bootstrapped.json", _) do
-    %{bootstrap: %{errors: "Already bootstrapped"}}
+    %{errors: %{bootstrap: "Already bootstrapped"}}
   end
 
   def render("bootstrap.json", %{user: user}) do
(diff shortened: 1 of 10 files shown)


web/views/bootstrap_view.ex
```diff
@@ -6,7 +6,7 @@ defmodule Cog.V1.BootstrapView do
   end
 
   def render("bootstrapped.json", _) do
-    %{bootstrap: %{error: "Already bootstrapped"}}
+    %{bootstrap: %{errors: "Already bootstrapped"}}
```
vanstee: This might cause some breakage in cogctl, but it'd be nice to swap the nesting of this error map to be `%{errors: %{bootstrap: "Already bootstrapped"}}`

vanstee:
:heart: :yellow_heart: :green_heart: :blue_heart: Thanks for taking the time to clean all that up.

web/views/bootstrap_view.ex
lhaskins: great suggestion! I'm changing cogctl anyway, so I changed it :)

kevsmith:
:+1:

PR #444, opened 2016-03-21
Pipeline ID can be passed into the executor

### LINKED ISSUE #417: Pass unique ID into pipeline executor
Currently the executor creates a unique identifier for each pipeline, which is included in all audit log messages emitted by the pipeline.

Once pipelines can be triggered via webhooks, we'll want a way to correlate HTTP requests with all pipeline activity that is kicked off by them. We already generate a unique ID for HTTP requests that is included in all API audit events.

Instead of generating an ID in the executor, it should be passed in. Chat adapters should generate an ID for pipelines that come in via chat, and the future API endpoint for webhooks can pass the unique ID it will already have into pipelines that it initiates.

christophermaier:
Previously the executor itself would create a unique ID for the pipeline
being executed. In order to facilitate better tracking for pipelines
that are initiated from external events (e.g., webhooks), we'll need to
externalize this ID generation. (this will allow us to correlate HTTP
request processing with the resulting pipeline).

Now the pipeline initializer creates the ID and passes it as an argument
to the new executor process.

Fixes #417

### DIFF
--- a/lib/cog/adapter.ex
+++ b/lib/cog/adapter.ex
@@ -45,7 +45,8 @@ defmodule Cog.Adapter do
       end
 
       defp payload(sender, room, text) do
-        %{sender: sender,
+        %{id: UUID.uuid4(:hex),
+          sender: sender,
           room: room,
           text: text,
           adapter: name(),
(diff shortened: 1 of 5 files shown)


lib/cog/adapter.ex
```diff
@@ -45,7 +45,8 @@ defmodule Cog.Adapter do
       end
 
       defp payload(sender, room, text) do
-        %{sender: sender,
+        %{id: UUID.uuid4(:hex),
```
vanstee: Any reason to use `:hex` value here? The `:default` version might be nice if we were want to store these in postgres.

christophermaier: This is the format we use elsewhere for such IDs. Need to do another PR to consolidate those calls a bit.

PR #445, opened 2016-03-21
Make Token controller use errors instead of error

lhaskins:
This fixes the tests that are failing on master. The errors are stored using ":errors" instead 'f ":error".


christophermaier:
:+1:

PR #446, opened 2016-03-21
Allow an initial context to be passed to Executor

### LINKED ISSUE #418: Allow "seed" context to be passed into the executor
Currently, the executor effectively passes the first command invocation of a pipeline a `cog_env` that is equivalent to `[%{}]`; that is, a single empty map used as a binding context; this same empty map is passed as a `cog_env` to the command. This has the nice effect of unifying downstream processing for commands, regardless of their execution mode.

When pipelines are triggered from webhooks, they'll come with some kind of body (think Github PR events, for example). This should become the binding context / seed for the first command in the pipeline.

We should pass in the appropriate seed context into the executor instead of having the executor use its default. One way to help with this will be to create some kind of defined structure (a struct!) for pipeline requests, with the seed defaulting to `[%{}]`. Chat adapters wouldn't set this, but webhook adapters would.

christophermaier:
Previously, we had a hard-coded initial context inside the Executor. For
pipelines triggered by external events, we'll need a mechanism by which
to seed that initial context with arbitrary data (e.g., webhook request
bodies).

This lays the groundwork for that: if a request comes in with a map or
list stored under the "initial_context" key, that will be used. If
nothing is explicitly set, the same initial context as before (namely,
an empty map) will be used. Thus, all chat-adapter initiated pipelines
will continue to behave as before.

Created `store_with_template/2` and `remove_template/1` helper functions
to better encapsulate some manipulation operations so they're more
obvious.

Fixes #418

Also adds a few Credo-related cleanups in the Executor.

### DIFF
--- a/lib/cog/command/pipeline/executor.ex
+++ b/lib/cog/command/pipeline/executor.ex
@@ -80,10 +80,6 @@ defmodule Cog.Command.Pipeline.Executor do
 
   use Adz
 
-  # Provide an empty map for initial binding. The `nil` is for a
-  # template, and just to maintain the proper "shape" of the data
-  @seed_context [{%{}, nil}]
-
   def start_link(request),
     do: :gen_fsm.start_link(__MODULE__, [request], [])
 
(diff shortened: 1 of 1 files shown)


lib/cog/command/pipeline/executor.ex
```diff
@@ -280,7 +283,8 @@ defmodule Cog.Command.Pipeline.Executor do
                                        # there's nothing to collect
                                        state.output
                                      body ->
-                                       state.output ++ Enum.map(List.wrap(body), &({&1, resp.template}))
+                                       state.output ++ Enum.map(List.wrap(body),
+                                                                &store_with_template(&1, resp.template))
```
vanstee: Probably don't need a function for this:

```
&{&1, resp.temlpate}
```

lib/cog/command/pipeline/executor.ex
```diff
@@ -412,8 +416,8 @@ defmodule Cog.Command.Pipeline.Executor do
     end
```
vanstee: You might be able to avoid returning a function here, by changing the signature to `response_generator(room, message, adapter)` and using `&response_generator_fn(&1, message, adapter)` in all the places your currently using this.

vanstee:
:+1:

lib/cog/command/pipeline/executor.ex
christophermaier: It's simple, but the same logic is used elsewhere, too; I wanted to encapsulate that functionality rather than have it in multiple places in not-so-greppable form.

lib/cog/command/pipeline/executor.ex
christophermaier: I think you may be right. When this function was originally written, our rendering logic was a bit different; now we always pass the rendered `message` in, so the original reason for function has kind of gone away. I can do a follow-on PR to simplify / make it disappear.

Issue #451, opened 2016-03-22
Lookup cog user and send along with command request


vanstee:
There are multiple places where we're converting the user data we receive from a chat provider back into a cog user (sometimes multiple times in a single pipeline).

To cut down on these lookups and provider a uniform way to access this data, let's lookup the cog user before starting a pipeline and send that data to each command.

Related to https://github.com/operable/cog/pull/426

PR #453, opened 2016-03-22
Include user in command request

vanstee:
We now lookup the user (from chat provider id) in the executor which is then passed on to the user permissions cache and further on to commands. All commands now pull user info out of the request struct they are given.

Closes https://github.com/operable/cog/issues/451

### DIFF
--- a/lib/cog/command/user_permissions_cache.ex
+++ b/lib/cog/command/user_permissions_cache.ex
@@ -3,23 +3,20 @@ defmodule Cog.Command.UserPermissionsCache do
   use GenServer
   require Logger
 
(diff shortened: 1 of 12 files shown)


lib/cog/command/user_permissions_cache.ex
```diff
@@ -3,23 +3,20 @@ defmodule Cog.Command.UserPermissionsCache do
   use GenServer
```
christophermaier: We might want to rename this now... it's really just a permission cache

christophermaier:
:+1:

PR #457, opened 2016-03-23
Fix breakage due to incorrect api call and integer value

vanstee:


lib/cog/adapters/hipchat.ex
```diff
@@ -20,7 +20,8 @@ defmodule Cog.Adapters.HipChat do
   end
 
   def lookup_user(opts) do
-    HipChat.API.lookup_user(opts)
+    {:ok, user} = HipChat.API.lookup_user(opts)
+    {:ok, Map.put(user, :id, to_string(user.id))}
```
vanstee: This should be coerced when the `ChatProvider` model is created but this will work for now.

christophermaier: @vanstee can you elaborate on this some?

vanstee: The `ChatProvider.chat_provider_user_id` expects a string, while our JSON parsing of this id gives us an integer. It's not really a concern of this adapter so it doesn't fit here, but this is the easiest place to add it for now, since it's the only adapter that doesn't give us a string id.

lib/cog/adapters/hipchat/api.ex
```diff
@@ -14,7 +14,7 @@ defmodule Cog.Adapters.HipChat.API do
   end
 
   def lookup_room("@" <> handle) do
-    {:ok, user} = lookup_user(name: "@" <> handle)
+    {:ok, user} = lookup_user(handle: handle)
```
christophermaier: Small formatting thing I just noticed: can we remove the spaces between heads of the same function? The general rule for Erlang and Elixir is that there's only a newline between completely separate functions.

lib/cog/adapters/hipchat.ex
christophermaier: Can't that go into a changeset cast, though?

vanstee: D'oh totally forgot about that. Let me add it there instead.

lib/cog/adapters/hipchat/api.ex
```diff
@@ -42,8 +42,8 @@ defmodule Cog.Adapters.HipChat.API do
     GenServer.call(__MODULE__, {:lookup_user, id: user_id})
   end
 
-  def lookup_user(name: username) do
-    GenServer.call(__MODULE__, {:lookup_user, id: username})
+  def lookup_user(handle: username) do
```
christophermaier: We should document what keys we're expecting in the adapter behavior, with types preferably so Dialyzer can help us catch this kind of stuff earlier.

lib/cog/adapters/hipchat/api.ex
vanstee: I've been following the conventions from Elixir's source. One-liner and smaller functions logically similar functions are grouped together, but most of the time there are newlines between functions. I think the idea is to remove newlines to say "these functions are logically similar and could be replaced by a simple case statement", while adding a newline means "these functions use pattern matching but have totally different implementations and could potentially be given unique names".

vanstee: Glance through https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/string.ex to see what I mean.

lib/cog/adapters/hipchat/api.ex
vanstee: Good call.

lib/cog/adapters/hipchat/api.ex
christophermaier: Hmm... that's rather disappointing, then. It's definitely not that way in the Erlang world, which I think makes things overall more readable.

vanstee: I agree it's more readable in Erlang, but I'm not a fan of it in Elixir, since it feels too dense. I can't easily find where one function ends and another begins.

christophermaier: Also, FWIW, "these functions use pattern matching but have totally different implementations and could potentially be given unique names" seems like the wrong way to go. Sharing a name should imply a common use / intent, independent of implementation. If a thing is properly named, I should be able to change the implementation and everything else still "makes sense".

vanstee: In Elixir names and function signatures are combined to convey intent. For instance `lookup_user` is used for both looking up by handle and by id. In a language that doesn't use pattern matching I would expect there to be `lookup_user_by_handle` and `lookup_user_by_id` functions. So, in my mind, `lookup_user(handle: handle)` and `lookup_user(id: id)` feel like two independent functions. Where as something like `is_true(true)` and `is_true(false)` feel like two pieces of a single function and would be grouped together without newlines separating them.

I'm ok with whatever we decide as a group though. I'm not even sure if how I interpreted Elixir's style as the reason it was written that way or not.

christophermaier: I think the operative bit there is "in a language that doesn't use pattern matching" :smile: 

With pattern matching, we can be a lot more efficient at capturing our intent. "Here's an ID; give me the user, please", rather than "I would like you to retrieve a user for me by this method, please". We can be declarative, and our designs can take advantage of that.

lib/cog/models/chat_handle.ex
```diff
@@ -19,8 +19,16 @@ defmodule Cog.Models.ChatHandle do
   detail_fields [:id, :handle, :user, :chat_provider, :chat_provider_user_id]
 
   def changeset(model, params \\ :empty) do
+    params = coerce_chat_provider_user_id(params)
+
     model
     |> cast(params, @required_fields, @optional_fields)
   end
 
+  defp coerce_chat_provider_user_id(%{"chat_provider_user_id" => chat_provider_user_id} = params),
```
christophermaier: I was initially thinking something along the lines of https://github.com/operable/cog/blob/51546b25be6e503913e8994d0afff3243336072c/lib/cog/models/user.ex#L81-L99 and putting that into the pipeline for the changeset. That way you'd get an error message if it failed.

vanstee: Me to. But, we never get past the `cast` which creates the changeset that we can add values to, because the integer and string types don't match up. I figured `cast` would try to coerce the value, but it doesn't. So we have to coerce on our own first and then push it into the `cast` call. The other option would be to use a virtual field with an integer type, but we also have `id`s that are strings so that wouldn't work well either. Really the cast call should allow you pass in functions to use for type conversions.

christophermaier: Fair point. Add some comments about why we're doing it this way and :shipit:

Issue #382, opened 2016-03-15
Embed permissions data in Role API


lhaskins:
Permission information should be included in the role API such that calls to 
`/v1/roles` and `/v1/roles/:id` (for consistency) return permissions data for each role in the format:
```
{
  "id" : "4eba6180-c467-490f-807e-3fd54e46cf1e",
  "name" : "superuser",
  "permissions" : {
    "operable": [
      {
        "id" : "af5f3403-f151-4f83-aafe-e7ccac810604",
        "name" : "manage_commands"
      },
      {
        "id" : "e79cc110-c5fa-4635-96a8-d95d55df11ea",
        "name" : "manage_permissions"
      },
      {
        "id" : "b78383ed-4632-430c-b518-3b8cee1d5ff3",
        "name" : "manage_roles"
      },
      {
        "id" : "1086addd-2492-472e-9af2-7e5526a864cc",
        "name" : "manage_users"
      }
    }
  }
}
```

lhaskins:
### Motivation
Since roles is a collection of permissions, we should include the permissions that are associated with a particular role from the API.

### Objective
A single call to `/v1/roles` or `/v1/roles/:id` will return the role name, id, and permissions associated with the role.

### Research
#### What impact will this have in cogctl?
When listing roles, cogctl currently only lists the id and role name. If we want to list the permissions for each role, I recommend including a `roles info` call where the table layout would look like the following:
```
ID          123456789087654321
Name  MyRole

Permissions
NAMESPACE         NAME                                       ID
operable                    manage_commands         987654321234567890
operable                    manage_roles                      123456789098765432
site                                justForMe                              345678909876543210
```

The `roles` command can still just list the names and ids. We can leave it up to the `info` subcommand to display the permissions.

#### What does this mean concerning permissions and how they are displayed - for consistency?

####What else would need to possibly change?

### Plan
The Role model currently included the permissions, but does not return the permission data in the detail fields. This will need to change to include the permissions.

In the controller's `show` and `list` functions, we should change how the role is being rendered such that the output for each role reflects the following structure:
``` 
{
  "id" : "4eba6180-c467-490f-807e-3fd54e46cf1e",
  "name" : "superuser",
  "permissions" : {
    "operable": [
      {
        "id" : "af5f3403-f151-4f83-aafe-e7ccac810604",
        "name" : "manage_commands"
      },
      {
        "id" : "e79cc110-c5fa-4635-96a8-d95d55df11ea",
        "name" : "manage_permissions"
[...]

PR #459, opened 2016-03-23
Render different formatting for permissions in the Role and Permissio…

lhaskins:
This PR
* adds the show api to the Role controller
* a new permission format
```
%{permissions: [%{namespace: "namespace1",
                               id: "some id",
                               name: "some permission"},
                              %{namespace: "namespace2",
                                  id: "some other id",
                                  name:"some other permission"}]
}
```
* adds permissions to the Role model

Fixes https://github.com/operable/cog/issues/382

### DIFF
--- a/lib/cog/models/ecto_json.ex
+++ b/lib/cog/models/ecto_json.ex
@@ -89,5 +89,4 @@ defmodule Cog.Models.EctoJson do
     non_struct
   end
 
-
 end
--- a/test/controllers/v1/permission_controller_test.exs
+++ b/test/controllers/v1/permission_controller_test.exs
@@ -54,11 +54,10 @@ defmodule Cog.V1.PermissionControllerTest do
   test "creates and renders resource when data is valid", %{user: user} do
     conn = api_request(user, :post, "/v1/permissions",
                        body: %{"permission" => %{name: "test_perm"}})
-    recv_perm = json_response(conn, 201)
-    assert recv_perm["permission"]["name"] == "test_perm"
-    id = recv_perm["permission"]["id"]
-    assert id != nil
-    assert Repo.get_by(Permission, id: id)
+    recv_perm = json_response(conn, 201)["permission"]
+    assert recv_perm["name"] == "test_perm"
+    assert recv_perm["id"] != nil
+    assert Repo.get(Permission, recv_perm["id"])
   end
 
   test "does not create resource and renders errors when attempting to create an existing permission", %{user: user} do
@@ -71,49 +70,34 @@ defmodule Cog.V1.PermissionControllerTest do
   test "shows chosen resource regardless of namespace", %{user: user} do
     # random namespace
     rand_perm = permission("joe:test_perm") |> Repo.preload(:namespace)
-    rand_namespace = rand_perm.namespace
     conn = api_request(user, :get, "/v1/permissions/#{rand_perm.id}")
     assert json_response(conn, 200) == %{
-      "permission" => %{
-        "id" => rand_perm.id,
-        "name" => rand_perm.name,
-        "namespace" => %{
-          "id" => rand_namespace.id,
-          "name" => rand_namespace.name
-        }
-      },
-    }
+      "permission" =>
+        %{"namespace" => rand_perm.namespace.name,
+          "id" => rand_perm.id,
+          "name" => rand_perm.name}
+      }
 
     # site namespace
     site_perm = permission("site:test_perm") |> Repo.preload(:namespace)
-    site_namespace = site_perm.namespace
     conn = api_request(user, :get, "/v1/permissions/#{site_perm.id}")
     assert json_response(conn, 200) == %{
-      "permission" => %{
-        "id" => site_perm.id,
-        "name" => site_perm.name,
-        "namespace" => %{
-          "id" => site_namespace.id,
-          "name" => site_namespace.name
-        }
-      },
-    }
+      "permission" =>
+        %{"namespace" => site_perm.namespace.name,
+          "id" => site_perm.id,
+          "name" => site_perm.name}
+      }
 
 
     # embedded bundle namespace
     embedded_perm = permission("#{Cog.embedded_bundle}:test_perm") |> Repo.preload(:namespace)
-    embedded_namespace = embedded_perm.namespace
     conn = api_request(user, :get, "/v1/permissions/#{embedded_perm.id}")
     assert json_response(conn, 200) == %{
-      "permission" => %{
-        "id" => embedded_perm.id,
-        "name" => embedded_perm.name,
-        "namespace" => %{
-          "id" => embedded_namespace.id,
-          "name" => embedded_namespace.name
-        }
-      },
-    }
+      "permission" =>
+        %{"namespace" => embedded_perm.namespace.name,
+          "id" => embedded_perm.id,
+          "name" => embedded_perm.name}
+      }
 
   end
 
--- a/web/views/permissions_view.ex
+++ b/web/views/permissions_view.ex
@@ -1,0 +1,18 @@
+defmodule Cog.V1.PermissionView do
+  use Cog.Web, :view
+
+  def render("permission.json", %{permission: permission}) do
+    %{id: permission.id,
+      name: permission.name,
+      namespace: permission.namespace.name}
+  end
+
+  def render("index.json", %{permissions: permissions}) do
+    %{permissions: render_many(permissions, __MODULE__, "permission.json")}
+  end
+
+  def render("show.json", %{permission: permission}) do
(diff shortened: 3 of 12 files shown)


lib/cog/models/ecto_json.ex
```diff
@@ -52,6 +52,37 @@ defmodule Cog.Models.EctoJson do
...
+    case Keyword.fetch!(opts, :envelope) do
+      nil ->
+        data
+      envelope ->
+        Map.put(%{}, envelope, data)
+    end
+  end
+
+  def format_permissions(data, []), do: data
+  def format_permissions(data, %{__struct__: Cog.Models.Role}), do: data
```
vanstee: I'd opt for the fancier way to specify a struct here: `%Cog.Models.Role{}`

lib/cog/models/ecto_json.ex
```diff
@@ -52,6 +52,37 @@ defmodule Cog.Models.EctoJson do
...
+  def format_permissions(data, model) when is_list(model) do
+    [item | _] = model
```
vanstee: Might be cool to move this into the function head so you can match a list with at least one element in one go:

```
def format_permissions(data, [item | _]),
  do: format_permissions(data, item)
```

lib/cog/models/ecto_json.ex
```diff
@@ -52,6 +52,37 @@ defmodule Cog.Models.EctoJson do
...
+    format_permissions(data, item)
+  end
+  def format_permissions(data, _model) when is_list(data) do
+    perms = Enum.group_by(data, &(Map.get(&1, :namespace)))
```
vanstee: You can drop the parens around `&(Map.get(&1, :namespace))` when you're calling a function like this, but it's mostly a style thing.

lib/cog/models/ecto_json.ex
```diff
@@ -73,13 +104,24 @@ defmodule Cog.Models.EctoJson do
...
+    fields = Cog.Models.Permission.Namespace.__json_fields__(render_policy)
+    Enum.reduce(fields, "",
+      fn(field, acc) ->
+        if field == :name do
+          Map.get(struct, field)
+        else
+          acc
+        end
+      end)
+  end
```
vanstee: I think you can get by without the `reduce` here since you're overwriting `acc` anything you find a `field` called `:name`. Would this work the same?

```
fields = Cog.Models.Permission.Namespace.__json_fields__(render_policy)

if :name in fields do
  Map.get(struct, :name)
else
  ""
end
```

Still seems like a weird implementation but I don't really know the specifics here.

lib/cog/models/ecto_json.ex
```diff
@@ -89,5 +131,25 @@ defmodule Cog.Models.EctoJson do
...
+  defp render_field(field, struct, render_policy, acc),
+    do: Map.put(acc, field, do_render(Map.get(struct, field), render_policy))
+
+  defp scrub(index, list_maps, key) when is_map(list_maps) do
+    case is_list(list_maps[index]) do
+      true ->
+        {index, Enum.map(list_maps[index], fn (data_map) ->
+          {_value, scrubbed} = Map.pop(data_map, key)
+          scrubbed
```
vanstee: Since your dropping the `value`, `Map.delete` might be a tiny bit nicer here. You might be able to just send `&Map.delete(&1, key)` into the `Enum.map` actually.

test/controllers/v1/permission_controller_test.exs
```diff
@@ -55,10 +55,11 @@ defmodule Cog.V1.PermissionControllerTest do
...
     recv_perm = json_response(conn, 201)
-    assert recv_perm["permission"]["name"] == "test_perm"
-    id = recv_perm["permission"]["id"]
-    assert id != nil
-    assert Repo.get_by(Permission, id: id)
+    [perm|remaining] = recv_perm["permission"]["site"]
+    assert length(remaining) == 0
+    assert perm["name"] == "test_perm"
+    assert perm["id"] != nil
+    assert Repo.get_by(Permission, id: perm["id"])
```
vanstee: When getting something by the primary key you can use the shorter `Repo.get(Permission, perm["id"])`.

lib/cog/models/ecto_json.ex
```diff
@@ -52,6 +52,37 @@ defmodule Cog.Models.EctoJson do
...
+  end
+  def format_permissions(data, _model) when is_list(data) do
+    perms = Enum.group_by(data, &(Map.get(&1, :namespace)))
+    Enum.into(Map.keys(perms), %{}, &scrub(&1, perms, :namespace))
+  end
+  def format_permissions(data, _model) when is_map(data) do
+    {ns, scrubbed} = Map.pop(data, :namespace)
+    %{ns => [scrubbed]}
+  end
+  def format_permissions(data, _model), do: data
```
vanstee: I kinda thing `EctoJson` might be the wrong place for this, but I don't really have a good idea of where it would go. If we were using phoenix views a `permission` view would be the perfect place to put this kind of stuff. I wonder why we're not using those, hrm.

lib/cog/models/ecto_json.ex
```diff
@@ -52,6 +52,37 @@ defmodule Cog.Models.EctoJson do
...
+    data = do_render(model, policy)
+    |> format_permissions(model)
+
+    case Keyword.fetch!(opts, :envelope) do
+      nil ->
+        data
+      envelope ->
+        Map.put(%{}, envelope, data)
+    end
+  end
```
vanstee: For posterity, we discussed in chat that moving away from EctoJson and using phoenix views as we touch controllers should avoid any custom render functions like this.

kevsmith: Embedding model-specific behavior in `EctoJson` seems wrong; akin to embedding knowledge about child classes in a parent class.

lib/cog/models/ecto_json.ex
christophermaier: I strongly agree with @vanstee... `EctoJson` is not the place for this kind of functionality.

kevsmith:
If we're going to gradually adopt Phoenix views then I think we should start with this PR. No sense in adding modifications to `EctoJson` if it's deprecated and on the way out.

lib/cog/models/ecto_json.ex
lhaskins: :+1: I will back out these changes and start with using phoenix views for the permissions changes.

web/views/role_grant_view.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  defp render_permissions(permissions) do
+    Enum.map(permissions, fn(permission) ->
+        %{id: permission.id,
+          name: permission.name,
+          namespace: permission.namespace.name}
+    end)
+  end
+
+end
```
lhaskins: I wanted to set this such that it would use the same views as RoleView. Any ideas? The code is identical.

web/views/permissions_view.ex
```diff
@@ -0,0 +1,18 @@
...
+                    namespace: permission.namespace.name} }
+  end
+
+  def render("index.json", %{permissions: permissions}) do
+    %{permissions: Enum.map(permissions, fn(permission) ->
+        %{id: permission.id,
+          name: permission.name,
+          namespace: permission.namespace.name}
+    end)}
+  end
```
vanstee: Nice! This looks really nice.

It's probably not necessary for this PR but one nice thing I've seen people do is add a `render` function head for `permission.json` and then use that in all of the action specific functions (`index`, `show`, etc), so that the logic is the same.

vanstee: Actually on second thought, it might actually be pretty nice, since you can easily use it with `render_one` and `render_many`. Here's some untested code:

```
def render("permission.json", %{permission: permission}) do
  %{id: permission.id,
    name: permission.name,
    namespace: permission.namespace.name}
end

def render("index.json", %{permissions: permissions}) do
  %{permissions: render_many(permissions, __MODULE__, "permission.json")}
end

def render("show.json", %{permission: permission}) do
  %{permission: render_one(permission, __MODULE__, "permission.json")}
end
```

web/views/role_grant_view.ex
vanstee: Oh yeah I think you can set that in the controller like:

`render(conn, RoleView, :show, %{roles: roles})`

But since you really just want to override the view for all actions in that controller, I think you can use a plug in the controller:

```
plug :put_view, Cog.V1.RoleView
```

vanstee:
Really liked how this view stuff turned out @lhaskins. Good work :+1:

web/views/role_grant_view.ex
lhaskins: Great! Thanks!

PR #460, opened 2016-03-24
Fixes "user not found" functionality

christophermaier:
A recent refactoring around chat handle management inadvertently broke
the "user not found" functionality in the executor.

This commit uses the correct query name, and adds the chat handle
preload back to the query.


mpeck:
:+1:

vanstee:
Might be worth adding a test for this.

PR #461, opened 2016-03-24
Target redirects to specific adapters

### LINKED ISSUE #415: Allow redirects to have adapter prefixes
With upcoming work on responding to external events, we will have the ability to initiate pipelines from outside of chat for the first time. Up until this point, redirection targets were implicitly scoped to the adapter initiating the request; that is, a target of `#dev` was resolved to a Slack channel if the request came in via Slack.

When pipelines are initiated outside of chat, however, this falls apart. We need a way to target output so that it gets in front of the relevant people's eyes; there are no chat rooms in a webhook, after all.

To simplify things, we'll restrict Cog to supporting only a single chat adapter at a time (e.g., you can run Slack, HipChat, IRC, or _X_, but only one of them). Then the prefix of `"chat"` will resolve the target according to the active adapter.

Thus, a redirect of `chat:#dev` with Slack running would lookup the `#dev` channel using Slack's API.

To preserve existing functionality, a redirect target with a missing prefix will resolve according to the adapter that originated the request.

christophermaier:
This is groundwork for upcoming work to trigger pipelines in response to
external events (e.g., webhooks). In such a case, a pipeline will be
triggered from one adapter (e.g., an "HTTP adapter"), but output will
need to be directed to another (namely, the chat adapter).

Here, we introduce an optional URL-style notation for specifying
redirect targets. By redirecting to `chat://#room`, a response will be
formatted and delivered via the currently-configured chat adapter. The
redirect target itself will be resolved according to the chat adapter,
as well. Redirects may continue to be specified without the URL
prefix (e.g., `#room`), in which case they will be resolved and
dispatched according to the adapter the request originated from.

The special `me` and `here` redirect targets are always resolved
relative to the originating adapter.

As part of this work, the reply topic of an adapter is now exposed as a
behaviour function, and the executor discovers the appropriate reply
topics to send responses to, rather than relying on whatever is sent in
the original request.

This change necessitated a modification to how we test
adapters. Previously, we created a unique reply topic for each message
and waited to receive a message on that topic. Since we no longer look
at any reply topic sent to the executor, but rather ask adapter modules
directly what topics they listen on, this broke several tests. Each
pipeline executor process is given a unique ID from the adapter now,
however, which we now incorporate into the outgoing response message. By
matching this value, we can salvage the adapter test helper.

Additionally, the `ExecutorHelpers` module was removed, its functions
absorbed into the executor in order to consolidate and unify how
messages are sent out from the executor. Further streamlining of error
handling is currently scheduled.

(Also does a bit of cleanup and tweaking of `cog.ex`, as well)

Fixes #415


mpeck:
:+1:

PR #463, opened 2016-03-24
Use different alias for Slack adapter

christophermaier:
The Slack library we're using has the top-level module name `Slack`,
which is the same as the alias we were using for
`Cog.Adapters.Slack`. While this didn't cause code problems, it did make
it very confusing to read and understand the code, because it wasn't
immediately clear if a function came from the library or the adapter.


vanstee:
:+1: Nice tiny PR

PR #464, opened 2016-03-24
Try out VCR with simple Hipchat.API tests

vanstee:
This is a PR to pilot out tests that are mocked out with `exvcr`, just to see how everyone likes it before I rework all the integration tests. This test is a bit simpler because we don't need to worry about mocking things out when restarting the application like we do during some of the more intense integration tests, so it was perfect to start with.


test/support/vcr.ex
```diff
@@ -0,0 +1,38 @@
...
+
+        quote do
+          use_cassette(unquote(test_name)) do
+            unquote(context)
+          end
+        end
+      end
+    end
+  end
+end
```
vanstee: `end`s all the way down

christophermaier:
We should namespace the cassettes; e.g., instead of `test/fixtures/cassettes/test_looking_up_a_room.json` we use `test/fixtures/cassettes/hipchat/test_looking_up_a_room.json`, etc.

How do you go about recording new cassettes?

vanstee:
@christophermaier The library doesn't easily support subdirs like that from what I can tell, but in a future PR I have files namespaced like this: `test/fixtures/cassettes/elixir.cog.adapters.hipchat.apitest.test_looking_up_a_room.json`

I can clean it up a bit if you want, but this seems to be working for now as it's guaranteed to be unique based on how exunit requires a unique test name per case.

christophermaier:
@vanstee Bit of a bummer that you can't do subdirectories, but your scheme should work (until we get hundreds of thousands of cassettes, that is :stuck_out_tongue: )

:+1:

PR #465, opened 2016-03-24
Signature removal

### LINKED ISSUE #401: Remove vestiges of Carrier signatures
In #400, we stubbed out the calls in carrier that referred to `enacl` and removed that dependency. We need to go through cog, relay, spanner, and carrier and remove the rest of the signature related dead code.

mpeck:
Removes carrier signatures

resolves #401


lib/cog/adapter.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Cog.Adapter do
   defmacro __using__(opts) do
-    quote bind_quoted: [opts: opts] do
+    quote bind_quoted: [opts: opts], location: :keep do
```
vanstee: :+1:

vanstee:
:+1:

lib/cog/adapter.ex
christophermaier: :confetti_ball:

lib/cog/relay/relays.ex
```diff
@@ -109,19 +109,17 @@ defmodule Cog.Relay.Relays do
...
   defp process_introduction(%{"relay" => id,
-                              "public_key" => pub_key,
                               "reply_to" => reply_to}, state) do
     case CredentialManager.get(id, by: :id) do
       {:ok, nil} ->
-        creds = %Credentials{id: id, public: Base.decode16!(pub_key, case: :lower)}
+        creds = %Credentials{id: id}
         CredentialManager.store(creds)
-        Logger.info("Stored public key for Relay #{id}")
+        Logger.info("Stored credentials for Relay #{id}")
```
christophermaier: Perhaps something for a future PR, but the `Credential*` modules and wording seem out-of-place now. We should think of a better name that's more reflective of their current purpose.

mix.exs
```diff
@@ -67,7 +67,7 @@ defmodule Cog.Mixfile do
      {:hedwig, "~> 0.3.0"},
      {:gproc, "~> 0.5.0", override: true},
      {:html_entities, "~> 0.3.0"},
-     {:spanner, github: "operable/spanner", ref: "106b9371fa77c9dcfbdb758598cb3beda13c30f2"},
+     {:spanner, github: "operable/spanner", ref: "c5b4e457f3a91434752b091b101c71eaf5d51940"},
```
christophermaier: Might as well switch over to using `master` here, too, if we're going to do that elsewhere.

test/integration/alias_execution_test.exs
```diff
@@ -10,51 +10,51 @@ defmodule Integration.AliasExecutionTest do
 
   test "alias executes properly", %{user: user} do
     send_message(user, "@bot: operable:alias new my-alias \"echo my alias\"")
-    response = send_message(user, "@bot: my-alias")
+    response = send_message(user, "@bot: my-alias") |> Map.fetch!("response")
```
christophermaier: What else is in the response that comes back? Can we fold this `Map.fetch!/2` call into the implementation of `send_message/2`?

christophermaier: Or perhaps, since we do tests on the full reply map elsewhere, an `assert_message(response, expected_message)` helper?

lib/cog/relay/relays.ex
mpeck: I agree. There are some remnants of the credentials left over, but we need them to identify relays. I think we should kill credentials completely and create a new module for identification.

mix.exs
mpeck: Will do on merge. Didn't want to accidentally break the tests for the pr.

PR #466, opened 2016-03-25
add assert_message helper

mpeck:
Adds an `assert_message/2` helper to the tests. Should make writing and modifying tests a bit easier.


vanstee:
:+1:

PR #468, opened 2016-03-25
Overhaul executor error-handling

### LINKED ISSUE #416: Consolidate and improve pipeline error messaging
Currently, error messaging in the executor is all over the place. We should:

* Consolidate error handling in one place, rather than scattering it around the executor and related modules
* We should favor errors-as-data rather than errors-as-message-strings until we get to the very edge of the system. Pass error tuples around, with enough data as needed to generate actionable messages right before we send the response back to the user. This also facilitates a pattern-matching based approach to error handling (see, e.g., [here](https://github.com/operable/cog/blob/99b3b0d1a7bddda765f3dc7accbd803cdf277754/lib/cog/commands/permissions.ex#L229-L250))
* Provide context to the error (this is the important part of this issue). Once pipelines can be initiated outside of chat, we'll need a lot more information in errors in order for people to make sense of them. At a minimum we should provide:
  - the user that initiated the pipeline
  - timestamp when the pipeline was initiated
  - timestamp when the pipeline failed
  - the unique pipeline ID
  - the raw text of the pipeline as provided to the executor
  - the text of the bound command invocation that failed
  - the error message

Incidentally, this context will be helpful for human-initiated pipelines as well (think busy channels, and long-running pipelines)

christophermaier:
NOTE: This is built on `cm/redirect-prefix`, not `master`

Previously, error messages were generated in a number of places, making
it difficult to get a grasp on the variety of responses generated.

Additionally, error messages did not contain all that much context about
the error, something we'd need to address before allowing pipelines to
be triggered by external events.

Here, (almost) all error handling is consolidated into one place in the executor,
with lots of additional context added.

(Error messages for people lacking a Cog account, as well as for bad
redirects, are still handled as they were, since these were already
complex messages. Once we have error templates, everything will be able
to be consolidated even further.)

Tests have been updated, but in an admittedly less-than-satisfying
way. They are still strongly coupled to the exact output of the error
messages, but upcoming work to template error responses will,
interestingly enough, allow us to perform comparisons on data
structures, and not text, which will make the tests more robust in the
face of future changes.

Fixes #416


lib/cog/command/pipeline/executor.ex
```diff
@@ -677,35 +663,160 @@ defmodule Cog.Command.Pipeline.Executor do
...
+                       nil
+                   end
+
+    # Finally, the text of the specific error that occurred
+    error_text = """
+    The specific error was:
+
+        #{error_message}
+
+    """
```
mpeck: This is great. A lot better than what we were doing. Not suggesting this now, but I wonder if we could clean it up more bu leveraging FuManchu? Having templates for errors might make things more maintainable.

lib/cog/command/pipeline/executor.ex
```diff
@@ -677,35 +663,160 @@ defmodule Cog.Command.Pipeline.Executor do
...
     end
     publish_response(message,
                      request["room"],
-                     request["adapter"], # TODO: respect redirects?
+                     request["adapter"],
                      state)
   end
 
+  # Turn an error tuple into a message intended for the audit log
+  defp format_audit_message({:parse_error, msg}, state) when is_binary(msg),
```
mpeck: Would it make sense to move these error handling functions to another module?

test/integration/alias_execution_test.exs
```diff
@@ -48,7 +48,7 @@ defmodule Integration.AliasExecutionTest do
   test "alias expansion fails on infinite expansion", %{user: user} do
     send_message(user, "@bot: operable:alias new my-alias \"my-alias\"")
     response = send_message(user, "@bot: my-alias") |> Map.fetch!("response")
-    assert "@vanstee Alias expansion limit (5) exceeded starting with alias 'my-alias'." = response
+    assert_error_message_contains(response, "An error occurred. Alias expansion limit (5) exceeded starting with alias 'my-alias'.")
```
mpeck: You are gonna have some merge conflicts with the tests. I added a helper to simplify stuff a bit.

lib/cog/command/pipeline/executor.ex
christophermaier: I agree completely; see #469 and #470 :smiley:

lib/cog/command/pipeline/executor.ex
christophermaier: Possibly... not quite sure at this point yet. I'd probably wait until #469 and #470 are done.

My only concern is about putting the error processing farther away from where it's used (if these were error strings used across multiple modules, that'd be a different story).

test/integration/alias_execution_test.exs
christophermaier: Thanks for the heads up :+1:

lib/cog/command/pipeline/executor.ex
mpeck: Ha, great minds think alike :)

Issue #470, opened 2016-03-26
Extract error templates from the executor


christophermaier:
_Depends on #469_

Currently the Executor has three error messages that it hard-codes:

* the message shown to users that don't have either a Cog account or a registered chat adapter handle 
* the error message shown whenever redirect targets are found to be invalid
* the general error message shown whenever another failure in pipeline processing ocurrs (once #468 merges, that is)

Once issue #469 is complete, each of these should be extracted from the executor and turned into proper error templates.

vanstee:
Would there be any utility in moving the generic errors into a template as well, now that we can render json?

christophermaier:
Maybe? Those are mainly to create a human-readable message which we then stick into the larger error message with all the additional context. I'm not sure yet if they merit their own templates. How about we see what it's like with these three "big ones", and see what things look like afterward?

PR #471, opened 2016-03-26
Overhaul executor error-handling

### LINKED ISSUE #416: Consolidate and improve pipeline error messaging

christophermaier:
Previously, error messages were generated in a number of places, making
it difficult to get a grasp on the variety of responses generated.

Additionally, error messages did not contain all that much context about
the error, something we'd need to address before allowing pipelines to
be triggered by external events.

Here, (almost) all error handling is consolidated into one place in the executor,
with lots of additional context added.

(Error messages for people lacking a Cog account, as well as for bad
redirects, are still handled as they were, since these were already
complex messages. Once we have error templates, everything will be able
to be consolidated even further.)

Tests have been updated, but in an admittedly less-than-satisfying
way. They are still strongly coupled to the exact output of the error
messages, but upcoming work to template error responses will,
interestingly enough, allow us to perform comparisons on data
structures, and not text, which will make the tests more robust in the
face of future changes.

Fixes #416

### DIFF
--- a/lib/cog/command/pipeline/executor.ex
+++ b/lib/cog/command/pipeline/executor.ex
@@ -85,6 +85,7 @@ defmodule Cog.Command.Pipeline.Executor do
   alias Piper.Command.Ast
   alias Piper.Command.Parser
   alias Piper.Command.ParserOptions
+  alias Cog.Command.Pipeline.Plan
 
   use Adz
 
@@ -132,8 +133,7 @@ defmodule Cog.Command.Pipeline.Executor do
                                                raw_destinations: Ast.Pipeline.redirect_targets(pipeline),
                                                invocations: Enum.to_list(pipeline)}, 0}
       {:error, msg} ->
-        send_reply(msg, state)
-        fail_pipeline(state, :parse_error, "Error parsing command pipeline '#{state.request["text"]}': #{msg}")
+        fail_pipeline_with_error({:parse_error, msg}, state)
     end
   end
 
@@ -166,9 +166,7 @@ defmodule Cog.Command.Pipeline.Executor do
 
         {:next_state, :plan_next_invocation, %{state | destinations: destinations}, 0}
       {:error, invalid} ->
-        message = redirection_error_message(invalid)
-        send_error(message, state)
-        fail_pipeline(state, :redirect_error, "Invalid redirects were specified: #{inspect invalid}")
+        fail_pipeline_with_error({:redirect_error, invalid}, state)
     end
   end
 
@@ -176,6 +174,7 @@ defmodule Cog.Command.Pipeline.Executor do
                                                  output: previous_output,
                                                  user_permissions: permissions}=state) do
 
+    state = %{state | current_plan: nil}
     # If a previous command generated output, we don't need to retain
     # any templating information, because the current command now
     # controls how the output will be presented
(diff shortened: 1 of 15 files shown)


lib/cog/command/pipeline/executor.ex
```diff
@@ -284,6 +266,8 @@ defmodule Cog.Command.Pipeline.Executor do
                                    # one flat list
                                end
             {:next_state, :execute_plan, %{state | output: collected_output}, 0}
+          "error" ->
+            fail_pipeline_with_error({:command_error, resp}, state)
```
vanstee: :+1: I like having `error` cases below `ok` ones when possible.

lib/cog/command/pipeline/executor.ex
```diff
@@ -677,35 +663,160 @@ defmodule Cog.Command.Pipeline.Executor do
...
-  end
+  # Create a standard error message for pipeline-related
+  # errors. Eventually this will be templated.
+  defp user_error(error, state) do
+    id = state.id
+    started = Cog.Events.Util.ts_iso8601_utc(state.started)
+    initiator = sender_mention_name(state)
+    pipeline_text = state.request["text"]
+    error_message = format_user_error(error, state)
+    failing = case state.current_plan do
```
vanstee: Instead of doing this dance to figure out what failed during execution, would it be worth storing `failing` in `state`? Not sure if that's a great idea or not, but it does seem like it would clean up a bit of this logic.

lib/cog/command/pipeline/executor.ex
```diff
@@ -677,35 +663,160 @@ defmodule Cog.Command.Pipeline.Executor do
...
+
+    # If the failure happened during planning or execution, we'll have
+    # a bit more information to show
+    command_text = case failing do
+                     %Ast.Invocation{} ->
+                     """
+                     The pipeline failed planning the invocation:
+
+                         `#{failing}`
+                     """
```
vanstee: I don't really have an opinion about including a line at the end or not, but we should be consistent about it. See line 718. Also fwiw the end of a heredoc includes a new line.

lib/cog/command/pipeline/executor.ex
```diff
@@ -677,35 +663,160 @@ defmodule Cog.Command.Pipeline.Executor do
...
+  defp format_user_error({:disabled_bundle, %Cog.Models.Bundle{name: name}}, _state),
+    do: "Whoops! An error occurred. The #{name} bundle is currently disabled"
+  defp format_user_error({:timeout, %Cog.Models.Command{}=command}, _state) do
+    name = Cog.Models.Command.full_name(command)
+    "The #{name} command timed out"
+  end
+  defp format_user_error({:template_rendering_error, {error, template, adapter}}, _state),
+    do: "Whoops! An error occurred. There was an error rendering the template '#{template}' for the adapter '#{adapter}': #{inspect error}"
+  defp format_user_error({:command_error, response}, _state),
+    do: "Whoops! An error occurred. " <> (response.status_message || response.body["message"])
```
vanstee: :+1: Having all these error messages together will help us keep the wording consistent as well.

vanstee:
Looks good. We really needed this refactor.

lib/cog/command/pipeline/executor.ex
christophermaier: Possibly, yeah. I think this will probably shake out in the implementation of #470. 

Sound OK to leave it until then?

lib/cog/command/pipeline/executor.ex
christophermaier: D'oh!

Think this could wait until #470 as well?

lib/cog/command/pipeline/executor.ex
christophermaier: Yup. I'd love to take another pass through these and unify the wording.

PR #472, opened 2016-03-26
Streamline migration check code

christophermaier:
I was in the code for some other stuff and took the opportunity to do
some refactoring. This cleans up the migration check logic, provides
more detailed messages, and removes some unnecessary helper
functions (`abort_cog/0` and `log_message/2`).

### DIFF
--- a/lib/cog.ex
+++ b/lib/cog.ex
@@ -18,13 +18,18 @@ defmodule Cog do
     case Supervisor.start_link(children, opts) do
       {:ok, top_sup} ->
         # Verify the latest schema migration after starting the database worker
-        {sm_status, sm_message} = verify_schema_migration()
-        log_message(sm_status, sm_message)
-        if sm_status == :error do
-          abort_cog()
-        else
-          {:ok, top_sup}
-        end
+        #
+        # NOTE: This _may_ cause issues in the future, since we
+        # currently announce and install the embedded bundle during
+        # startup (i.e., before doing the schema migration check). If
+        # at some point we were to automatically upgrade the embedded
+        # bundle in such a way that required a database migration
+        # beforehand, the resulting error might be confusing.
+        #
+        # Consider doing this in a process that runs after the Repo
+        # comes up, but before anything else is done
+        :ok = verify_schema_migration!
+        {:ok, top_sup}
       error ->
         error
     end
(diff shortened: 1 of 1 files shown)


lib/cog.ex
```diff
@@ -18,13 +18,18 @@ defmodule Cog do
...
+        # NOTE: This _may_ cause issues in the future, since we
+        # currently announce and install the embedded bundle during
+        # startup (i.e., before doing the schema migration check). If
+        # at some point we were to automatically upgrade the embedded
+        # bundle in such a way that required a database migration
+        # beforehand, the resulting error might be confusing.
+        #
+        # Consider doing this in a process that runs after the Repo
+        # comes up, but before anything else is done
+        :ok = verify_schema_migration!
```
vanstee: If we raise here due to the migrations not being up to date, does that still stop the supervisor from starting up? Also, how does that change the error message printed?

lib/cog.ex
```diff
@@ -100,39 +105,43 @@ defmodule Cog do
...
+        end
     end
   end
 
-  defp migration_needed?() do
-    [last_file_version, _] = Path.join([:code.priv_dir(:cog), "repo", "migrations"])
+  # Determine if the database has had all migrations applied. If not,
+  # return the currently applied version and the latest unapplied version
+  @spec migration_status :: :ok | {:error, have, want} when have: pos_integer,
+                                                            want: pos_integer
```
vanstee: :+1:

vanstee:
Looks good. Was mainly curious about how this changes the failure scenario.

lib/cog.ex
christophermaier: Here's an example:

```

=INFO REPORT==== 28-Mar-2016::10:29:54 ===
    application: logger
    exited: stopped
    type: temporary
** (Mix) Could not start application cog: exited in: Cog.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (RuntimeError) The database schema is at version 20160325211544, but the latest schema is 20160322222933. Please perform a migration and restart Cog.
            (cog) lib/cog.ex:117: Cog.verify_schema_migration!/0
            (cog) lib/cog.ex:31: Cog.start/2
            (kernel) application_master.erl:273: :application_master.start_it_old/4
make: *** [run] Error 1
```
Though I'm just noticing that the "have" version doesn't appear to correspond to any migration version when you rollback (which I did here... Hmmm...)

Issue #474, opened 2016-03-28
Add the groups of a user


lhaskins:
The User model should include the group membership information when performing a show in the API

lhaskins:
### Motivation
Returning group membership information gives additional information using a single call instead of multiple calls when this information will be used the majority of the time when retrieving user information through the API.

### Objective
When performing a GET for User information, we should return group membership information as well.

### Research
Preload group_membership when performing a show. Should this be done for index as well?

### Plan
Preload group memberships when performing show and list for users.

### Tasks
- [ ] Preload membership info into User structure.
- [ ] Move to use view instead of EctoJson.
- [ ] Update/add tests.
- [ ] Update/add documentation.

PR #475, opened 2016-03-28
Add group memberships to users

lhaskins:
Adds the groups that a user is in when retrieving a user. Instead of using EctoJson, the User view is used for formatting the response.

Fixes https://github.com/operable/cog/issues/474

### DIFF
--- a/web/views/user_view.ex
+++ b/web/views/user_view.ex
@@ -1,0 +1,29 @@
+defmodule Cog.V1.UserView do
+  use Cog.Web, :view
+
+  def render("user.json", %{user: user}) do
+    %{id: user.id,
+      username: user.username,
+      first_name: user.first_name,
+      last_name: user.last_name,
+      email_address: user.email_address,
+      groups: render_groups(user.group_memberships)}
+  end
+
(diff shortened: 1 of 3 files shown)


web/views/user_view.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Cog.V1.UserView do
+  use Cog.Web, :view
+
+  def render("user.json", %{user: user}) do
+    %{id: user.id,
+      username: user.username,
+      first_name: user.first_name,
+      last_name: user.last_name,
+      email_address: user.email_address,
+      group_memberships: render_groups(user.group_memberships)}
```
imbriaco: Is there a reason to call this `group_memberships` rather than just `groups`?

imbriaco:
Aside from the `group_memberships` vs. `groups` question, this looks good to me.

web/views/user_view.ex
lhaskins: Not really. It was more about consistency with what field is accessed here. That said, I can easily change it to groups, if that's preferred.

PR #478, opened 2016-03-29
Relay management API

### LINKED ISSUE #403: Add Relay model and CRUD endpoints
As a first step in the upcoming Relay management improvements, we need to get a basic Relay model and associated CRUD API endpoints in place. 

The following schema sketch should illustrate the required fields.

* relays
  * id (uuid)
  * name (string)
  * token (string)
  * enabled (boolean)
  * connected (boolean)
  * last_connected (timestamp)
  * inserted_at (timestamp)
  * updated_at (timestamp)

* relay_tags
  * relay_id (uuid foreign key to relays)
  * name (string)

This issue requires fleshing out when we move to researching.

kevsmith:
Adds REST API w/supporting data model and migrations to manage relays and relay groups. Also lays foundation for Relay 2.0 changes.

Fixes #403.

### DIFF
--- a/lib/cog/commands/permissions.ex
+++ b/lib/cog/commands/permissions.ex
@@ -41,6 +41,7 @@ defmodule Cog.Commands.Permissions do
   permission "manage_users"
   permission "manage_roles"
   permission "manage_groups"
+  permission "manage_relays"
 
   rule "when command is #{Cog.embedded_bundle}:permissions with option[create] == true must have #{Cog.embedded_bundle}:manage_permissions"
   rule "when command is #{Cog.embedded_bundle}:permissions with option[delete] == true must have #{Cog.embedded_bundle}:manage_permissions"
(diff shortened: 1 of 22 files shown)


imbriaco:
We also need to support associating Bundles with RelayGroups. I'm happy to take a stab at that so you can keep moving on the other parts of go-relay. 

From an implementation point of view, perhaps we could manage the memberships beneath the Relay and Bundle resources much like we do for permissions grants under their related resources. That would make these endpoints the place where we manage the memberships:

```
/v1/relays/:id/relay_groups
/v1/bundles/:id/relay_groups
```

If that looks okay with you I can go down that path.

imbriaco:
To recap a face to face chat with @kevsemith, I'm going to take a stab at implementing the Bundle assignment part of this by adding a RelayGroupAssignment model that manages the relationship between bundles and relay groups.

imbriaco:
Looks good to me. Will tackle Bundle <-> RelayGroup assignment in a separate PR.

Issue #462, opened 2016-03-24
Add roles to group membership API


lhaskins:

lhaskins:
### Motivation
Returning roles using a single call instead of multiple calls when group information is retrieved through the API.

### Objective
When performing a GET for Group information, we should return roles that belong to the group as well.

### Research
Preload roles when performing a groups show or index. There is no pressing reason for not also sending this information when an index is performed. The groups that will be listed will be the "direct_group_memberships". 

### Plan
Preload roles when performing show and list for groups. Ensure that groups and users are also preloaded.
Returning structure will look like the following:
```
%{"id" => "some-id",
  "name" => "some-group"
  "members" => %{"groups" => [[%{"id" => "group-id",
                         "name" => "some-group-name"}],
                 "roles" => [[%{"id" => "role-id",
                          "name" => "some-rolename"}],
                 "users" => [%{"id" => "user-id",
                         "username" => "some-username"}]
      }
}
```
This format will be the same for group membership and the show and index APIs for Group.

Currently, 2 calls are being made from the cog-api-client to gather the membership information when a group show is performed. We should change this to only perform one call to the Cog controller.

### Tasks
- [ ] Preload roles, groups, and users info into Group structure.
- [ ] Move to use view instead of EctoJson.
- [ ] Update cogctl to display this information when performing the group info subcommand.
- [ ] Change the cog-api-client to not perform 2 calls, but only 1 group show/index.
- [ ] Update/add tests.

PR #480, opened 2016-03-29
Add groups, users, and roles to group#index and #show

lhaskins:
This PR adds roles to the groups#show and groups#index API calls and retrieves groups, roles and users in a single call. (Instead of also needing to make a call to the group membership API).

The structure returned looks like the following:
```
%{"id" => "some-id",
  "name" => "some-group"
  "members" => %{"groups" => [[%{"id" => "group-id",
                         "name" => "some-group-name"}],
                 "roles" => [[%{"id" => "role-id",
                          "name" => "some-rolename"}],
                 "users" => [%{"id" => "user-id",
                         "username" => "some-username",
                         "email_address" => "some-email_address",
                         "first_name" => "some-firstname",
                         "last_name" => "some-lastname"}]
      }
}
```

Fixes https://github.com/operable/cog/issues/462


kevsmith:
:+1:

PR #481, opened 2016-03-29
Use ExVCR for integration tests

vanstee:
Integration tests now no longer make HTTP requests every time they are run. The app still thinks we're making actual requests, but recorded responses are matched up during tests and mocked out as responses from `ibrowse` calls. Instructions on how to regenerate the tests are included in the README.


mix.exs
```diff
@@ -80,7 +80,7 @@ defmodule Cog.Mixfile do
      {:ex_doc, "~> 0.10", only: :dev},
      {:mix_test_watch, "~> 0.1.1", only: [:test, :dev]},
      {:excoveralls, "~> 0.5", only: :test},
-     {:exvcr, "~> 0.7", only: :test}
+     {:exvcr, github: "vanstee/exvcr", branch: "fix-cached-status-code", only: [:dev, :test]}
```
kevsmith: Can we move this fork to `operable`?

kevsmith:
:+1: modulo question re: `exvcr` fork.

README.md
```diff
@@ -48,6 +48,21 @@ With those installed, setup your computer with:
...
+future test runs. To regenerate these stubs by making actual HTTP
+requests delete the files you wish to regenerate and run the tests.
+
+For instance, to update all tests you could run this:
+
+    $ rm test/fixtures/cassettes/* && make test
+
+Just make sure you have the `TEST_SLACK=1` and `TEST_HIPCHAT=1`
+environment variables set.
+
```
lhaskins: What about adding new cassettes? Can you add some steps about that

mix.exs
vanstee: Good call

README.md
vanstee: Good point. New cassettes are created automatically when using the `use_cassette` macro or `AdapterCase`. I'll include a note about that.

PR #483, opened 2016-03-30
allow bundles to be assigned to relay groups

imbriaco:
This PR allows bundles to be associated with a group of relays through a RelayGroupAssignment association. From a high level, this involves:

* Add RelayGroupAssignments association between Bundles and RelayGroups
* Update RelayGroupMembershipController to support memberships and assignments and update routes accordingly
* Add Bundles to RelayGroupController#show and RelayGroupController#index
* Update BundlesController to use Phoenix views for rendering JSON and include RelayGroups

### DIFF
--- a/lib/cog/models/relay_group_assignment.ex
+++ b/lib/cog/models/relay_group_assignment.ex
@@ -1,0 +1,19 @@
+defmodule Cog.Models.RelayGroupAssignment do
+  use Cog.Model
+  @primary_key false
+
+  schema "relay_group_assignments" do
+    belongs_to :bundle, Cog.Models.Bundle, references: :id
+    belongs_to :group, Cog.Models.RelayGroup, references: :id
+  end
+
+  @required_fields ~w(bundle_id group_id)
+  @optional_fields ~w()
+
+  def changeset(model, params) do
+    model
+    |> cast(params, @required_fields, @optional_fields)
+    |> unique_constraint(:bundle_id, name: :relay_group_assignments_bundle_id_group_id_index)
+  end
+
+end
--- a/lib/cog/queries/bundles.ex
+++ b/lib/cog/queries/bundles.ex
@@ -4,21 +4,14 @@ defmodule Cog.Queries.Bundles do
 
   alias Cog.Models.Bundle
 
-  def bundle_details(id) do
+  def all do
     from b in Bundle,
-    where: b.id == ^id,
-    preload: [:commands, :namespace]
+    preload: [:commands, :relay_groups, :namespace]
   end
 
-  def bundle_summary(id) do
-    from b in Bundle,
-    where: b.id == ^id,
-    preload: [:namespace]
-  end
-
-  def bundle_summaries() do
-    from b in Bundle,
-    preload: [:namespace]
+  def for_id(id) do
+    all
+    |> where([b], b.id == ^id)
   end
 
   def bundle_id_from_name(name) do
@@ -27,4 +20,5 @@ defmodule Cog.Queries.Bundles do
     select: b.id,
     limit: 1
   end
+
 end
--- a/lib/cog/queries/relay_group_queries.ex
+++ b/lib/cog/queries/relay_group_queries.ex
@@ -1,31 +1,14 @@
 defmodule Cog.Queries.RelayGroup do
   use Cog.Queries
 
-  alias Cog.Models.RelayGroupMembership
-  alias Cog.Repo
-
   def all() do
     from rg in RelayGroup,
-    preload: [:relays]
+    preload: [:relays, :bundles]
   end
 
   def for_id(id) do
     all
     |> where([rg], rg.id == ^id)
-  end
-
-  def add_relays!(group_id, relay_ids) when is_list(relay_ids) do
-    Enum.reduce(relay_ids, [], fn(id, acc) ->
-      rgm = Repo.insert!(%RelayGroupMembership{relay_id: id,
-                                               group_id: group_id})
-      [rgm|acc]
-    end)
-  end
-
-  def remove_relays(group_id, relay_ids) when is_list(relay_ids) do
-    {count, _} = Repo.delete_all(from rgm in RelayGroupMembership,
-                                 where: rgm.group_id == ^group_id and rgm.relay_id in ^relay_ids)
-    count
   end
 
 end
--- a/lib/cog/repo.ex
+++ b/lib/cog/repo.ex
@@ -1,3 +1,13 @@
 defmodule Cog.Repo do
   use Ecto.Repo, otp_app: :cog
+
+  alias Cog.Queries.Repo, as: Queries
+
+  def exists?(model, id) do
+    case one!(Queries.count_by_id(model, id)) do
+      1 -> true
+      _ -> false
+    end
+  end
+
 end
--- a/lib/cog/support/model_utilities.ex
+++ b/lib/cog/support/model_utilities.ex
@@ -131,11 +131,11 @@ defmodule Cog.Support.ModelUtilities do
   end
 
   @doc "Create or retrieve a permission namespace with the given `name`"
-  def namespace(name) do
+  def namespace(name, bundle_id \\ nil) do
     namespace = Repo.get_by(Namespace, name: name)
     case namespace do
       nil ->
-        Repo.insert! %Namespace{name: name}
+        Repo.insert! %Namespace{name: name, bundle_id: bundle_id}
       _ ->
         namespace
     end
@@ -188,20 +188,6 @@ defmodule Cog.Support.ModelUtilities do
   end
 
   @doc """
-  Remove everything from the database.
-
-  As this removes bundles, too, we want to terminate any running
-  bundle processes, so they won't interfere with anything you might
-  reload later.
-  """
-  def clean_db! do
-    [Bundle, User, Group, Role, Namespace]
-    |> Enum.each(&Repo.delete_all/1)
-
-    Supervisor.terminate_child(Cog.Relay.RelaySup, Cog.Bundle.Embedded)
-  end
-
-  @doc """
   Creates a relay record
   """
   def relay(name, token, opts \\ []) do
@@ -230,6 +216,30 @@ defmodule Cog.Support.ModelUtilities do
     |> RelayGroupMembership.changeset(%{group_id: group_id,
                                         relay_id: relay_id})
     |> Repo.insert!
+  end
+
+  @doc """
+  Assigns a bundle to a relay group
+  """
+  def assign_bundle_to_group(group_id, bundle_id) do
+    %RelayGroupAssignment{}
+    |> RelayGroupAssignment.changeset(%{group_id: group_id,
+                                        bundle_id: bundle_id})
+    |> Repo.insert!
+  end
+
+  @doc """
+  Remove everything from the database.
+
+  As this removes bundles, too, we want to terminate any running
+  bundle processes, so they won't interfere with anything you might
+  reload later.
+  """
+  def clean_db! do
+    [Bundle, User, Group, Relay, Role, Namespace]
+    |> Enum.each(&Repo.delete_all/1)
+
+    Supervisor.terminate_child(Cog.Relay.RelaySup, Cog.Bundle.Embedded)
   end
 
 end
--- a/test/controllers/v1/relay_group_membership_controller_test.exs
+++ b/test/controllers/v1/relay_group_membership_controller_test.exs
@@ -7,13 +7,11 @@ defmodule Cog.V1.RelayGroupMembershipControllerTest do
   @relay_group_create_attrs %{name: "test-relay-group-1"}
 
   setup do
-    # Requests handled by the role controller require this permission
-    required_permission = permission("#{Cog.embedded_bundle}:manage_relays")
-
     # This user will be used to test the normal operation of the controller
     authed_user = user("cog")
     |> with_token
-    |> with_permission(required_permission)
+    |> with_permission("#{Cog.embedded_bundle}:manage_relays")
+    |> with_permission("#{Cog.embedded_bundle}:manage_commands")
 
     # This user will be used to verify that the above permission is
     # indeed required for requests
@@ -29,7 +27,6 @@ defmodule Cog.V1.RelayGroupMembershipControllerTest do
     add_relay_to_group(relay_group.id, relay.id)
     conn = api_request(requestor, :get, "/v1/relay_groups/#{relay_group.id}/relays")
     relays = json_response(conn, 200)["relays"]
-    assert length(relays) == 1
     assert relays == [%{"id" => relay.id,
                         "name" => relay.name}]
   end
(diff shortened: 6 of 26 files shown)


lib/cog/support/model_utilities.ex
```diff
@@ -232,4 +218,28 @@ defmodule Cog.Support.ModelUtilities do
...
+
+  @doc """
+  Remove everything from the database.
+
+  As this removes bundles, too, we want to terminate any running
+  bundle processes, so they won't interfere with anything you might
+  reload later.
+  """
+  def clean_db! do
+    [Bundle, User, Group, Role, Namespace]
```
kevsmith: I forgot about `clean_db!`. Probably need to add `Relay` to the list, yeah?

imbriaco: Good call. Added.

lib/cog/queries/relay_group_queries.ex
```diff
@@ -14,10 +15,24 @@ defmodule Cog.Queries.RelayGroup do
     |> where([rg], rg.id == ^id)
   end
 
+  def add!(:bundle, group_id, bundle_id),
```
christophermaier: I think we're mixing concerns in this module now.

These query modules should contain functions that generate Ecto queries (i.e., create data structures), not functions that actually _execute_ the queries. Ecto queries have the nice quality of composability, so we can provide query "building blocks" here that we can use elsewhere in various combinations. Having functions that also execute queries as a side effect destroys this composability.

These functions actually execute operations in the database, and so are actually defining the actual behavioral API we want to expose around bundles and their groups.

I've been meaning to write up something on shifting toward a more "classic DDD repository" approach for how we interact with the system, where we expose the actual behavioral API in a module / series of modules, and _that_ is how we manipulate the models in our domain. Basically, all database queries and `Cog.Repo` access occur behind that API. This encapsulates the behavior very clearly, ensures consistency, allows us more leeway in testing (simply substitute something else that implements that interface), and allows us to introduce things like caching transparently (as opposed to our past/current practice of calling caches directly in code).

We don't need to adopt this approach right now in this PR, but I would recommend at least that we keep actual database manipulation calls out of these query modules.

web/controllers/v1/relay_group_membership_controller.ex
```diff
@@ -12,24 +12,33 @@ defmodule Cog.V1.RelayGroupMembershipController do
...
-            Repo.rollback(:removing_relays_failed)
+            Repo.rollback(:remove_failed)
           else
-            Repo.preload(group, :relays)
+            group
+            |> Repo.preload(:bundles)
+            |> Repo.preload(:relays)
           end
         ids ->
           Repo.rollback({:unknown_relays, ids})
```
kevsmith: Is `:unknown_relays` still accurate given the controller's expanded functionality?

web/controllers/v1/relay_group_membership_controller.ex
```diff
@@ -41,10 +50,11 @@ defmodule Cog.V1.RelayGroupMembershipController do
         render(conn, Cog.V1.RelayGroupView, "show.json", relay_group: relay_group)
       {:error, reason} ->
         errors = case reason do
-                   {:removing_relays_failed} ->
+                   :remove_failed ->
                      "delete_failed"
-                   {:unknown_relays, ids} ->
-                     %{"not_found" => %{"relays" => ids}}
+                   {:unknown_, ids} ->
+                     name = String.to_atom(type <> "s")
```
kevsmith: Can we use real atoms here?

test/controllers/v1/relay_group_membership_controller_test.exs
```diff
@@ -48,6 +46,20 @@ defmodule Cog.V1.RelayGroupMembershipControllerTest do
...
+    relay_group = relay_group("test-relay-group-1")
+    assign_bundle_to_group(relay_group.id, bundle.id)
+    conn = api_request(requestor, :get, "/v1/relay_groups/#{relay_group.id}")
+    fetched = json_response(conn, 200)["relay_group"]
+    assert relay_group.id == fetched["id"]
+    assert relay_group.name == fetched["name"]
+    assert length(fetched["bundles"]) == 1
+    member = Enum.at(fetched["bundles"], 0)
+    assert member["id"] == bundle.id
+    assert member["name"] == bundle.name
```
christophermaier: I think all these asserts could be collapsed into one, using the match operator:

```elixir
assert %{"id" => ^relay_group_id,
         "name" => ^relay_group_name,
         "bundles" => [%{"id" => ^bundle_id,
                         "name" => ^bundle_name}]} = fetched
```
There is a downside in that things like `bundle.id` can't be in matches; you'd have to capture those in variables beforehand, but I think this gives a better sense of the shape of the data, and probably gives more usable error messages.

christophermaier: Similar for other asserts in this PR

web/views/bundles_view.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  def render("bundle.json", %{bundle: bundle}) do
+    %{id: bundle.id,
+      name: bundle.name,
+      namespace: bundle.namespace.name,
+      enabled: bundle.enabled,
+      relay_groups: render_relay_groups(bundle.relay_groups),
+      inserted_at: bundle.inserted_at,
+      updated_at: bundle.updated_at}
+  end
```
lhaskins: `commands` should also be listed here. The `show` endpoint expects commands from the detailed query.

test/controllers/v1/relay_group_membership_controller_test.exs
imbriaco: I like this. Unless I'm missing something though, we don't need to worry about the match operator and can just use `bundle.id` and friends on the LHS where we are creating the map that we compare with `fetched`. Am I missing something, @christophermaier?

lib/cog/queries/bundles.ex
```diff
@@ -27,4 +28,11 @@ defmodule Cog.Queries.Bundles do
     select: b.id,
     limit: 1
   end
+
+  def exists?(id) do
+    Repo.one!(from r in Bundle,
+              where: r.id == ^id,
+              select: count(r.id)) == 1
+  end
```
vanstee: This might even be worth adding to `Cog.Repo` since I've needed stuff like this before.

test/controllers/v1/relay_group_membership_controller_test.exs
christophermaier: @imbriaco if you're using a strict equality operator (`==`) that's true. If what I wrote out is the entirety of the result that comes back, I'd use that, which means we could use things like `bundle.id` directly. However, if there are other keys in either of those maps that are not currently being matched on, `==` won't work.

lib/cog/queries/relay_group_queries.ex
```diff
@@ -14,10 +15,24 @@ defmodule Cog.Queries.RelayGroup do
...
   def add_relays!(group_id, relay_ids) when is_list(relay_ids) do
     Enum.reduce(relay_ids, [], fn(id, acc) ->
-      rgm = Repo.insert!(%RelayGroupMembership{relay_id: id,
-                                               group_id: group_id})
+      rgm =
+        %RelayGroupMembership{}
+        |> RelayGroupMembership.changeset(%{relay_id: id, group_id: group_id})
+        |> Repo.insert!
       [rgm|acc]
     end)
```
vanstee: I know you didn't actually touch the outer code here but an `Enum.map` would be a bit simpler compared to `Enum.reduce`

lib/cog/queries/relay_group_queries.ex
```diff
@@ -28,4 +43,20 @@ defmodule Cog.Queries.RelayGroup do
...
   end
 
+  def add_bundles!(group_id, bundle_ids) when is_list(bundle_ids) do
+    Enum.reduce(bundle_ids, [], fn(id, acc) ->
+      rga =
+        %RelayGroupAssignment{}
+        |> RelayGroupAssignment.changeset(%{bundle_id: id, group_id: group_id})
+        |> Repo.insert!
+      [rga|acc]
+    end)
```
vanstee: Another place we could use `Enum.map` instead of `Enum.reduce`

test/controllers/v1/relay_group_membership_controller_test.exs
christophermaier: FWIW, I assumed that there was more that you weren't matching on when I wrote my initial comment.

imbriaco: Gotcha, that makes sense.

imbriaco: Ah, and I completely read over the `=` in your example as `==`. Derp.

lib/cog/queries/bundles.ex
```diff
@@ -18,7 +19,7 @@ defmodule Cog.Queries.Bundles do
 
   def bundle_summaries() do
     from b in Bundle,
-    preload: [:namespace]
+    preload: [:namespace, :relay_groups]
```
lhaskins: I'm wondering if both the bundle_details and bundle_summaries queries can go away since we now want to supply the commands list for both index and show. We can simply use `Repo.all` and `Repo.get` with preload of commands, namespace and relay_groups and call it a day.

lib/cog/queries/bundles.ex
christophermaier: How is this being used? Curious what the use case would be where access to the bundle model itself isn't also needed.

imbriaco: https://github.com/operable/cog/blob/imbriaco/bundle-relay-groups/web/controllers/v1/relay_group_membership_controller.ex#L74

christophermaier: OK, fair enough :smile:

test/controllers/v1/bundles_controller_test.ex
```diff
@@ -0,0 +1,47 @@
...
+           unauthed: unauthed_user]}
+  end
+
+  test "shows chosen resource", %{authed: requestor} do
+    bundle = bundle("test-1")
+    conn = api_request(requestor, :get, "/v1/bundles/#{bundle.id}")
+    assert %{"bundle" => %{"id" => bundle.id,
+                           "name" => bundle.name,
+                           "enabled" => bundle.enabled,
+                           "namespace" => bundle.name,
```
christophermaier: Why is there a namespace for bundles being exposed?

test/controllers/v1/relay_group_membership_controller_test.exs
```diff
@@ -7,13 +7,11 @@ defmodule Cog.V1.RelayGroupMembershipControllerTest do
...
 
   setup do
-    # Requests handled by the role controller require this permission
-    required_permission = permission("#{Cog.embedded_bundle}:manage_relays")
-
     # This user will be used to test the normal operation of the controller
     authed_user = user("cog")
     |> with_token
-    |> with_permission(required_permission)
+    |> with_permission("#{Cog.embedded_bundle}:manage_relays")
```
christophermaier: Is this permission being used in an embedded bundle command? If not, is it being explicitly added into the embedded bundle until it _is_ required by an embedded bundle command?

test/controllers/v1/relay_group_membership_controller_test.exs
```diff
@@ -101,5 +133,37 @@ defmodule Cog.V1.RelayGroupMembershipControllerTest do
...
+    bundles = for name <- ["test-bundle-1", "test-bundle-2", "test-bundle-3"] do
+      bundle(name)
+    end
+    bundle_ids = Enum.sort(Enum.map(bundles, &(&1.id)))
+    conn = api_request(requestor, :post, "/v1/relay_groups/#{relay_group.id}/assignment",
+                       body: %{"bundles" => %{"add" => bundle_ids}})
+    updated = json_response(conn, 200)["relay_group"]
+    assert updated["id"] == relay_group.id
+    assert updated["name"] == relay_group.name
+    assert length(updated["bundles"]) == length(bundle_ids)
```
christophermaier: This assert is effectively subsumed by the one that follows.

test/controllers/v1/relay_group_membership_controller_test.exs
```diff
@@ -101,5 +133,37 @@ defmodule Cog.V1.RelayGroupMembershipControllerTest do
...
+
+  test "removing bundles via REST endpoint", %{authed: requestor} do
+    relay_group = relay_group("test-relay-group-3")
+    bundles = for name <- ["test-bundle-1", "test-bundle-2", "test-bundle-3"] do
+      bundle(name)
+    end
+    for bundle <- bundles do
+      assign_bundle_to_group(relay_group.id, bundle.id)
+    end
+    bundle_ids = Enum.sort(Enum.map(bundles, &(&1.id)))
```
christophermaier:
```elixir
[keep | to_remove] = Enum.sort(Enum.map(bundles, &(&1.id)))
```
might be a little clearer, and lets you avoid the `hd` and `tl` calls.

test/controllers/v1/relay_group_membership_controller_test.exs
```diff
@@ -101,5 +133,37 @@ defmodule Cog.V1.RelayGroupMembershipControllerTest do
...
+    to_remove = tl(bundle_ids)
+    conn = api_request(requestor, :post, "/v1/relay_groups/#{relay_group.id}/assignment",
+                       body: %{"bundles" => %{"remove" => to_remove}})
+    updated = json_response(conn, 200)["relay_group"]
+    assert length(updated["bundles"]) == 1
+    [remaining] = updated["bundles"]
```
christophermaier: Then, you could just do
```elixir
assert [%{"id" => ^keep}] = updated["bundles"]
```
and skip the `assert length(...)` call above.

web/controllers/v1/relay_group_membership_controller.ex
```diff
@@ -12,27 +12,36 @@ defmodule Cog.V1.RelayGroupMembershipController do
...
-  def manage_membership(conn, %{"id" => id, "relays" => member_spec}) do
+  def manage_association(conn, type, %{"id" => id, "members" => member_spec}) do
     result = Repo.transaction(fn() ->
       group = Repo.get!(RelayGroup, id)
 
-      to_add = Enum.reduce(Map.get(member_spec, "add", []),
-                           %{found: [], not_found: []}, &lookup_or_fail/2)
-      to_remove = Enum.reduce(Map.get(member_spec, "remove", []),
-                           %{found: [], not_found: []}, &lookup_or_fail/2)
+      to_add = Enum.reduce(Map.get(member_spec, "add", []), %{found: [], not_found: []},
```
christophermaier: I don't like how `lookup_or_fail/3` embeds logic from this reduce call in it, making _this_ code a little less transparent.

A cleaner approach would be to encapsulate the entire reduce call into a function of its own, passing in the `type`, and the list of IDs. Doing something with `Enum.group_by/2` would be better still; just have the new version of `lookup_or_fail` return `:found` or `:not_found` and you get a bunch of stuff for free :smile:

web/views/bundles_view.ex
```diff
@@ -0,0 +1,45 @@
...
+  def render("index.json", %{bundles: bundles}) do
+    %{bundles: render_many(bundles, __MODULE__, "bundle.json", as: :bundle)}
+  end
+  def render("show.json", %{bundle: bundle}) do
+    %{bundle: render_one(bundle, __MODULE__, "bundle.json", as: :bundle)}
+  end
+
+  def render_relay_groups(%Association.NotLoaded{}) do
```
christophermaier: Doesn't this mean that we'll show different results to the user depending on which code path our data took to get here? Seems like we should always load the association if it's part of our API contract to show them.

web/views/bundles_view.ex
```diff
@@ -0,0 +1,45 @@
...
+    []
+  end
+  def render_relay_groups(groups) do
+    for group <- groups do
+      %{id: group.id,
+        name: group.name}
+    end
+  end
+
+  def render_commands(%Association.NotLoaded{}) do
```
christophermaier: Ditto for this.

web/views/relay_group_view.ex
```diff
@@ -17,17 +18,23 @@ defmodule Cog.V1.RelayGroupView do
...
   def render("relays.json", %{relay_group: relay_group}) do
-    %{relays: render_members(relay_group.relays())}
+    %{relays: render_relays(relay_group.relays())}
   end
 
-  defp render_members(%Association.NotLoaded{}) do
-    []
-  end
-  defp render_members(relays) do
+  defp render_relays(%Association.NotLoaded{}), do: []
```
christophermaier: Ditto here

web/views/relay_group_view.ex
```diff
@@ -17,17 +18,23 @@ defmodule Cog.V1.RelayGroupView do
...
+  defp render_relays(relays) do
     for relay <- relays do
       %{id: relay.id,
         name: relay.name}
     end
   end
 
+  defp render_bundles(%Association.NotLoaded{}), do: []
```
christophermaier: And here

lhaskins:
I'm guessing now that we have the APIs for displaying relay groups and bundles on a relay that there will be a issue for adding this to cogctl?

lib/cog/queries/bundles.ex
kevsmith: Also see https://github.com/operable/cog/blob/kevsmith/go-relay/lib/cog/bus_creds.ex

test/controllers/v1/bundles_controller_test.ex
imbriaco: That was a plain old bug. Removed.

web/views/bundles_view.ex
imbriaco: Both index and show delegate to `bundle.json`. The only difference is whether they return a single bundle or a list of bundles. I think that's mostly a guard to keep from blowing up if the association isn't loaded for some reason, but we should probably crash instead of returning a possibly inaccurate empty list of `relay_groups`.

I'll make that change.

lib/cog/queries/relay_group_queries.ex
imbriaco: I moved these queries to the Model, ala: `Cog.Models.Bundle.exists?(id)`

web/controllers/v1/relay_group_membership_controller.ex
kevsmith: fwiw I originally lifted most of this logic from `group_membership_controller.ex` for consistency and speed of implementation. 

https://github.com/operable/cog/blob/master/web/controllers/v1/group_membership_controller.ex#L21

test/controllers/v1/relay_group_membership_controller_test.exs
imbriaco: It looks like @kevsmith added it to `Cog.Commands.Permissions` when he initially implemented this.

imbriaco:
@lhaskins Yes, we need to add the relays and relay group management functions to cogctl. I'll create a stub issue for it.

imbriaco:
@christophermaier @vanstee I think I addressed basically all of the feedback. Would appreciate another quick once-over to make sure nothing jumps out and I'll plan to merge this in the morning. Thanks for the review.

lib/cog/repo.ex
```diff
@@ -1,3 +1,13 @@
 defmodule Cog.Repo do
   use Ecto.Repo, otp_app: :cog
+
+  alias Cog.Queries.Repo, as: Queries
+
+  def exists?(model, id, opts \\ [true: true, false: false]) do
```
vanstee: Is this true false thing used anywhere yet? If not, I'd opt to remove it and just return true/false.

imbriaco: Ah, that's not necessary anymore. That's because I wanted to be able to optionally return `:found` or `:not_found` but the code that needed it got brutally refactored. I'll rip that out.

lib/cog/queries/bundles.ex
```diff
@@ -4,21 +4,14 @@ defmodule Cog.Queries.Bundles do
...
-    where: b.id == ^id,
-    preload: [:namespace]
-  end
-
-  def bundle_summaries() do
-    from b in Bundle,
-    preload: [:namespace]
+  def for_id(id) do
+    all
+    |> where([b], b.id == ^id)
```
vanstee: If we can, it'd be nice to replace this query with calls to `Repo.get(Queries.Bundle.all, id)`

vanstee: Ah looks like it was copied from the pattern in other query modules, we probably need to refactor all of those away sometime soon.

lib/cog/models/relay_group_assignment.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Cog.Models.RelayGroupAssignment do
+  use Cog.Model
+  @primary_key false
+
+  schema "relay_group_assignments" do
+    belongs_to :bundle, Cog.Models.Bundle, references: :id
+    belongs_to :group, Cog.Models.RelayGroup, references: :id
```
vanstee: I think `references: :id` is the default so it probably isn't necessary here.

vanstee:
@imbriaco Looks good, just a couple of minor things that don't necessarily need fixing. Ready to merge whenever.

PR #484, opened 2016-03-30
Avoid unnecessary implementation of  macro

vanstee:


christophermaier:
_and there was great rejoicing_

PR #492, opened 2016-04-01
Simplify permission related APIs

imbriaco:
This PR significantly simplifies the behavior of permission related APIs in Cog. In the simplified model, users acquire a permission by belonging to a group that has been granted a role that has been granted some permissions.

This will cause some breakage in `cogctl` and `cog-api-client` when it is merged until they are updated to address the changes. The required updates are very minor since these changes largely remove endpoints, with just one exception for managing users within groups.

### Model Changes:

- Permissions cannot be granted directly to users or groups. They are only granted to roles.
- Roles cannot be directly granted to users. They are only granted to groups.
- Groups cannot contain other groups. Only users can be members of a group.

### API Changes: 

- **Groups API**:
  - (REMOVED) `/v1/groups/:group_id/membership`
  - (NEW) `/v1/groups/:group_id/users`
    - Behaves similarly to previous `/v1/groups/:id/membership` endpoint with the top-level members map replaced with the previously nested users map. **NOTE: The users map refers to users by email address rather than username**
    - JSON payload examples:
      - `{ "users": { "add": [ "mark@example.com", "kevin@example.com" ] }`
      - `{ "users": { "remove": [ "removed@example.com" ] }`
      - `{ "users": { "add": [ "mark@example.com", "kevin@example.com" ], "remove": [ "removed@example.com" ] }`
- **Permission API**:
  - Permissions may only be granted to roles; the ability to grant permissions directly to users and groups has been removed.
  - (REMOVED) `/v1/users/:user_id/permissions`
  - (REMOVED) `/v1/groups/:group_id/permissions`
- **Role API**:
  - (REMOVED) `/v1/users/:id/roles`

### DIFF
--- a/test/controllers/v1/permission_grant_controller_test.exs
+++ b/test/controllers/v1/permission_grant_controller_test.exs
@@ -2,6 +2,12 @@ defmodule Cog.V1.PermissionGrantController.Test do
   use Cog.ModelCase
   use Cog.ConnCase
 
+  # When these tests were written, permissions were allowed to be
+  # associated with users, groups, or roles. This model has been
+  # simplified to only allow permissions to be associated with roles,
+  # but the parameterized test scaffolding has been retained for
+  # the time being. More information about this structure is below.
+  #
   # All the permission granting/revoking API endpoints share similar
   # structure, and are all implemented by the same code. As such, the
   # tests for granting permissions to users and granting them to roles
@@ -67,7 +73,7 @@ defmodule Cog.V1.PermissionGrantController.Test do
   # For each kind of entity we want to test, we'll generate a whole
   # series of tests focused around testing the manipulation of
   # permissions on that kind of entity.
-  [:user, :role, :group] |> Enum.each(
+  [:role] |> Enum.each(
     fn(base) ->
 
       # Grant
--- a/test/controllers/v1/permission_grant_controller_test.exs
+++ b/test/controllers/v1/permission_grant_controller_test.exs
@@ -2,6 +2,9 @@ defmodule Cog.V1.PermissionGrantController.Test do
   use Cog.ModelCase
   use Cog.ConnCase
 
+  # TODO: Clean this up when the underlying data model is simplified
+  #       to match what's exposed in the API.
+
   # When these tests were written, permissions were allowed to be
   # associated with users, groups, or roles. This model has been
   # simplified to only allow permissions to be associated with roles,
--- a/web/controllers/v1/group_membership_controller.ex
+++ b/web/controllers/v1/group_membership_controller.ex
@@ -18,6 +18,13 @@ defmodule Cog.V1.GroupMembershipController do
     render(conn, "show.json", group: group)
   end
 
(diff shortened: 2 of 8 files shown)


test/controllers/v1/permission_grant_controller_test.exs
```diff
@@ -2,6 +2,12 @@ defmodule Cog.V1.PermissionGrantController.Test do
   use Cog.ModelCase
   use Cog.ConnCase
 
+  # When these tests were written, permissions were allowed to be
+  # associated with users, groups, or roles. This model has been
+  # simplified to only allow permissions to be associated with roles,
+  # but the parameterized test scaffolding has been retained for
+  # the time being. More information about this structure is below.
+  #
```
christophermaier: Are we leaving this in because we're going to add something soon that can use it again, or just for expediency? If the latter, can we put an issue on the backlog to remove it?

web/controllers/v1/group_membership_controller.ex
```diff
@@ -18,6 +18,13 @@ defmodule Cog.V1.GroupMembershipController do
     render(conn, "show.json", group: group)
   end
 
+  def manage_group_users(conn, %{"users" => user_spec}=params) do
+    params
+    |> Map.put("members", %{"users" => user_spec})
+    |> Map.delete("users")
+    |> fn(param) -> manage_membership(conn, param) end.()
```
christophermaier: This reads really weirdly. Any reason to not just edit `manage_membership/2` to get rid of the roles and groups instead?

test/controllers/v1/permission_grant_controller_test.exs
imbriaco: I left it there because we intentionally aren't changing the underlying data model. Also, I wanted to keep the tests as close to the original as possible to have the maximum amount of confidence that may changes to controllers and routes were correct and not just passing because I had changed the behavior of the tests. I'm open to a backlog item to refactor these, though.

web/controllers/v1/group_membership_controller.ex
imbriaco: Not really, it's just another instance of me trying to make the minimum possible changes so that I could add functionality back later if we change our minds. In this case I'm happy to just rename manage_group_users though since it's pretty small change.

lhaskins:
It may be worth getting the fixes for cogctl and CogApi before merging this. I think the resulting changes may be more than it seems...

web/controllers/v1/group_membership_controller.ex
imbriaco: Actually, `manage_memberships` is still used for granting roles to groups. I will double check that I haven't accidentally broken the tests to exercise that. I'll also pull the group related logic out of this function but I don't think there's more that needs to be done here to solve the problem this PR is trying to solve versus a general structural refactoring, which is out of scope.

christophermaier:
:+1:

imbriaco:
@lhaskins I agree. I'm doing the necessary work in `cogctl` and `cog-api-client` right now.

PR #495, opened 2016-04-04
Trigger pipelines in response to external events

### LINKED ISSUE #344: Consume and respond to external events
Think of kicking off a command pipeline in response to an external webhook (e.g., when a PR is created or merged in Github).

### LINKED ISSUE #385: Event Hook REST API & Data Model
- [ ] Define `event_hook` table schema
- [ ] Define Ecto model
- [ ] Create Ecto migration
- [ ] Create `EventHookController`

See #344 "Managing event hooks" for details.

### LINKED ISSUE #387: Event Hook Endpoint
### Summary

The event hook endpoint should be a separate Cowboy listener ([docs](http://ninenines.eu/docs/en/cowboy/HEAD/guide/introduction/)) which accepts requests for routes shaped like `/event_hooks/1581880e-c337-45cc-a1ec-7a09be2f6612` where the UUID is the unique id of a event hook definition. Listen address and port should be configured in Cog's config file. The config should be smart enough to use environment variables for these values if the variables exist.

Internally the endpoint should look like another chat adapter to allow for hook-specific output templates and easy integration with existing pipeline execution infra.

### External event handling step-by-step

1. Does the URL root begin with `event_hooks`?
  - No: Return 404.
  - Yes: Continue.
1. Is the id a valid UUID?
  - No: Return 400 and "Event hook ids must be valid UUIDs" message.
  - Yes: Continue.
1. Does a event hook exist with the given id?
  - No: Return 404.
  - Yes: Continue.
1. Does the user pointed to by `as_user` exist?
  - No: Return 422 and "The user associated with this event hook does not exist" message.
  - Yes: Continue.
1. Is the request's HTTP verb listed in the hooks `verbs` field?
  - No: Return 400 and "Method XXX not supported. Supported HTTP methods are X, Y, Z" message.
  - Yes: Continue.
1. Assign the request a unique request id.
  - `:erlang.monotonic_time/0` is sufficient for initial implementation
1. Create initial calling arg map.
  - Add request parameters, if any, to the map.
  - Gather any headers specified in the hook's `capture_headers` field into a map. Add this map to the args map with the name `headers`. If none match or the hook definition contains none use an empty map as a placeholder value.
  - If a request body exists, add it to the args map with the name `body`.
1. Create a new `Spanner.Command.Request` using the retrieved user as `requestor`, the map built in step 7 as `args`, and a `reply_to` containing a MQTT topic string like so: `event_hooks/<hook_id>/<request_id>`.
1. Send the constructed request to the executor
  - If the pipeline succeeds return a 200 and use any pipeline output returned as the response body.
  - If the pipeline crashes send a 500 and the error message returned by the executor.

### To Dos
- Define how to route executor messages sent via `reply_to` topic to the correct Erlang process.

Part of #344

christophermaier:
Now pipelines may be triggered by HTTP requests instead of just via chat
messages. This will allow Cog to, e.g., respond to webhooks.

Users may define "event hooks", which define a pipeline to run in
response to an HTTP POST request. The request itself acts as input to
the pipeline (a COG_ENV for the first command in a pipeline,
technically), allowing query parameters, request headers and JSON
request bodies to be accessible to the command. Pipelines may run as a
defined user, or may require token-based authentication for
execution. Additionally, pipelines may send complete output back to the
requestor via HTTP response, or direct output to the currently-enabled
Cog chat adapter, or both.

A "CRUD" REST API is provided for event hooks and is currently the sole
means of manipulating event hooks (i.e., no chat commands for this just
yet). Execution of hooks, while done via HTTP request, is handled by a
different Phoenix Endpoint than the REST API (defaulting to
port 4001, vs. the REST API's default port 4000).

To do this, we introduce a new "HTTP adapter", which is an
implementation of our current Adapter behaviour. This adapter is
different than previous ones in a few ways, though.

First, it is a synchronous adapter. All chat adapters thus far have been
asynchronous, in that responding to a Cog request does not need to be
coupled to the _receiving_ of that request. As we are now responding to
HTTP requests, this is no longer the case; we _must_ send a response
when we receive a request! To deal with this, an "AdapterBridge" is
introduced to track which pipeline executions correspond to which
currently-processing HTTP request. (If you squint, this "AdapterBridge"
performs the same function for the HTTP adapter as our Slack RTM/API and
HipChat API modules perform for those adapters.)

Secondly, as a synchronous adapter, the HTTP adapter is the first to
send output to mulitple places _via multiple adapters_. As mentioned, it
must _always_ send something back for the HTTP response. However, it
must _also_ be able to (optionally) send output to chat.

Third, the HTTP adapter is the first to seed a pipeline with data; a
`COG_ENV` is now presented to the pipeline from the beginning, rather
than being generated during execution. This allows commands access to
the data in the HTTP request. To facilitate correlation of HTTP request
processing and resultant pipeline execution, the HTTP adapter also
passes the unique ID of the HTTP request to the Executor as the
pipeline's unique ID. As a result, the Adapter behaviour contract was
modified to allow passing these two pieces of data in with a request.

To help manage these differences in a general way, adapters now know
whether they are a chat adapter or not, with the HTTP adapter being the
first non-chat adapter. Behavior in the Executor and associated code can
ask the adapter which it is in order to do the right thing.

Other changes of note:

* The `NOCHAT` behavior has been dropped. If you want to run without a
  chat adapter, choose the `null` adapter.
* A new embedded permission `operable:manage:hooks` has been added; all
  CRUD REST API interaction around event hooks requires this permission.
* The `TemplateCache` has been moved from the top-level supervisor down
  to the `CommandSup` supervisor.
* Pipeline destination processing grew more complicated, now that output
  can be "full" or "status-only" (the latter for HTTP request that don't
  redirect output to `here`, but still require something for the HTTP
  response), so a new `Cog.Command.Pipeline.Destination` module was
  created to manage destination processing. This removes a fair bit of
  code from the Executor, as well as an entire FSM state.
* Lots of tests
* A new testing utility `Cog.Snoop` is used to eavesdrop on MQTT message
  traffic. Here it is used to verify certain aspects of event hook
  processing, but may be useful in other tests,as well.

Closes #344
Closes #385
Closes #387


lib/cog.ex
```diff
@@ -69,27 +101,8 @@ defmodule Cog do
     end
   end
 
-  ########################################################################
-
-  defp build_children(:dev, nochat, _) when nochat != nil do
-    [worker(Cog.Repo, []),
-     worker(Cog.TokenReaper, []),
-     supervisor(Cog.Endpoint, [])]
-  end
```
vanstee: Yay! :bomb: NOCHAT

lib/cog/command/pipeline/destination.ex
```diff
@@ -0,0 +1,123 @@
...
+  # adapter. We do this because, e.g., HTTP requests need a response
+  # of some kind, and we want to indicate whether or not the pipeline
+  # succeeded.
+  #
+  # If the origin _is_ one of the destinations, we do nothing special;
+  # all explicitly listed destinations receive full output.
+  defp maybe_add_origin([], _origin_adapter),
+    do: [make_destination("here")]
+  defp maybe_add_origin(destinations, origin_adapter) do
+    if !originator_is_destination?(destinations) && !origin_adapter.chat_adapter? do
```
vanstee: Might be nice to invert the boolean expression here to `originator_is_destination?(destinations) || origin_adapter.chat_adapter?`

lib/cog/command/pipeline/destination.ex
```diff
@@ -0,0 +1,123 @@
...
+        {:ok, :ok} -> # Everything is good; collect
+          {:ok, [value|values]}
+        {:ok, :error} -> # We hit the first error; discard everything before
+          {:error, [value]}
+        {:error, :error} -> # Still hitting errors; collect
+          {:error, [value|values]}
+        {:error, :ok} -> # Ignore successful resolutions when we've encountered errors before
+          acc
+      end
+    end)
```
vanstee: I've run into the pattern a few times already. We might be able to wrap it up in a helper function somewhere.

lib/cog/event_hook_endpoint.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Cog.EventHookEndpoint do
+  use Phoenix.Endpoint, otp_app: :cog
+
+  plug Plug.RequestId
+  # plug Plug.Logger
```
vanstee: Is there any harm in leaving this in here in production?

lib/cog/repository/event_hooks.ex
```diff
@@ -0,0 +1,63 @@
...
+        nil ->
+          {:error, :not_found}
+      end
+    else
+      {:error, :bad_id}
+    end
+  end
+
+  def hook_definition!(id),
+    do: Repo.get!(EventHook, id)
```
vanstee: Is this used anywhere?

test/cog_test.exs
```diff
@@ -0,0 +1,22 @@
+defmodule CogTest do
+  use ExUnit.Case
```
vanstee: These tests can probably run async

web/controllers/v1/event_hook_controller.ex
```diff
@@ -0,0 +1,84 @@
...
+  plug Cog.Plug.Authorization, permission: "#{Cog.embedded_bundle}:manage_hooks"
+
+  alias Cog.Repository.EventHooks
+
+  def index(conn, _params),
+    do: render(conn, "index.json", event_hooks: EventHooks.all)
+
+  def show(conn, %{"id" => id}) do
+    # hook = EventHooks.hook_definition!(id)
+    # render(conn, "show.json", event_hook: hook)
```
vanstee: Was this meant to be left here?

web/controllers/v1/event_hook_execution_controller.ex
```diff
@@ -0,0 +1,116 @@
...
+      case Map.fetch(acc, k) do
+        :error ->
+          Map.put(acc, k, v)
+        multiple when is_list(multiple) ->
+          Map.put(acc, k, Enum.concat(multiple,[v]))
+        single ->
+          Map.put(acc, k, [single, v])
+      end
+    end)
+  end
```
vanstee: Just curious, what's the usecase for multiple header values? Couldn't think of one off the top of my head.

vanstee:
This was a lot to read though, but so far it looks great. I'll take a look again with a fresh brain later today.

lib/cog/command/pipeline/destination.ex
christophermaier: Good call; fixing

lib/cog/command/pipeline/destination.ex
christophermaier: Yeah, I've had the same thought; it feels like there's something we could extract. I took a stab at it a little while back, but I think it might need some tweaking of existing functions to fit some kind of consistent interface.

Might be worth revisiting now. Maybe a separate PR, though?

vanstee: Yeah sorry, not necessary for this PR.

lib/cog/event_hook_endpoint.ex
christophermaier: No, just an oversight :smile:

lib/cog/repository/event_hooks.ex
christophermaier: No, it's not... looks like that was something I tried but ended up backing out of. I'll remove it.

web/controllers/v1/event_hook_controller.ex
christophermaier: Nope... removing

test/cog_test.exs
christophermaier: Indeed, thanks!

web/controllers/v1/event_hook_execution_controller.ex
christophermaier: Multiple headers are allowed. It's a way you can specify a list of values.

https://tools.ietf.org/html/rfc7230#section-3.2

vanstee: Ah TIL.

_shakes fist at `Set-Cookie`_

lib/cog.ex
```diff
@@ -41,55 +44,65 @@ defmodule Cog do
...
+
+  # TODO: the "null" and "test" adapters need to be brought in only in
+  # non-prod environments
+  def chat_adapters do
+    %{"slack"   => Cog.Adapters.Slack,
+      "hipchat" => Cog.Adapters.HipChat,
+      "irc"     => Cog.Adapters.IRC,
+      "null"    => Cog.Adapters.Null,
+      "test"    => Cog.Adapters.Test}
+  end
```
mpeck: I wish this was more data driven. Not suggesting you change it now btw. It just bothers me for some reason to have the adapters hard coded.

lib/cog/adapters/null.ex
```diff
@@ -22,7 +19,7 @@ defmodule Cog.Adapters.Null do
     true
   end
 
-  def lookup_user(handle: "admininator") do
+  def lookup_user("admininator") do
```
mpeck: What is an "admininator"? Should that be "administrator"?

lib/cog/command/pipeline/destination.ex
```diff
@@ -0,0 +1,123 @@
...
+  end
+
+  # Redirect destinations may be targeted to an adapter different from
+  # where they originated from.
+  #
+  # Destinations prefixed with "chat://" will be routed through the
+  # active chat adapter module. Anything else will be routed through
+  # the adapter that initially serviced the request.
+  @spec adapter_destination(String.t, atom) :: {atom, String.t}
+  defp adapter_destination("chat://" <> destination, _origin_adapter) do
```
mpeck: Is there a potential to have multiple active chat adapters? If so, would we want to route based on something like `slack://` or `hipchat://`?

lib/cog/command/pipeline/executor.ex
```diff
@@ -286,47 +257,6 @@ defmodule Cog.Command.Pipeline.Executor do
   ########################################################################
   # Redirection Resolution Functions
 
-  # Returns {:ok, {adapter, room}} or {:error, invalid_redirect}
-  defp lookup_room("me", state) do
```
mpeck: I'm quite happy to see this being removed from the executor.

web/controllers/v1/trigger_execution_controller.ex
```diff
@@ -0,0 +1,166 @@
...
+  alias Cog.Plug.Authentication
+
+  alias Cog.Models.User
+  alias Cog.Models.Trigger
+  alias Cog.Repository.Triggers
+  alias Cog.Adapters.Http.AdapterBridge
+
+  plug :parse
+
+  def execute_trigger(conn, %{"id" => trigger_id}) do
```
mpeck: This function is kinda hard for me to follow. Would it be clearer if it was broken up a bit? Maybe using `with` and some more intermediary functions?

lib/cog/command/pipeline/destination.ex
kevsmith: Maybe in the future? 

I can't recall the concrete issues but I recall discussing the problems encountered by running multiple adapters simultaneously with @imbriaco and others a while back.

config/config.exs
```diff
@@ -118,6 +118,18 @@ config :cog, Cog.Endpoint,
...
+config :cog, Cog.TriggerEndpoint,
+  http: [port: System.get_env("COG_EVENT_HOOK_PORT") || 4001],
+  root: Path.dirname(__DIR__),
+  debug_errors: false,
+  cache_static_lookup: false,
+  check_origin: true,
+  render_errors: [accepts: ~w(json)],
+  # pubsub: [name: Carrier.Messaging.Connection,
+  #          adapter: Phoenix.PubSub.PG2],
+  secret_key_base: System.get_env("COG_COOKIE_SECRET")
```
kevsmith: Are there any issues created by sharing the cookie secret with the REST API?

lib/cog/adapters/http.ex
```diff
@@ -0,0 +1,33 @@
...
+  def name,
+    do: "http"
+
+  def display_name,
+    do: "HTTP"
+
+  def chat_adapter?,
+    do: false
+
+end
```
kevsmith: So cool :)

lib/cog/adapters/http/adapter_bridge.ex
```diff
@@ -0,0 +1,49 @@
...
+                     timeout)
+    catch
+      :exit, {:timeout,_} ->
+        {:error, :timeout}
+    end
+  end
+
+  def finish_request(room, response),
+    do: GenServer.call(__MODULE__, {:finish_request, room, response})
+
```
kevsmith: I think a timeout of `:infinity` might be better on both `GenServer.call`s. Since there's only one of these bridge processes, and it's one of the few Cog components exposed to potential high load (assuming most triggers are activated via m2m traffic), there's the possibility these calls could timeout which could lead to some interesting failure scenarios.

lib/cog/adapters/http/adapter_bridge.ex
```diff
@@ -0,0 +1,49 @@
...
+  ########################################################################
+
+  def init([]),
+    do: {:ok, %{}}
+
+  # TODO: pass timeout in order to queue up clearing the data from the map?
+  def handle_call({:submit_request, requestor, id, initial_context, pipeline}, from, state) do
```
kevsmith: I agree with this TODO. A good candidate for a follow-on PR, imo.

lib/cog/command/pipeline/executor.ex
```diff
@@ -387,26 +317,39 @@ defmodule Cog.Command.Pipeline.Executor do
...
+    |> Map.get(:status_only, [])
+    |> Enum.each(fn(dest) ->
+      publish_response(%{status: "ok"}, dest.room, dest.adapter, state)
+    end)
+
   end
 
   defp succeed_with_response(state) do
     respond(state)
+    # TODO: what happens if we fail due to a template error?
```
kevsmith: This is a good question to answer. I'd imagine there will be more than a few first time devs stubbing their toes on template typos.

kevsmith:
:+1:

imbriaco:
:zap:

PR #498, opened 2016-04-04
fetch_user_permissions honors group-granted roles

### LINKED ISSUE #496: Ensure that executor permission checks can see all permissions
There is a bug where users that have a permission because they're in a group that has a role with that permission aren't being recognized as actually having that permission.

christophermaier:
Previously, the `fetch_user_permissions` stored procedure returned
inconsistent results because it did not retrieve permissions a user had
by being in a group that had a role that had the permission.

Fixes #496

### DIFF
--- a/priv/repo/migrations/20160404135526_fetch_user_permissions_honors_roles.exs
+++ b/priv/repo/migrations/20160404135526_fetch_user_permissions_honors_roles.exs
@@ -1,0 +1,73 @@
+defmodule Cog.Repo.Migrations.FetchUserPermissionsHonorsRoles do
+  use Ecto.Migration
+
+  def up do
+    execute """
+CREATE OR REPLACE FUNCTION fetch_user_permissions (
+  p_user users.id%TYPE)
+RETURNS TABLE(id uuid, name text)
+LANGUAGE plpgsql STABLE STRICT
+AS $$
+BEGIN
+
+RETURN QUERY WITH
+
+  -- Walk the tree of group memberships and find
+  -- all the groups the user is a direct and
+  -- indirect member of.
+  all_groups as (
+  SELECT group_id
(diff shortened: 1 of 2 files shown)


priv/repo/migrations/20160404135526_fetch_user_permissions_honors_roles.exs
```diff
@@ -0,0 +1,73 @@
...
+SELECT p.id, ns.name||':'||p.name as name
+  FROM permissions as p, namespaces as ns, all_permissions as ap
+ WHERE ap.permission_id = p.id and p.namespace_id = ns.id;
+END;
+$$;
+   """
+  end
+
+  def down do
+    Code.eval_file("20151118002150_fetch_user_permissions.exs", __DIR__)
```
vanstee: This is pretty clever. Does the right migration version get set in the schema?

I'd be happy with just raising "Unreversable migration" here as well.

vanstee:
Looks good. Wish I could see a diff of these functions somehow, but I don't even think storing a schema file in the repo would fix that.

priv/repo/migrations/20160404135526_fetch_user_permissions_honors_roles.exs
christophermaier: Hmm... that's a good question re: migration version. I'll have to check.

Worst case, I'd just paste in the previous code into this `down` function, since it's reversible (just ugly to do so).

Also, this is why I :heart: http://sqitch.org

christophermaier: Yeah, it looks like the migration version is fine... we're still calling `down` from _this_ migration, so that's what Ecto knows about... the rest is just clever hackery to stop us from having to type more :)

Issue #469, opened 2016-03-26
Rewrite the TemplateCache to expose "core" templates


christophermaier:
The current `TemplateCache` is tightly coupled to our adapters (specifically the Slack and HipChat adapters), as well as to our "embedded" templates (`raw`, `text`, and `json`).

To remedy this, we should create the notion of a "core", or bundle-independent, template. Current real templates are tied to bundles, but things like `raw`, `text`, and `json` are general-purpose templates, dependent only on the adapter being used. We'll need to devise a way to store these in the database. Since these are effectively embedded templates, we don't need to worry about providing a way for users to manipulate the templates (e.g. install their own).

Additionally, logic for template "fallback" is currently embedded in the executor. For instance, if a command has declared that its output should be formatted with the "foo" template, but no "foo" template exists for the current chat adapter, we can fall back to one of the embedded templates. This should be incorporated into the main template lookup logic. This also implies that in addition to supplying "core" templates for each of our adapters, we should also provide a bare-bones fallback template for "any" adapter. This ensures that any adapter will be able to generate output, whether or not it has a template that's been tailored to it.

Since the current system is built around anonymous functions that perform the actual template rendering, we can also simplify the executor a bit by baking error-recovery into these template functions. Eventually, we'll want to incorporate our error messaging into templates, and this error-recovery probably helps get around the interesting situation of how you display an error rendering an error template. (It may also be necessary to do something like implementing core template fallbacks as code to provide some stable "base case" for situations like this.)

Finally, the interface that the executor retrieves templates from should not indicate anything about it being implemented in terms of a cache. That is an implementation detail, and the executor shouldn't need to change if we remove the cache, or dramatically alter its implementation.

The cache implementation itself could be simplified, as well; take a look at https://github.com/operable/cog/blob/2ac109995e0a996783ceacff5a4e81af230b3d8b/lib/cog/command/bundle_cache.ex for suggestions / patterns.

vanstee:
Plan of action:

- [ ] Move template code out of executor into Template module with cache related API stuff hidden away
- [ ] Use a function partial (new feature for fumanchu) to render json from a mustache template
- [ ] Store templates in the database and look them up on demand
- [ ] Add "any" fallback templates

PR #499, opened 2016-04-04
Fallback templates

vanstee:
This PR does a few things with the goal of implementing better fallback templates:

* Simplify template interface for executor
* Allow mustache to render function partials
* Move all templates (even raw, json, and text) into the database
* Fallback when looking for templates (exact -> fallback for adapter -> generic fallback)
* Split up template caching logic
* Remove test templates and rework tests to match json results

Closes https://github.com/operable/cog/issues/469


lib/cog/template_cache.ex
```diff
@@ -1,7 +1,6 @@
 defmodule Cog.TemplateCache do
```
vanstee: This `TemplateCache` is really just a GenServer wrapper around an ets table. I wasn't happy with tangling this all up with the `Template` module (which isn't even it's own process) so it seemed reasonable to just throw the cache in its own process. The only downside is that it takes 2 message sends to lookup and then insert, where before we were able to do that in one call. Feedback on this decision would be great though.

lib/cog/template.ex
```diff
@@ -0,0 +1,91 @@
...
+  # Used in integration tests.
+  def fetch_source("test", _bundle_id, _template, context) do
+    fetch_source(@fallback_adapter, nil, "raw", context)
+  end
+
+  def fetch_source(adapter, bundle_id, template, _context) do
+    with {:error, :template_not_found} <- fetch(adapter, bundle_id, template),
+         {:error, :template_not_found} <- fetch(adapter, nil, template),
+         {:error, :template_not_found} <- fetch(@fallback_adapter, nil, template),
+         do: {:error, :template_not_found}
```
vanstee: After trying a few flows for this, this still seemed like the best way to implement it. Falling through with a `cond` with `match?` clauses doesn't give you access to the result of the function you just called, otherwise that would work.

lib/cog/template.ex
```diff
@@ -0,0 +1,91 @@
...
+      template_fun.(context)
+    end
+  end
+
+  def fetch_compiled_fun(adapter, bundle_id, template, context) do
+    with :error              <- TemplateCache.lookup(adapter, bundle_id, template),
+         {:ok, source}       <- fetch_source(adapter, bundle_id, template, context),
+         {:ok, template_fun} <- compile(source),
+         :ok                 <- TemplateCache.insert(adapter, bundle_id, template, template_fun),
+         do: {:ok, template_fun}
```
vanstee: Would also like feedback here, since this is a little weird with the first match being `:error`.

lib/cog/command/pipeline/executor.ex
```diff
@@ -428,50 +428,38 @@ defmodule Cog.Command.Pipeline.Executor do
...
-      fun when is_function(fun) ->
-        fun.(context)
-      nil ->
-        Logger.warn("The template `#{template}` was not found for adapter `#{adapter}` in bundle `#{command.bundle.name}`; falling back to the default")
-        render_template(command, adapter, nil, context)
+  defp render_template(adapter, bundle, template, context) do
+    case Template.render(adapter, bundle.id, template, context) do
+      {:ok, output} ->
+        {:ok, output}
+      {:error, :template_not_found} ->
```
christophermaier: Still reading through the PR... would it make sense to move this fallback logic into `Template`? That way the logic is all consolidated.

lib/cog/queries/template.ex
```diff
@@ -2,7 +2,15 @@ defmodule Cog.Queries.Template do
   use Cog.Queries
   alias Cog.Models.Template
 
-  def template_source(bundle_id, adapter, name) do
+  def template_source(adapter, nil, name) do
+    from t in Template,
+    where: is_nil(t.bundle_id) and
```
christophermaier: Does `where: t.bundle_id == nil` not work?

lib/cog/command/pipeline/executor.ex
vanstee: I don't really consider this as part of the fallback logic. This is a real error case where the template specified could not be found, which I think the executor should be able to react to. In the future I expect this clause to eventually render a "not_found" template where we print an error and return the result in json as we currently do.

Also as a minor aside, this prints the bundle name which isn't something the `Template` module knows about. Woudln't be hard to change but I do like the API surface we currently have.

lib/cog/template.ex
```diff
@@ -0,0 +1,91 @@
...
+
+  @fallback_adapter "any"
+
+  def render(adapter, bundle_id, template, context) do
+    with {:ok, template_fun} <- fetch_compiled_fun(adapter, bundle_id, template, context) do
+      template_fun.(context)
+    end
+  end
+
+  def fetch_compiled_fun(adapter, bundle_id, template, context) do
```
christophermaier: I think this can be private.

lib/cog/template.ex
```diff
@@ -0,0 +1,91 @@
...
+    with :error              <- TemplateCache.lookup(adapter, bundle_id, template),
+         {:ok, source}       <- fetch_source(adapter, bundle_id, template, context),
+         {:ok, template_fun} <- compile(source),
+         :ok                 <- TemplateCache.insert(adapter, bundle_id, template, template_fun),
+         do: {:ok, template_fun}
+  end
+
+  def fetch_source(adapter, bundle_id, nil, context) do
```
christophermaier: I think this can be private, too.

lib/cog/queries/template.ex
vanstee: Yeah, kinda sucks:

```
** (Ecto.Query.CompileError) comparison with nil is forbidden as it always evaluates to false. If you want to check if a value is (not) nil, use is_nil/1 instead
```

christophermaier: Was afraid of that :frowning: Yay, three-valued logic.

lib/cog/template.ex
```diff
@@ -0,0 +1,91 @@
...
+    fetch_source(adapter, bundle_id, default_template(context), context)
+  end
+
+  # Always use the raw template when responding to the test adapter.
+  # Used in integration tests.
+  def fetch_source("test", _bundle_id, _template, context) do
```
christophermaier: Should this be the first function head, then? Also, does the fallback logic in the executor not account for this?

lib/cog/template.ex
christophermaier: I think the code is good, but some comments giving a little explanation would help out a lot.

lib/cog/template.ex
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def fetch_source(adapter, bundle_id, template, _context) do
+    with {:error, :template_not_found} <- fetch(adapter, bundle_id, template),
+         {:error, :template_not_found} <- fetch(adapter, nil, template),
+         {:error, :template_not_found} <- fetch(@fallback_adapter, nil, template),
+         do: {:error, :template_not_found}
+  end
+
+  def compile(source) do
```
christophermaier: Private?

lib/cog/command/pipeline/executor.ex
christophermaier: Fair point; :+1:

lib/cog/template.ex
christophermaier: Just some documentation explaining the rationale.

I guess the only thing that I'm not super jazzed about with these "negative `with`" calls is that, absent some specs or docs, it's a little hard to figure out what all you can expect to return (though, to be fair, I think that applies equally to "positive `with`" calls, as well).

lib/cog/template_cache.ex
christophermaier: I think that's fine; after all, the idea is that with a cache, we'll be saving a lot more calls in the long run.

Having it be its own module is a nice separation of concerns, and makes both the caching logic and the "normal" logic easier to understand.

test/cog/template_test.exs
```diff
@@ -0,0 +1,30 @@
+defmodule Cog.TemplateTest do
+  use ExUnit.Case
+  alias Cog.Template
+
+  test "rendering json for the slack adapter" do
```
christophermaier: Mwahaha... testing the templates independent of any command execution is :+1: :+1: :100: :confetti_ball:

test/support/adapter_assertions.ex
```diff
@@ -5,6 +5,24 @@ defmodule Cog.AdapterAssertions do
...
 
+  # Because the raw template returns encoded json joined by newlines we have to
+  # add in newlines and then pull out the single result or return a list of
+  # result for those cases.
+  def assert_payload(%{"response" => response}, expected_payload) do
+    response = response
+    |> String.replace(~r/^}/m, "},")
+    |> String.rstrip(?,)
+
+    payload = case Poison.decode!("[#{response}]", keys: :atoms) do
```
christophermaier: Why the wrapping with a list here?

vanstee: This is pretty gnarly. The comment above the function helps a little bit.

We don't know how many json results are printed (separated by newlines) so to make sure we can always parse it, I need to add commas and wrap it in an array before parsing. Eventually I hope these tests just intercept things on the message queue, but for now to get these tests passing, this was the best solution.

test/support/adapter_assertions.ex
```diff
@@ -5,6 +5,24 @@ defmodule Cog.AdapterAssertions do
     assert response == expected_message
   end
 
+  # Because the raw template returns encoded json joined by newlines we have to
+  # add in newlines and then pull out the single result or return a list of
+  # result for those cases.
+  def assert_payload(%{"response" => response}, expected_payload) do
+    response = response
+    |> String.replace(~r/^}/m, "},")
+    |> String.rstrip(?,)
```
christophermaier: Earlier, I wrote some code similar to this in the process of doing work that for a number of reasons ended up not being merged. I'd shared it with some other folks and thought it had been incorporated that way, but I can't seem to find it in the codebase.

```elixir
  #
  # Example:
  #
  # lookbehind.
  #
  # Now you have `n` problems.
```
I wonder if being stricter in the regex would be useful here.

christophermaier:
Looks good! :+1: Excited to see this :ship:

lib/cog/template.ex
vanstee: Oh, yep. The only difference is that we're using `raw` here instead of `json`, which is easier to parse. It's not perfect, but I'm ok merging this and removing it once we have a better way to test commands in isolation.

test/support/adapter_assertions.ex
vanstee: I'd rather avoid using a positive lookahead, since I can't easily parse it at first glance. Although, the regex I wrote isn't much easier to understand. Maybe some more in-depth comments would be enough.

PR #500, opened 2016-04-04
Clean up the sanitization steps and comments

vanstee:


kevsmith:
I'm :+1: if we add back the comments. I think the narrative is useful for understanding why each regex was chosen.

christophermaier:
:+1:

Issue #489, opened 2016-04-01
Allow a user to update their own user information


lhaskins:
A user should be able to update their own information without having the manage_users permission.

lhaskins:
### Motivation/Objective
A user should be able to update their own user information regardless of permissions.

### Research
Instead of updating in the authorization plug to check for the method and the data, have an intermediate plug that will determine if authorization is needed in the first place. 

This will allow for an isolation of this change such that it stays in the User controller and the authorization isn't called if it isn't necessary.

I'm not sure if I am able to use certain functions in a guard such that I can access the connection that is used in the plug. 

The plug updates the connection data.

### Plan
Since what I can do in a guard function is limited in accessing the connection map, I need to perform much of the logic in a separate plug and then update the plug only if it is the same user updating their own information. 

### Tasks
- [ ] Write new plug method.
- [ ] Write new guard macro for use when determining whether to authenticate the execution of the command.
- [ ] Update/add tests.
- [ ] Update/add documentation.

PR #501, opened 2016-04-04
Allow a user to update their own user data with or without the manage_users permission

lhaskins:
Add a new plug in the UserController that determines if this is a self updating task or if the manage_user permission is needed. Perform authorization only is not self_updating.

Fixes https://github.com/operable/cog/issues/489

### DIFF
--- a/web/controllers/v1/user_controller.ex
+++ b/web/controllers/v1/user_controller.ex
@@ -1,6 +1,4 @@
 defmodule Cog.V1.UserController do
-  import Plug.Conn
-
   use Cog.Web, :controller
 
   alias Cog.Models.User
@@ -61,10 +59,14 @@ defmodule Cog.V1.UserController do
 
   ###############################################
   # Plug function - local only to user controller
+  #
+  # This local plug allows a user to update and view
+  # thieir own user information without needing
+  # the "manage_users" permission.
   ###############################################
-  @spec check_self_updating(Plug.Conn.t, []) :: Plug.Conn.t
-  def check_self_updating(conn, opts \\ []) do
-    if determine_self_updating(conn) do
+  @spec check_self_updating(Plug.Conn.t, [Keyword.t]) :: Plug.Conn.t
+  defp check_self_updating(conn, opts) do
+    if self_updating?(conn) do
       conn
     else
       plug_opts = Cog.Plug.Authorization.init(opts)
@@ -72,13 +74,9 @@ defmodule Cog.V1.UserController do
     end
   end
 
-    @doc """
-  Store the self updating user flag in the `conn`.
-  """
-  @spec determine_self_updating(%Plug.Conn{}) :: true | false
-  def determine_self_updating(conn) do
-    Map.has_key?(conn.private, :phoenix_action) and
-        conn.private.phoenix_action == :update and
+  @spec self_updating?(%Plug.Conn{}) :: true | false
+  defp self_updating?(conn) do
+    conn.private.phoenix_action in [:update, :show] and
         conn.assigns.user.id == conn.params["id"]
   end
 
--- a/web/controllers/v1/user_controller.ex
+++ b/web/controllers/v1/user_controller.ex
@@ -62,18 +62,13 @@ defmodule Cog.V1.UserController do
   ###############################################
   # Plug function - local only to user controller
   ###############################################
-  @spec check_self_updating(Plug.Conn.t, Plug.Conn.t) :: Plug.Conn.t
+  @spec check_self_updating(Plug.Conn.t, []) :: Plug.Conn.t
   def check_self_updating(conn, opts \\ []) do
     if determine_self_updating(conn) do
       conn
     else
       plug_opts = Cog.Plug.Authorization.init(opts)
-      case Cog.Plug.Authorization.call(conn, plug_opts) do
-        {:error, error} ->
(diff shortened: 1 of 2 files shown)


web/controllers/v1/user_controller.ex
```diff
@@ -56,4 +58,34 @@ defmodule Cog.V1.UserController do
     Repo.delete!(user)
     send_resp(conn, :no_content, "")
   end
+
+  ###############################################
+  # Plug function - local only to user controller
+  ###############################################
+  @spec check_self_updating(Plug.Conn.t, Plug.Conn.t) :: Plug.Conn.t
```
vanstee: Is the second arg a keyword list?

web/controllers/v1/user_controller.ex
```diff
@@ -56,4 +58,34 @@ defmodule Cog.V1.UserController do
...
+  def check_self_updating(conn, opts \\ []) do
+    if determine_self_updating(conn) do
+      conn
+    else
+      plug_opts = Cog.Plug.Authorization.init(opts)
+      case Cog.Plug.Authorization.call(conn, plug_opts) do
+        {:error, error} ->
+          error
```
vanstee: Is this matching a specific error? It looks like in the authorization plug we always return a conn, which has been halted if the user is not authorized.

vanstee:
Looks good. If we need other custom authz  checks in the future it might be cool to support passing in a function that defaults to `User.has_permission?`

lhaskins:
True, I was trying to keep the change that affects the UserController module in the UserController module and not in the Authorization plug. That would be good to keep in mind if we need to customize the authorization process further.

web/controllers/v1/user_controller.ex
```diff
@@ -56,4 +58,29 @@ defmodule Cog.V1.UserController do
     Repo.delete!(user)
     send_resp(conn, :no_content, "")
   end
+
+  ###############################################
+  # Plug function - local only to user controller
+  ###############################################
+  @spec check_self_updating(Plug.Conn.t, []) :: Plug.Conn.t
```
christophermaier: This spec is saying that the second argument is an empty list. You'd want `List.t` at least, but in this case, I think it's really `Keyword.t`

web/controllers/v1/user_controller.ex
```diff
@@ -56,4 +58,29 @@ defmodule Cog.V1.UserController do
...
+  def check_self_updating(conn, opts \\ []) do
+    if determine_self_updating(conn) do
+      conn
+    else
+      plug_opts = Cog.Plug.Authorization.init(opts)
+      Cog.Plug.Authorization.call(conn, plug_opts)
+    end
+  end
+
+    @doc """
```
christophermaier: Indentation

web/controllers/v1/user_controller.ex
```diff
@@ -56,4 +58,29 @@ defmodule Cog.V1.UserController do
...
+  Store the self updating user flag in the `conn`.
+  """
+  @spec determine_self_updating(%Plug.Conn{}) :: true | false
+  def determine_self_updating(conn) do
+    Map.has_key?(conn.private, :phoenix_action) and
```
christophermaier: Won't it always have this key?

web/controllers/v1/user_controller.ex
```diff
@@ -56,4 +58,29 @@ defmodule Cog.V1.UserController do
...
+      plug_opts = Cog.Plug.Authorization.init(opts)
+      Cog.Plug.Authorization.call(conn, plug_opts)
+    end
+  end
+
+    @doc """
+  Store the self updating user flag in the `conn`.
+  """
+  @spec determine_self_updating(%Plug.Conn{}) :: true | false
+  def determine_self_updating(conn) do
```
christophermaier: Style-wise a name like `self_updating?` would be more idiomatic.

web/controllers/v1/user_controller.ex
```diff
@@ -56,4 +58,29 @@ defmodule Cog.V1.UserController do
     Repo.delete!(user)
     send_resp(conn, :no_content, "")
   end
+
+  ###############################################
+  # Plug function - local only to user controller
+  ###############################################
+  @spec check_self_updating(Plug.Conn.t, []) :: Plug.Conn.t
+  def check_self_updating(conn, opts \\ []) do
```
christophermaier: The default isn't needed, as the authorization plug will raise a `KeyError` if there's no `:permission` key.

web/controllers/v1/user_controller.ex
lhaskins: I thought so too, but during the research phase, I found there were times when it was not present. It may be always at this stage. I'll verify.

web/controllers/v1/user_controller.ex
```diff
@@ -1,10 +1,12 @@
 defmodule Cog.V1.UserController do
+  import Plug.Conn
```
christophermaier: Why is this imported?

web/controllers/v1/user_controller.ex
christophermaier: Some documentation explaining the rationale for this would be useful.

Also, can this be private?

christophermaier:
Minor comments, but overall looks nice :+1:

PR #502, opened 2016-04-06
Error templates

vanstee:
This provides only 2 basic error templates to start; an unregistered user template and a generic error template used from the `user_error` function. Also all custom messages for the generic errors and audit messages were pulled out into their own modules.

Closes https://github.com/operable/cog/issues/470

### DIFF
--- a/lib/cog/error_response.ex
+++ b/lib/cog/error_response.ex
@@ -1,0 +1,78 @@
+defmodule Cog.ErrorResponse do
+  alias Piper.Permissions.Ast
+  alias Cog.Models.Bundle
+  alias Cog.Models.Command
+
+  def render({:parse_error, msg}) when is_binary(msg),
+    do: "Whoops! An error occurred. " <> msg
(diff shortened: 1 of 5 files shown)


lib/cog/error_response.ex
```diff
@@ -0,0 +1,78 @@
...
+    do: "Whoops! An error occurred. There was an error rendering the template '#{template}' for the adapter '#{adapter}': #{inspect error}"
+  def render({:command_error, response}) do
+    error = response.status_message || response.body["message"]
+    "Whoops! An error occurred. " <> error
+  end
+
+  # `errors` is a keyword list of [reason: name] for all bad redirect
+  # destinations that were found. `name` is the value as originally
+  # typed by the user.
+  defp redirection_error_message(errors) do
```
vanstee: @christophermaier Putting this in a template felt weird because it's really only part of the final template rendered. I opted to leave it here for now.

kevsmith:
:+1:

Issue #477, opened 2016-03-29
Relay & Command Bundles 2.0


kevsmith:
## Motivation
Cog's current design wrt Relay and command bundles was good enough for our initial uses. As Cog has gotten wider use it's becoming clear we need to re-think both Relay and command bundle structure for at least the following reasons:

1. Defining and maintaining a bundle file format has turned into a distraction from the core problems we're trying to solve w/Cog. Other projects, like Docker, have done a much better job of defining a packaging format and have significantly greater mindshare.
1. Our current execution model provides some level of isolation between command invocations. Commands cannot see each other's execution environments but that's where our notion of isolation ends. Adopting current state of the art technology, such as Linux containers or BSD jails, would significantly increase isolation between command invocations and strengthen the security guarantees Cog can offer.
1. There's been a lot of interest in using Relays as executors pervasively deployed throughout an environment. This is significantly different from the uses cases informing Relay's original design. Using Relays in this way makes a lot of sense from both a technical and operational point of view. Relay's current design and the way it interacts w/Cog makes supporting this type of use awkward.

## Plan
Addressing the above points by iteratively improving Relay is possible but would take several man-weeks to execute. Even then we'd still have concerns with Relay's deployment footprint and setup experience since it would still be written in Elixir and thus bring non-trivial runtime requirements.

The best way forward is to aggressively reimplement Relay now while we still have the flexibility to make significant non-backwards compatible changes. At a high level the plan is as follows:

* Reimplement Relay in Go (single executable + config file deployment)
* Replace standard (zip file) bundles with Docker container images
* Continue to support simple bundle execution as currently provided
* Replace standard bundle command execution with Docker containers
* Place Cog in the driver's seat wrt authorizing and enabling new Relay hosts
* Allow users to group Relays into named groups
* Deploy bundles onto named Relay groups

## Tasks
- [x] Research Go MQTT and Docker client libs
- [x] Figure out how to invoke a Docker container
  - [x] Intercept stdout, stderr
  - [x] Send input over stdin
- [x] Implement Relay API
  - [x] Relay data model & migrations
  - [x] Relay CRUD REST API
  - [x] Relay group data model & migrations
  - [x] Relay group CRUD REST API
  - [x] cogctl support for above APIs
- [x] Rework Relay registration process
  - [x] Shared secret authentication
  - [x] Explicit enabled flag per-Relay
  - [x] Integrate w/emqttd's authentication and authorization subsystems
- [x] Rework bundle deployment process
[...]

PR #503, opened 2016-04-06
Go relay support

mpeck:
This is the work done to support the new relay. There are still some things to be done, but we thought it would be a good idea to get some feedback now rather than later.

Note: The relay test is currently disabled. It will need to be rewritten after the go-relay lands.

reference #477


web/controllers/v1/bundles_controller.ex
```diff
@@ -37,4 +39,114 @@ defmodule Cog.V1.BundlesController do
...
+      {:error, err} ->
+        send_failure(conn, err)
+    end
+  end
+
+  #### BUNDLE CREATION HELPERS ####
+
+  # First let's grab the config
+  defp get_config(conn) do
+    case conn.params do
```
vanstee: Is `conn.params` different from the `params` (second arg) passed into the controller action? If not it might be nice to do this bit in the action function head and then you only have to send `params` to this function instead of the entire `conn`.

test/cog/queries/command_test.exs
```diff
@@ -7,7 +7,7 @@ defmodule Cog.Queries.Command.Test do
   alias Cog.Repo
 
   setup do
-    {:ok, bundle} = Repo.insert(%Bundle{name: "test_bundle", config_file: %{}, manifest_file: %{}})
+    {:ok, bundle} = Repo.insert(%Bundle{name: "test_bundle", config_file: %{}})
```
vanstee: I think `Repo.insert!` might give us a better error message here instead of a bad match against `{:ok, bundle}`.

Issue #487, opened 2016-03-31
Manage relays and relay groups in cogctl


imbriaco:
Add support for the new Relay and RelayGroup endpoints to `cogctl`.

lhaskins:
### Motivation/Objective
There are no commands in cogctl that allow for the viewing or manipulation of relays for Cog. The ability to perform the following must be added:

#### Relays
* relay#index (a list of all relay names, ids, and creation times of relays for the Cog instance)
* relay#create (allows for the creation of a relay with the specified name  and token - Choose wisely. The relay name is currently immutable.)
* relay#show (show the details of a specific relay to include: name, id, creation time, status, description, and group association)
* relay#delete (deletes a particular relay and if it is the last relay in a group, deletes the corresponding group)
* relay#enable (enables a relay)
* relay#disable (disables a relay)
* relay#update (allows for the updating of the name and description of a relay)

#### Relay Groups
* relay_groups#index (a list of all relay groups containing: group names, ids, and creation times for the Cog instance)
* relay_groups#create (all for the creation of a relay group with a specified name)
* relay_groups#show (shows the details of a relay group to include: group name, id, creation time, associated bundles and associated relays. The bundles and relays should list names and IDs.)
* relay_groups#update (updates the group that the relay is associated with)
* relay_groups#delete (deletes a relay group. This should verify that there are no associated relays in the group before the group is deleted.)
* relay_groups#add (Adds a relay to a relay group)
* relay_groups#remove (Removes a relay from a relay group (If it is the last relay in the group, the relay group should be deleted.) 
* relay_groups#assign (Assigns a bundle to a relay group)
* relay_groups#retract (Retracts a bundle from a relay group)

### Research/Questions
Should the relay group be deleted when the last relay is deleted? Should we display a message that states that there are no more relays in the group upon the deletion of the last relay from that group?

### Plan
The results from the command `cogctl relays` should display in the following format:
```
NAME          STATUS       CREATED                   ID
test-relay    enabled      2016-04-05T16:00:00Z      123456
my-relay      enabled      2016-04-01T08:02:18Z      987654
```

The results from the command `cogctl relays create --name=my-relay --token=special_token` should display the following:
```
Create relay named my-relay.
```

The results from the command `cogctl relays update --description="This is a description of my relay" my-relay` should display the following:
```
Updated relay named my-relay.

ID              24680
Name            my-relay
Creation Time   2016-04-01T08:02:18Z
Status          enabled
Description     This is a description of my relay

[...]

PR #504, opened 2016-04-07
Rename Relay desc to description

lhaskins:
This is a quick PR to change the desc field to description, so that it is consistent throughout.

Fixes https://github.com/operable/cog/issues/487

### DIFF
--- a/priv/repo/migrations/20160328115654_create_relays.exs
+++ b/priv/repo/migrations/20160328115654_create_relays.exs
@@ -7,7 +7,7 @@ defmodule Cog.Repo.Migrations.AddRelays do
       add :name, :text, null: false
       add :token_digest, :text, null: false
       add :enabled, :boolean, null: false
-      add :desc, :text, null: true
+      add :description, :text, null: true
 
       timestamps
     end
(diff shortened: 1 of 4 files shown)


priv/repo/migrations/20160328115654_create_relays.exs
```diff
@@ -7,7 +7,7 @@ defmodule Cog.Repo.Migrations.AddRelays do
       add :name, :text, null: false
       add :token_digest, :text, null: false
       add :enabled, :boolean, null: false
-      add :desc, :text, null: true
```
vanstee: Should we create a new migration for this?

lhaskins: Yeah, I wondered the same thing, but from looking at our discussions in the past, it seemed since we hadn't released this yet, that this would be OK. 

I'm cool with just adding a new migration, if that's preferred at this point.

vanstee: Oh yeah good point about the release. I'd probably bias towards a new migration but this is probably fine since no one has run it yet. Whichever is good.

vanstee:
LGTM

priv/repo/migrations/20160328115654_create_relays.exs
lhaskins: yeah, last time I wrote a separate migration, since that was better to me, but was asked to just modify the existing one since the code hadn't been released, so I'm trying to be consistent... :smile:

PR #508, opened 2016-04-07
Fix granting a group a role

lhaskins:
This allows a group to be assigned a role. The namespace information in a permission was not being preloaded.


kevsmith:
:+1: to merge

PR #509, opened 2016-04-07
Include rules in bundles#show and bundles#index

imbriaco:
* Updated `GET /v1/bundles` and `GET /v1/bundles/:id` to include rules under the commands for each returned bundle and added a test.
* Refactored the related views in order to have a single canonical view for each resource. Rather than embedding functions like `render_rules` in `Cog.V1.BundleView`, we call `render(rule, RuleView, "rule.json")`. Note: I only applied this refactoring to views that were directly related to this change, rather than making it more globally.

### DIFF
--- a/test/controllers/v1/bundles_controller_test.exs
+++ b/test/controllers/v1/bundles_controller_test.exs
@@ -33,6 +33,27 @@ defmodule Cog.V1.BundlesControllerTest do
                            "updated_at" => "#{DateTime.to_iso8601(bundle.updated_at)}"}} == json_response(conn, 200)
   end
 
+  test "includes rules in bundle resource", %{authed: requestor} do
+    bundle = bundle("cog")
+    command = command("hola")
+    permission("cog:hola")
+    rule_text = "when command is cog:hola must have cog:hola"
+    rule = rule(rule_text)
+
+    bundle_id = bundle.id
+    command_id = command.id
+    rule_id = rule.id
+
+    conn = api_request(requestor, :get, "/v1/bundles/#{bundle.id}")
+    assert %{"bundle" => %{"id" => ^bundle_id,
+                           "commands" => [
+                             %{"id" => ^command_id,
+                               "rules" => [
+                                 %{"id" => ^rule_id,
+                                   "command" => "cog:hola",
+                                   "rule" => ^rule_text}]}]}} = json_response(conn, 200)
+  end
+
   test "cannot view bundle without permission", %{unauthed: requestor} do
     bundle = bundle("test-1")
     conn = api_request(requestor, :get, "/v1/bundles/#{bundle.id}")
--- a/web/controllers/v1/rule_controller.ex
+++ b/web/controllers/v1/rule_controller.ex
@@ -14,8 +14,7 @@ defmodule Cog.V1.RuleController do
       {:ok, rule} ->
         conn
         |> put_status(:created)
-        |> json(%{"id" => rule.id,
-                  "rule" => rule_text})
+        |> render("rule.json", rule: rule)
       {:error, errors} ->
         for error <- errors do
           Logger.error("Error ingesting \"#{rule_text}\": #{inspect error}")
@@ -28,15 +27,17 @@ defmodule Cog.V1.RuleController do
 
   def show(conn, %{"for-command" => command}) do
     rules = Repo.all(Cog.Queries.Command.rules_for_cmd(command))
-    case format_response(rules) do
-      {:ok, result} ->
-        json(conn, %{"rules" => result})
-      {:error, err} ->
+
+    case rules do
+      [] ->
         conn
         |> put_status(:unprocessable_entity)
-        |> json(%{"errors" => err})
+        |> json(%{errors: "No rules for command found"})
+      _ ->
+        render(conn, "index.json", rules: rules)
     end
   end
+
   def show(conn, params) do
     conn
     |> put_status(:unprocessable_entity)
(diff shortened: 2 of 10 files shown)


imbriaco:
cc: @lhaskins

test/controllers/v1/bundles_controller_test.exs
```diff
@@ -33,6 +33,27 @@ defmodule Cog.V1.BundlesControllerTest do
...
+
+    conn = api_request(requestor, :get, "/v1/bundles/#{bundle.id}")
+    assert %{"bundle" => %{"id" => ^bundle_id,
+                           "commands" => [
+                             %{"id" => ^command_id,
+                               "rules" => [
+                                 %{"id" => ^rule_id,
+                                   "command" => "cog:hola",
+                                   "rule" => ^rule_text}]}]}} = json_response(conn, 200)
+  end
```
lhaskins: I don't think the `assert` is necessary when using the `^` for testing the match. Though it doesn't hurt either...

web/controllers/v1/rule_controller.ex
```diff
@@ -28,15 +27,17 @@ defmodule Cog.V1.RuleController do
...
+      [] ->
         conn
         |> put_status(:unprocessable_entity)
-        |> json(%{"errors" => err})
+        |> json(%{errors: "No rules for command found"})
+      _ ->
+        render(conn, "index.json", rules: rules)
     end
   end
+
```
lhaskins: You can delete this line since these `show` functions are both arity: 2.

web/controllers/v1/rule_controller.ex
```diff
@@ -73,26 +74,4 @@ defmodule Cog.V1.RuleController do
...
-  defp build_single_expr(%Cog.Models.Rule{}=rule) do
-    ast = Piper.Permissions.Parser.json_to_rule!(rule.parse_tree)
-    %{id: rule.id,
-      command: ast.command,
-      rule: "#{ast}"}
-  end
-
-  defp format_response([]), do: {:error, "No rules for command found"}
-  defp format_response(response), do: {:ok, build_permission_expressions(response)}
-
```
lhaskins: This cleaned up very nicely!

web/views/rule_view.ex
```diff
@@ -0,0 +1,27 @@
...
+    %{rules: render_many(rules, __MODULE__, "rule.json")}
+  end
+
+  def render("show.json", %{rule: rule}) do
+    %{rule: render_one(rule, __MODULE__, "rule.json")}
+  end
+
+  defp to_ast(rule) do
+    Piper.Permissions.Parser.json_to_rule!(rule.parse_tree)
+  end
```
lhaskins: :+1: Way better than the `build_permission_expression` functions

web/views/rule_view.ex
```diff
@@ -0,0 +1,27 @@
...
+  def render("rule.json", %{rule: rule}) do
+    ast = to_ast(rule)
+    %{id: rule.id,
+      command: ast.command,
+      rule: to_string(ast)}
+  end
+
+  def render("rule_expression.json", %{rule: rule}) do
+    rule |> to_ast |> to_string
+  end
```
lhaskins: It's not clear where this is being called.

test/controllers/v1/bundles_controller_test.exs
imbriaco: Yeah, I like having it there so I can tell at a glance that this is one of the things that I'm testing. Without it you have to carefully parse the left hand side of the expression to see that it's not just an assignment.

web/views/rule_view.ex
imbriaco: Good catch! I feel like I added it thinking I was going to use it but never did. I'll delete that.

web/views/bundles_view.ex
```diff
@@ -1,36 +1,45 @@
...
+    Map.get(params, :include, [])
+    |> Enum.reduce(%{}, fn(inc, reply) -> Map.put(reply, inc, render_include(inc, bundle)) end)
   end
+
+  defp render_include(:commands, bundle) do
+    render_many(bundle.commands, __MODULE__, "command.json", as: :command)
+  end
+  defp render_include(:relay_groups, bundle) do
+    render_many(bundle.relay_groups, RelayGroupView, "relay_group.json")
+  end
```
lhaskins: Not sure how the include works with building the structure, but do you not need to put this in map form when returning it? i.e.
```elixir
%{commands: render_many(bundle.commands, __MODULE__, "command.json", as: :command)}
```

imbriaco: Nope, that happens a couple of lines up in the `render_includes` function that calls the individual includes and merges them into the reply map.

web/views/relay_group_view.ex
```diff
@@ -1,36 +1,37 @@
...
-  defp render_bundles(bundles) do
-    for bundle <- bundles do
-      %{id: bundle.id,
-        name: bundle.name}
-    end
+  defp render_include(:bundles, relay_group) do
+    render_many(relay_group.bundles, BundlesView, "bundle.json", as: :bundle)
+  end
+  defp render_include(:relays, relay_group) do
+    render_many(relay_group.relays, RelayView, "relay.json", as: :relay)
```
lhaskins: This will end up sending more data than the basics and if we aren't careful we can get into a data loop of sorts where the bundle includes the relays, which includes the bundles, which includes the relays, which [blah, blah, blah]... I like the `render_include`, but I think we may want to use it sparingly, so we don't get twisted.

web/views/bundles_view.ex
lhaskins: Ah, yes, the `Map.put` - got it! :+1:

lhaskins:
just a few comments - LGTM

web/views/relay_group_view.ex
imbriaco: Yeah, it's tough. The fear of creating a loop is exactly the reason for making the `resource.json` version of the template **not** do any includes by default, forcing you to be explicit if you want them. 

My motivation was to centralize the rendered representation of each resource so that when we return a Relay (for example) it's always the same, no matter what URL you're getting it from. It may not include children, but the core resource is the same whether it's `/v1/relays/123` or included under `/v1/relay_groups/456` as a member of a relay group.

I'm not at all sure this is the best approach. Definitely think more about this as we're working on other endpoints and we can see if a better pattern emerges.

PR #510, opened 2016-04-11
Fixups

christophermaier:
A couple fixups that snuck through the triggers work.

### DIFF
--- a/lib/cog/adapters/hipchat/api.ex
+++ b/lib/cog/adapters/hipchat/api.ex
@@ -38,6 +38,10 @@ defmodule Cog.Adapters.HipChat.API do
     end
   end
 
+  def lookup_user("@" <> handle),
+    do: lookup_user(handle: handle)
+  def lookup_user(handle) when is_binary(handle),
+    do: lookup_user(handle: handle)
   def lookup_user(id: user_id) do
     GenServer.call(__MODULE__, {:lookup_user, id: user_id})
   end
--- a/lib/cog/support/model_utilities.ex
+++ b/lib/cog/support/model_utilities.ex
@@ -104,7 +104,8 @@ defmodule Cog.Support.ModelUtilities do
     provider =  ChatProvider |> Repo.get_by!(name: String.downcase(provider))
 
     {:ok, adapter} = Cog.chat_adapter_module(provider.name)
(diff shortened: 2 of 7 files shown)


mpeck:
My cogctl tests are passing against your branch now. And your branch is passing locally for me.

web/controllers/v1/chat_handle_controller.ex
```diff
@@ -91,7 +91,7 @@ defmodule Cog.V1.ChatHandleController do
       chat_provider ->
         case Cog.chat_adapter_module(chat_provider.name) do
           {:ok, adapter} ->
-            case adapter.lookup_user(handle: handle) do
+            case adapter.lookup_user(handle) do
```
vanstee: Does this work with the slack adapter? I think we only match against `id: id` and `handle: handle`.

lib/cog/support/model_utilities.ex
```diff
@@ -104,7 +104,8 @@ defmodule Cog.Support.ModelUtilities do
     provider =  ChatProvider |> Repo.get_by!(name: String.downcase(provider))
 
     {:ok, adapter} = Cog.chat_adapter_module(provider.name)
-    {:ok, external_user} = adapter.lookup_user(handle: internal_user.username)
+    handle = adapter.mention_name(internal_user.username)
+    {:ok, external_user} = adapter.lookup_user(handle)
```
vanstee: What was the reason behind making this change? I think all the adapters expect `handle: handle` or `id: id` to be passed to `lookup_user`.

christophermaier: They did in the past, but it looks like we only deal with handles now.

lib/cog/adapters/hipchat/api.ex
```diff
@@ -38,6 +38,10 @@ defmodule Cog.Adapters.HipChat.API do
     end
   end
 
+  def lookup_user("@" <> handle),
+    do: lookup_user(handle: handle)
+  def lookup_user(handle) when is_binary(handle),
+    do: lookup_user(handle: handle)
```
vanstee: I'm not a huge fan of the different arg types here, but I can live with it. Ideally we'd either have another function like `lookup_user_by_formatted_handle` (with a shorter name) for this. I can deal with it though.

Issue #98, opened 2016-01-27
Version bundles not commands


kevsmith:
I just noticed Cog versions commands instead of bundles. This strikes me as unnecessarily complicated since bundled commands all live within the same bundle directory structure.  Versioning bundles makes more sense to me and seems as thought it'd be easier to implement.

vanstee:
Just to update this issue, we no longer version commands. But versioning bundles is still something we need to implement.

vanstee:
Also worth noting that the bundle config now includes a version number. But, we'll still need to make sure that only one version of each bundle is enabled at a time and that you can enable/disable bundle versions via cogctl.

vanstee:
For now we'll just persist the value form the bundle config and revisit the cogctl stuff in a later issue.

PR #511, opened 2016-04-11
Bundle versions are persisted when they are installed

vanstee:
:warning: Based on https://github.com/operable/cog/pull/503

We're not currently handling multiple versions of bundles at the same time, but we are now keeping track of the version installed. If you try to install another version of a previously installed bundle, you'll receive an error.

Closes https://github.com/operable/cog/issues/98

### DIFF
--- a/test/cog/rule_test.exs
+++ b/test/cog/rule_test.exs
@@ -5,7 +5,7 @@ defmodule RuleTest do
   alias Cog.Models.Rule
 
   setup do
-    bundle = Bundle.changeset(%Bundle{}, %{name: "test_bundle", config_file: %{}, manifest_file: %{}}) |> Repo.insert!
+    bundle = Bundle.changeset(%Bundle{}, %{name: "test_bundle", version: "0.0.1", config_file: %{}, manifest_file: %{}}) |> Repo.insert!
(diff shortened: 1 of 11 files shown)


mpeck:
I think if we rebase this guy it will make it a bit easier to review.

test/cog/rule_test.exs
```diff
@@ -5,7 +5,7 @@ defmodule RuleTest do
   alias Cog.Models.Rule
 
   setup do
-    bundle = Bundle.changeset(%Bundle{}, %{name: "test_bundle", config_file: %{}, manifest_file: %{}}) |> Repo.insert!
+    bundle = Bundle.changeset(%Bundle{}, %{name: "test_bundle", version: "0.0.1", config_file: %{}, manifest_file: %{}}) |> Repo.insert!
```
mpeck: We don't actually need the manifest_file bit. That was removed earlier. Just missed removing from some tests. If you want to remove it cool, if not, we can do it later.

vanstee: I'll go ahead and remove it real quick

mpeck:
:+1: Looks good

PR #515, opened 2016-04-12
Embedded commands conform to new authz model

### LINKED ISSUE #512: Update operable bundle authz management commands
The operable bundle administrative commands need to be updated to reflect the simplified authorization model that was merged in #492. I believe these are the required changes:

* `operable:group --[add|remove]` -- should only allow users to be group members
* `operable:permissions --[grant|revoke]` -- should only allow granting to a role
* `operable:role --[grant|revoke]`-- should only allow granting to a group

christophermaier:
Recently we simplified the authorization model. Now, groups may only
contain users, roles can only be granted to groups, and permissions may
only be assigned to roles.

Now, the embedded bundle commands conform to this; you can no longer add
groups to groups, grant permissions to users or groups, or grant roles
to users.

Permissions used by the commands were cleaned up, and authorization
rules were tightened up.

Fixes #512

### DIFF
--- a/lib/cog/commands/group.ex
+++ b/lib/cog/commands/group.ex
@@ -1,19 +1,17 @@
 defmodule Cog.Commands.Group do
   @moduledoc """
-  This command allows the user to manage groups. Groups can contain other groups
-  or users.
+  This command allows the user to manage groups of users.
 
   Usage:
       group --create <groupname>
       group --drop <groupname>
-      group --add [--user=<username> | --group=<groupname>] <groupname>
-      group --remove [--user=<username> | --group=<groupname>] <groupname>
+      group --add --user=<username> <groupname>
+      group --remove --user=<username> <groupname>
       group --list
   Examples:
   > @bot operable:group --create ops
   > @bot operable:group --create engineering
   > @bot operable:group --add --user=bob ops
-  > @bot operable:group --add --group=ops engineering
   > @bot operable:group --remove --user=bob ops
   > @bot operable:group --drop ops
   > @bot operable:group --list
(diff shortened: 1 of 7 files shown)


lib/cog/commands/group.ex
```diff
@@ -27,12 +25,15 @@ defmodule Cog.Commands.Group do
...
-  option "group", type: "string"
 
   permission "manage_groups"
   permission "manage_users"
 
-  rule "when command is #{Cog.embedded_bundle}:group must have #{Cog.embedded_bundle}:manage_groups"
+  rule "when command is #{Cog.embedded_bundle}:group with option[create] == true must have #{Cog.embedded_bundle}:manage_groups"
+  rule "when command is #{Cog.embedded_bundle}:group with option[drop] == true must have #{Cog.embedded_bundle}:manage_groups"
+  rule "when command is #{Cog.embedded_bundle}:group with option[list] == true must have #{Cog.embedded_bundle}:manage_groups"
+  rule "when command is #{Cog.embedded_bundle}:group with option[add] == true must have #{Cog.embedded_bundle}:manage_users"
```
mpeck: Should user have the manage_users and manage_groups permissions to do this?

lib/cog/commands/role.ex
```diff
@@ -1,18 +1,16 @@
...
   @moduledoc """
-  This command allows the user to manage roles. Roles can contain users.
+  This command allows the user to manage roles.
 
   Usage:
       role --create <rolename>
       role --drop <rolename>
-      role --grant [--user=<username>|--group=<groupname>] <rolename>
-      role --revoke [--user=<username>|--group=<groupname>]  <rolename>
+      role --grant --group=<groupname> <rolename>
```
mpeck: Since you can only grant roles to groups would it better syntax if group was an arg?

`role --grant <groupname> <rolename>`

Or is that something to revisit later?

lib/cog/commands/role.ex
```diff
@@ -25,16 +23,16 @@ defmodule Cog.Commands.Role do
...
-  permission "manage_users"
   permission "manage_groups"
 
-  rule "when command is #{Cog.embedded_bundle}:role must have #{Cog.embedded_bundle}:manage_roles"
-  rule "when command is #{Cog.embedded_bundle}:role must have #{Cog.embedded_bundle}:manage_users"
-  rule "when command is #{Cog.embedded_bundle}:role must have #{Cog.embedded_bundle}:manage_groups"
+  rule "when command is #{Cog.embedded_bundle}:role with option[create] == true must have #{Cog.embedded_bundle}:manage_roles"
+  rule "when command is #{Cog.embedded_bundle}:role with option[drop] == true must have #{Cog.embedded_bundle}:manage_roles"
+  rule "when command is #{Cog.embedded_bundle}:role with option[list] == true must have #{Cog.embedded_bundle}:manage_roles"
+  rule "when command is #{Cog.embedded_bundle}:role with option[grant] == true must have #{Cog.embedded_bundle}:manage_groups"
```
mpeck: Same as above, shouldn't we require both manage_groups and manage_roles perms here?

lib/cog/commands/group.ex
christophermaier: I believe the rationale from the long-long-ago was that by adding a user to a group, you're affecting the user more than you're affecting the group (e.g., giving them permissions they didn't have before)

@imbriaco Sound fair?

lib/cog/commands/role.ex
christophermaier: We need to do those kind of changes across all the embedded commands, so it's something for later.

imbriaco:
I think we need to more generally revisit our permission model for our admin commands -- I think it's too fine grained to be useful right now. That's out of scope for this PR though.

mpeck:
Given that I think this is good to ship. :+1:

imbriaco:
:+1:

PR #523, opened 2016-04-12
update docker tooling

imbriaco:
* Move base to Alpine Linux
* Add helper scripts for building and starting Cog
* Update docker-compose to only start Postgres and Cog. Relay management needs to change as a result of the new Relay auth model.

**Before:**

```
$ docker images operable/cog:0.3.1
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
operable/cog        0.3.1               423f59e67f17        2 weeks ago         704.9 MB
```

**After:**

```
$ docker images operable/cog:alpine
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
operable/cog        alpine              652ba4cacf0f        17 minutes ago      118.5 MB
```


Dockerfile
```diff
@@ -1,24 +1,44 @@
...
+USER root
+RUN apk update -U && \
+    apk add expat-dev gcc g++ make && \
+    # build cog and cogctl \
+    su operable - -c /home/operable/cog/scripts/docker-build && \
+    # install cogctl and delete source directory \
+    cp /home/operable/cogctl/cogctl /usr/local/bin/cogctl && \
+    rm -rf /home/operable/cogctl && \
+    # cleanup dependencies
+    apk del gcc g++ && \
```
vanstee: Could we also remove `elixir` here as well? I think only erlang is required to run the escript.

vanstee: D'oh nvm, cog needs elixir.

Issue #514, opened 2016-04-12
Roles in User endpoint


lhaskins:
The Cog API should return the roles of a user

lhaskins:
### Motivation/Objective
We need to view the roles (and permissions with those roles) for a user.

### Research
We should preload the roles for a user. There are lots of extra details in the user model that are not being used at this moment. Any references outside of the model can be removed.

### Plan
Preload the roles in the User controller. Be sure that the view pulls in and formats the output with all the necessary info.

### Tasks
- [ ] Cog: Update Controller to include preloads.
- [ ] Cog: Update Controller to accept name instead of id for #show.
- [ ] CogAPI: Update to take a name as well as an id
- [ ] CogCtl: Update commands to use API calls.
- [ ] CogCtl: Update #info to show associated roles.
- [ ] Update/add tests.
- [ ] Update/add documentation.

PR #524, opened 2016-04-13
Roles in Users

lhaskins:
This PR adds roles and their corresponding permissions into the user output. 

Fixes https://github.com/operable/cog/issues/514


imbriaco:
@lhaskins Would it make more sense for these to be nested individually under each group since that's how the relationship works?

imbriaco:
This looks fine. As we discussed in chat, I'd like to ultimately refactor the way we render resources so that we **always** render a given resource type using the same view, but that shouldn't block this PR from landing.

PR #527, opened 2016-04-13
Send early-termination output only if chat-initiated

christophermaier:
If our pipeline has "run dry" at any point, we'll send a
response back to all destinations, but only if the pipeline was
initiated from chat.

If it was initiated from a trigger, though, we don't want to
output to any chat destinations. Otherwise, you'd end up with
"mystery" messages saying "Pipeline succeeded but there was no
output!" without indication of what's going on. If we trigger
something that succeeds without output, we'll just be silent in
chat.

Additionally, we now terminate the pipeline early if output is ever
empty, instead of waiting until the end.

### DIFF
--- a/lib/cog/command/pipeline/executor.ex
+++ b/lib/cog/command/pipeline/executor.ex
@@ -144,6 +144,8 @@ defmodule Cog.Command.Pipeline.Executor do
     end
   end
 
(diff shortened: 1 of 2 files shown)


lhaskins:
LGTM

PR #529, opened 2016-04-13
make bootstrap group and role aware and allow custom user params

imbriaco:
**Updated bootstrap flow:**

* Create a new administrative user, using a parameter hash for the fields in the `%User{}` map, if present. If not, use the previous defaults.
* Create a `cog-admin` group and add the new administrative user to it.
* Create a `cog-admin` role, grant all of the `operable:*` permissions to it, and grant it to the `cog-admin` group.

### DIFF
--- a/lib/cog/bootstrap.ex
+++ b/lib/cog/bootstrap.ex
@@ -5,20 +5,25 @@ defmodule Cog.Bootstrap do
 
   alias Cog.Models.Permission.Namespace
   alias Cog.Models.User
+  alias Cog.Models.Group
+  alias Cog.Models.Role
   alias Cog.Repo
 
-  @admin_name "admin"
-
-  @site_namespace "site"
   @command_option_types ["int", "float", "string", "bool"]
+  @default_admin_params %{
+    "username" => "admin",
+    "first_name" => "Cog",
+    "last_name" => "Administrator",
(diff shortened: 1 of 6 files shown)


lib/cog/bootstrap.ex
```diff
@@ -27,52 +32,65 @@ defmodule Cog.Bootstrap do
...
+  # Create a bootstrap admin user from the given parameter map. If
+  # the password is empty, generate a random one. Returns the username
+  # and password in the response.
+  defp create_admin(params) do
+    params = case Map.get(params, "password") do
+               nil ->
+                 Map.put(params, "password", generate_safe_password)
+               _ ->
+                 params
+             end
```
vanstee: Totally not necessary, but might be nice to use `Map.put_new` or `Map.put_new_lazy` here:

```elixir
Map.put_new_lazy(params, "password", &generate_safe_password/0)
```

vanstee:
LGTM :+1:

vanstee:
Oh actually. I wonder if this will break cogctl tests because now we'll have an extra group and role after bootstrapping.

imbriaco:
It does break `cogctl` tests, but I fix them in https://github.com/operable/cogctl/pull/43. I'll merge that immediately after this one.

Issue #525, opened 2016-04-13
Add groups to the roles API endpoint


lhaskins:

lhaskins:
### Motivation/Objective
We need to view the groups that are associated with a role.

### Research
We should preload the groups associated with a role. There are lots of extra details in the user model that are not being used at this moment. Any references outside of the model can be removed.

There is definitely a potential to end up in a data loop between roles and groups, so care has to be taken to build these structures separately.

### Plan
Preload the groups in the Role controller. Be sure that the view pulls in and formats the output with all the necessary info.
The JSON format of the response will look like:
```
%{"id" => "12345678",
  "name" => "take-over",
  "groups" => [%{"id" => "98765432",
                 "name" => "robots"}],
  "permissions" => [%{"id" => "876543",
                      "name" => "world",
                      "namespace" => "site"}]
  }
```
The `roles info` command should return groups such that the output looks like the following:
```
NAME       ID
take-over  12345678

Permissions:
NAMESPACE  NAME   ID
site       world  876543

Groups:
NAME      ID
robots    98765432
```

### Tasks
- [ ] Cog: Update Controller to include preloads.
- [ ] Cog: Update Controller to accept name instead of id for #show.
- [ ] CogAPI: Update to take a name as well as an id
- [ ] CogAPI: Update the User resource to include roles
- [ ] CogCtl: Update commands to use Client API calls (instead of Internal)(delete and update).
- [ ] CogCtl: Update #info to show associated groupss.
- [ ] Update/add tests.
- [ ] Update/add documentation.

imbriaco:
This looks good.

PR #531, opened 2016-04-13
Add groups to roles

lhaskins:
This PR adds groups into the roles resource. 

Fixes https://github.com/operable/cog/issues/525

### DIFF
--- a/web/controllers/v1/role_controller.ex
+++ b/web/controllers/v1/role_controller.ex
@@ -6,9 +6,15 @@ defmodule Cog.V1.RoleController do
   plug Cog.Plug.Authentication
   plug Cog.Plug.Authorization, permission: "#{Cog.embedded_bundle}:manage_roles"
 
+  # Search by name only for now
+  def index(conn, %{"name" => name}) do
+    role = Repo.get_by!(Role, name: name)
+    |> Repo.preload([[permissions: :namespace], [group_grants: :group]])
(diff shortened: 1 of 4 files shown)


web/controllers/v1/role_controller.ex
```diff
@@ -6,9 +6,15 @@ defmodule Cog.V1.RoleController do
   plug Cog.Plug.Authentication
   plug Cog.Plug.Authorization, permission: "#{Cog.embedded_bundle}:manage_roles"
 
+  # Search by username only for now
```
mpeck: Are we searching by username or role name here?

web/views/roles_view.ex
```diff
@@ -1,28 +1,53 @@
...
-  defp render_permissions(permissions) when is_list(permissions) do
-    Enum.map(permissions, fn(permission) ->
-        %{id: permission.id,
-          name: permission.name,
-          namespace: permission.namespace.name}
+  defp render_includes(resource, role) do
+    Map.get(resource, :include, []) |> Enum.reduce(%{}, fn(field, reply) -> 
+      case render_include(field, role) do
+        nil ->
+          %{}
```
vanstee: Does this remove all includes if one include returns `nil`? I think I'd expect to see `reply` here.

mpeck:
:+1: Looks good to me

PR #538, opened 2016-04-14
Do not install non-embedded bundles

kevsmith:
This PR does a few things:

1. Prevents Cog from installing non-embedded bundles. Unknown bundles announced from external relays are treated as an error.
1. Removes the introduction step of the announcement waltz. This is no longer necessary.
1. Fixes a potentially misleading log message about relay removals.

### DIFF
--- a/lib/cog/relay/relays.ex
+++ b/lib/cog/relay/relays.ex
@@ -78,8 +78,6 @@ defmodule Cog.Relay.Relays do
     do: {:reply, Tracker.relays(state.tracker, bundle_name), state}
 
   def handle_info({:publish, @relays_discovery_topic, message}, state) do
-    # Not authenticating messages here, because we don't have the keys
-    # to authenticate them at this point!
     case Poison.decode(message) do
       {:ok, %{"announce" => announcement}} ->
         state = process_announcement(announcement, state)
--- a/lib/cog/relay/relays.ex
+++ b/lib/cog/relay/relays.ex
@@ -8,8 +8,6 @@ defmodule Cog.Relay.Relays do
   use Adz
   use GenServer
 
-  alias Carrier.Credentials
-  alias Carrier.CredentialManager
   alias Carrier.Messaging
   alias Cog.Models.Bundle
   alias Cog.Repo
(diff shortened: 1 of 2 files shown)


lib/cog/relay/relays.ex
```diff
@@ -83,8 +81,8 @@ defmodule Cog.Relay.Relays do
     # Not authenticating messages here, because we don't have the keys
```
christophermaier: I guess we can remove this comment, now, eh?

lib/cog/relay/relays.ex
```diff
@@ -96,38 +94,10 @@ defmodule Cog.Relay.Relays do
...
-        :ok
-    end
-    {:ok, my_creds} = CredentialManager.get()
-    Messaging.Connection.publish(state.mq_conn, %{intro: %{id: my_creds.id,
-                                                           role: "bot"}}, routed_by: reply_to)
-    state
-  end
-
-  defp process_announcement(announcement, %__MODULE__{tracker: tracker}=state) do
+  defp process_announcement(announcement, %__MODULE__{tracker: tracker}=state, internal \\ false) do
```
christophermaier: Could `internal` be an atom instead of a boolean? Makes for easier-to-read functions, IMO.

christophermaier: Meh... not really needed here, after reading the rest of the code. Ignore me.

christophermaier:
:shipit:

lib/cog/relay/relays.ex
```diff
@@ -181,20 +152,28 @@ defmodule Cog.Relay.Relays do
...
       nil ->
-        Logger.info("Installing bundle: #{inspect name}")
-        case Cog.Bundle.Install.install_bundle(%{name: name, version: version, config_file: config}) do
-          {:ok, bundle} ->
-            {:ok, bundle}
-          {:error, error} ->
-            Logger.error("Error! Unable to install bundle #{inspect name}: #{inspect error}")
+        case internal do
+          # This is the embedded bundle
+          true ->
```
vanstee: Since we have the entire bundle loaded here, could we check `bundle.name == Cog.embedded_bundle`, instead of passing in the `internal` variable? Or is this going to be used for other non-embedded bundle stuff eventually?

kevsmith: I wrote that first but decided against it in case we ever change the name of the embedded bundle. Setting the `internal` flag is driven purely by how the announcement is received -- MQTT or GenServer.call -- which seemed to me to be less likely to trip us up until we have time to eliminate the embedded bundle's special snowflake status.

lib/cog/relay/relays.ex
kevsmith: Ah yes!

PR #539, opened 2016-04-14
Rename a trigger-related environment variable

christophermaier:
This was leftover from the rename of "event hooks" to "triggers"


vanstee:
:shipit:

PR #544, opened 2016-04-15
Implement Groupable for relay groups

mpeck:
We were crashing when trying to add bundles that already existed in the group to relay groups. This rescues from the crash and returns a message to the user.

### DIFF
--- a/web/controllers/v1/relay_group_membership_controller.ex
+++ b/web/controllers/v1/relay_group_membership_controller.ex
@@ -29,7 +29,17 @@ defmodule Cog.V1.RelayGroupMembershipController do
 
       case to_add.not_found ++ to_remove.not_found do
         [] ->
-          RelayGroup.add!(type, group.id, to_add.found)
+          try do
+            RelayGroup.add!(type, group.id, to_add.found)
+          rescue
+            err in [Ecto.InvalidChangesetError] ->
+              case err.changeset.errors do
+                [bundle_id: _] ->
+                  Repo.rollback({:add_failed, "Bundle already exists in relay group."})
+                _errors ->
+                  Repo.rollback({:add_failed, "Failed to add bundle to relay group."})
+              end
+          end
           removed = RelayGroup.remove(type, group.id, to_remove.found)
           if removed != length(to_remove.found) do
             Repo.rollback(:remove_failed)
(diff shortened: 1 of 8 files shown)


web/router.ex
```diff
@@ -48,10 +48,10 @@ defmodule Cog.Router do
...
     resources "/v1/relays", V1.RelayController
     resources "/v1/relay_groups", V1.RelayGroupController
-    get "/v1/relay_groups/:id/relays", V1.RelayGroupMembershipController, :index
-    post "/v1/relay_groups/:id/membership", V1.RelayGroupMembershipController, :manage_membership
-    get "/v1/relay_groups/:id/bundles", V1.RelayGroupMembershipController, :index
-    post "/v1/relay_groups/:id/assignment", V1.RelayGroupMembershipController, :manage_assignment
+    get "/v1/relay_groups/:id/relays", V1.RelayGroupMembershipController, :relay_index
+    post "/v1/relay_groups/:id/relays", V1.RelayGroupMembershipController, :manage_relay_membership
+    get "/v1/relay_groups/:id/bundles", V1.RelayGroupMembershipController, :bundle_index
+    post "/v1/relay_groups/:id/bundles", V1.RelayGroupMembershipController, :manage_bundle_assignment
```
vanstee: For convention I would probably expect these two to be under separate controllers `RelayGroupRelaysController` and `RelayGroupBundlesController`, but this is probably fine.

web/controllers/v1/relay_group_membership_controller.ex
```diff
@@ -2,83 +2,130 @@ defmodule Cog.V1.RelayGroupMembershipController do
...
+  defp lookup_all(_, []), do: {:ok, []} # Don't bother with a DB lookup
+  defp lookup_all(kind, ids) when kind in ["relays", "bundles"] do
+
+    type = kind_to_type(kind) # e.g. "relays" -> Relay
+
+    # Since we are using ids we need to make sure that they are all valid UUIDs
+    # before we query the db. Otherwise Ecto with crash with a CastError
+    case good_ids?(ids) do
+      true ->
+        results = Repo.all(from t in type, where: field(t, :id) in ^ids)
```
vanstee: Do you need to use `field` here?

vanstee:
LGTM

Issue #540, opened 2016-04-15
Add permissions to bundles endpoint


lhaskins:
### Motivation/Objective
The resource should include what permissions are necessary in order to use the commands in a particular bundle.
 
### Research/Plan
Include the permissions for each command in a bundle. The structure should continue to be consistent with the structure of the bundles now. That is,
```
{"bundle":{
 "updated_at":"2016-04-15T19:32:00Z",
 "relay_groups":[],
 "name":"my_bundle",
 "inserted_at":"2016-04-15T19:32:00Z",
 "id":"123456789",
 "enabled":true,
 "commands":[
     {"name":"cmd",
     "id":"3456789",
     "execution":"multiple",
     "enforcing":false,
     "rules":[
       {"rule":"when command is my_bundle:cmd must have my_bundle:my_perm",
       "permissions":[{"namespace":"my_bundle",
                       "name":"my_perm",
                       "id":"876543}],
       "id":"264847",
       "command":"my_bundle:cmd"}],
   "documentation":"Some docs"},
}]}```

### Tasks
- [ ] Add permissions preloads in the queries.
- [ ] Add a new view for commands.
- [ ] Update the view for bundles.
- [ ] Update/add tests.
- [ ] Update/add documentation.

PR #545, opened 2016-04-15
Add permissions to bundles

lhaskins:
This change adds permissions to the bundle under the command/rules path.

This PR also adds a CommandView for the command structure.

Fixes https://github.com/operable/cog/issues/540


imbriaco:
@lhaskins Can you add a quick test to make sure we don't regress on this? Otherwise looks good to me.

lhaskins:
Sure! I thought I had one. I wonder if it got lost in all the rebases... I'll add it back, though.

PR #547, opened 2016-04-18
Add some slack to trigger timeouts

christophermaier:
Trigger timeouts are defined according to the needs of the
requestor (e.g. Github webhooks must respond within 30 seconds), which
includes network roundtrip time, as well as Cog's internal
processing. Cog itself can't wait that long to respond, as that'll be
guaranteed to exceed the HTTP requestor's timeout. As such, we'll bake
some slack into our internal timeout, allowing us to return to the
requestor within their timeout.

This can be configured on a system-wide basis.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -133,4 +133,11 @@ config :cog, Cog.TriggerEndpoint,
 config :cog, :token_lifetime, {1, :week}
 config :cog, :token_reap_period, {1, :day}
 
+# Trigger timeouts are defined according to the needs of the
+# requestor, which includes network roundtrip time, as well as Cog's
+# internal processing. Cog itself can't wait that long to respond, as
+# that'll be guaranteed to exceed the HTTP requestor's timeout. As
+# such, we'll incorporate a buffer into our internal timeout.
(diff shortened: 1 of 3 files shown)


config/config.exs
```diff
@@ -133,4 +133,11 @@ config :cog, Cog.TriggerEndpoint,
 config :cog, :token_lifetime, {1, :week}
 config :cog, :token_reap_period, {1, :day}
 
+# Trigger timeouts are defined according to the needs of the
+# requestor, which includes network roundtrip time, as well as Cog's
+# internal processing. Cog itself can't wait that long to respond, as
+# that'll be guaranteed to exceed the HTTP requestor's timeout. As
+# such, we'll bake some slack into our internal timeout.
+config :cog, :trigger_timeout_slack_sec, (System.get_env("COG_TRIGGER_TIMEOUT_SLACK_SEC") || 2)
+
```
lhaskins: To avoid possible confusion, can you change the variable names to not use "slack", but maybe "buffer"?

christophermaier: @lhaskins Heh, excellent suggestion 😄 Coming right up.

PR #549, opened 2016-04-18
Add API error design guide

drapergeek:
Hello Coggers!

One of the things we've ran into a few times is trying to properly parse and display user error messages in flywheel. This is my first pass at trying to define a basic starting style guide for how to format error messages so we can make sure display the properly.

If you have any thoughts or ideas I'd love to hear them!

### DIFF
--- a/docs/api/errors.md
+++ b/docs/api/errors.md
@@ -37,3 +37,45 @@ The following fields are required in the object, but can be empty:
 
 * `url` - a link that could be provided to the user for clarification on how to
   resolve the error
+
+### Example
+
+#### Normal Error, not associated with a param
+
+```javascript
+{
+  "errors" => [
+    {
+      "code" => "cog_error",
+      "message" => "Your token has expired, please reauthenticate.",
+      "id" => "",
+      "url" => "https://docs.operable.io/kb/made-up-url-do-not-click",
+      "param" => "",
+    }
+  ]
+}
+
+```
+
+#### Form Error
+
+```javascript
+{
+  "errors" => [
+    {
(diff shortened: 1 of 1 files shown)


mpeck:
@drapergeek can you include an example of what a response would look like?

drapergeek:
@mpeck added a few examples, let me know if that helps.

docs/api/errors.md
```diff
@@ -0,0 +1,81 @@
...
+* `param` - if the error is directly tied to an attribute on a resource, include the
+  resource attribute. For example, if user creation fails because the user's email
+  is invalid, then `param` would be "email".
+
+  If the param is given, you can assume that the message will be displayed
+  alongside the param so it does not have to be included in the message. Eg:
+  "Email is invalid". In this case, `param` would be "email" and `message`
+  would be "is invalid".
+
+* `id` - a unique identifier to the error
```
mpeck: So what is the purpose of id? Does this mean that we should persist errors?

drapergeek: I think this is more to the type of error. For example, an error that was returned for being unable to delete the `operable` bundle.

mpeck:
I like the idea of a style guide for things like this. I'm still a bit foggy on what the error id would be used for, but for the most part I dig it. I would like to get some other folks opinions too though.

drapergeek:
>I like the idea of a style guide for things like this. I'm still a bit foggy on what the error id would be used for, but for the most part I dig it. I would like to get some other folks opinions too though.

I'm fine with removing the `id`. This was gathered from a few various sources.

PR #550, opened 2016-04-18
Apply bundle attributes that are not in config during creation

vanstee:
This changes the expected json posted fora  bundle create to be:

```
{
  "bundle": {
    "config": ...config stuff...,
    "enabled": true
  }
}
```

where we used to only allow bundle config like:

```
{
  "bundle": ...config stuff...
}
```

This allows us to easily set bundle attributes on create that aren't present in the config, which is used in https://github.com/operable/cogctl/pull/47

### DIFF
--- a/test/controllers/v1/bundles_controller_test.exs
+++ b/test/controllers/v1/bundles_controller_test.exs
@@ -43,7 +43,7 @@ defmodule Cog.V1.BundlesControllerTest do
       upload = %Plug.Upload{path: config_path, filename: config_file_name}
 
       # upload the config
-      conn = api_request(requestor, :post, "/v1/bundles", body: %{bundle_config: upload}, content_type: :multipart)
+      conn = api_request(requestor, :post, "/v1/bundles", body: %{bundle: %{config_file: upload}}, content_type: :multipart)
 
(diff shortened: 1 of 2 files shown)


mpeck:
:+1: :shipit:

test/controllers/v1/bundles_controller_test.exs
```diff
@@ -90,14 +90,14 @@ defmodule Cog.V1.BundlesControllerTest do
...
-    conn = api_request(requestor, :post, "/v1/bundles", body: %{bundle_config: upload}, content_type: :multipart)
+    conn = api_request(requestor, :post, "/v1/bundles", body: %{bundle: %{config_file: upload}}, content_type: :multipart)
 
     assert conn.status == 422
   end
 
   test "accepts a valid config passed as json", %{authed: requestor} do
     config = config(:map)
-    conn = api_request(requestor, :post, "/v1/bundles", body: %{"bundle" => config})
+    conn = api_request(requestor, :post, "/v1/bundles", body: %{"bundle" => %{"config" => config}})
```
christophermaier: Is `"config"` supposed to be `"config_file"` here?

christophermaier:
👍

PR #551, opened 2016-04-18
prevent rename or delete of cog-admin group and role

imbriaco:
* adds  constraint triggers to Postgres to prevent the cog-admin group and role and the operable bundle from being deleted or renamed.
* adds a changeset function to validate deletes of Group and Role models.
* updates changeset functions in Group and Role to reject name changes for the admin group and role
* updates the Group and Role API controllers to use the new changesets to validate deletes.

Note, this **does not** update the commands in the embedded bundle to use a changeset, so attempts to delete protected objects will result in a crash of the command when Postgrex raises an exception. These should be cleaned up in a future pull request.

### DIFF
--- a/lib/cog/models/group.ex
+++ b/lib/cog/models/group.ex
@@ -4,6 +4,7 @@ defmodule Cog.Models.Group do
 
   alias Cog.Models.UserGroupMembership
   alias Cog.Models.GroupGroupMembership
+  alias Ecto.Changeset
 
   schema "groups" do
     field :name, :string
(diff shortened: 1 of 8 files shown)


lib/cog/models/group.ex
```diff
@@ -28,12 +29,30 @@ defmodule Cog.Models.Group do
...
+  @doc """
+  Creates a changeset based on the `model` to validate a delete
+  action.
+  """
+  def changeset(model, :delete) do
+    changeset = %{Changeset.change(model) | action: :delete}
+
+    if model.name == Cog.admin_group do
+      changeset
+      |> add_error(:name, "admin group may not be modified")
```
christophermaier: Might want to change the error message to `"admin group may not be deleted"`.

Of course, since we're checking against a specific name, we probably also need to prevent the name from being changed, as well.

lib/cog/models/role.ex
```diff
@@ -18,7 +20,31 @@ defmodule Cog.Models.Role do
...
+  @doc """
+  Creates a changeset based on the `model` to validate a delete
+  action.
+  """
+  def changeset(model, :delete) do
+    changeset = %{Changeset.change(model) | action: :delete}
+
+    if model.name == Cog.admin_role do
+      changeset
+      |> add_error(:name, "admin role may not be modified")
```
christophermaier: Same comment above applies here, too.

lib/cog/models/group.ex
christophermaier: Never mind... this is done in the triggers.

imbriaco: It's a good point though, I should update the changeset to give better error feedback even if the real protection happens in the trigger. I'll take a peek at that quickly.

christophermaier: We could have changeset functions that bail if you try to delete or rename a protected "thing", which would provide protection in Elixir, since recovering from trigger-thrown exceptions in a nice customizable way with Ecto / Postgrex is not as straightforward as I'd like (at least, it wasn't back when I tried to do it a few months ago; not sure if anything has changed on that front since then).

Triggers could provide the last line of defence, should we ever hit the database without going through the app.

imbriaco: Right, that's exactly what I attempted to do with the `changeset(model, :delete)` functions, so that we'd bail in Elixir before we hit the Postgrex exception. I also couldn't find a clean way to turn those into something useful to a user, unfortunately. For this PR, I'm going to update the changesets to bail when attempting to edit the name of the cog-admin role or group and call this one done.

christophermaier: Sounds good 👍  :shipit:

PR #552, opened 2016-04-18
Replace slack token with fake token in cassettes

vanstee:


christophermaier:
Do we need to change any tokens we use in tests to be `"xoxb-filtered-token"` now?

christophermaier:
Also, are there any implications for how we re-create / add new cassettes now?

vanstee:
Nope. Exvcr still matches requests correctly even though we're using the real Slack token while making those requests. We're just no longer writing that real token to the recorded cassette. Everything else stays the same.

christophermaier:
@vanstee Sweet... :shipit:

PR #559, opened 2016-04-21
Include *all* permissions in a bundle

lhaskins:
Include all the possible permissions in a bundle and not only the permissions used in the bundle rules.

Connected to https://github.com/operable/cog/issues/540

### DIFF
--- a/web/controllers/v1/bundles_controller.ex
+++ b/web/controllers/v1/bundles_controller.ex
@@ -42,7 +42,7 @@ defmodule Cog.V1.BundlesController do
   def create(conn, %{"bundle" => params}) do
     case install_bundle(params) do
       {:ok, bundle} ->
-        bundle = Repo.preload(bundle, commands: [rules: [permissions: :namespace]])
+        bundle = Repo.preload(bundle, [commands: [rules: [permissions: :namespace]], namespace: [permissions: :namespace]])
         conn
         |> put_status(:created)
         |> put_resp_header("location", bundles_path(conn, :show, bundle))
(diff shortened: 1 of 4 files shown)


web/views/bundles_view.ex
```diff
@@ -14,11 +15,11 @@ defmodule Cog.V1.BundlesView do
   end
 
   def render("index.json", %{bundles: bundles}) do
-    %{bundles: render_many(bundles, __MODULE__, "bundle.json", as: :bundle, include: [:commands, :relay_groups])}
+    %{bundles: render_many(bundles, __MODULE__, "bundle.json", as: :bundle, include: [:commands, :relay_groups, :namespace])}
```
vanstee: Nice, I really like how these views work with the dynamic `include`. :+1:

web/controllers/v1/bundles_controller.ex
```diff
@@ -42,7 +42,7 @@ defmodule Cog.V1.BundlesController do
   def create(conn, %{"bundle" => params}) do
     case install_bundle(params) do
       {:ok, bundle} ->
-        bundle = Repo.preload(bundle, commands: [rules: [permissions: :namespace]])
+        bundle = Repo.preload(bundle, [commands: [rules: [permissions: :namespace]], namespace: [permissions: :namespace]])
```
vanstee: It's really a bummer that we have to define both sides of the `namespace` relation for `permissions`. I think that was fixed in Ecto 2.0 but we'll have to wait a bit.

vanstee:
Looks good :+1:

PR #560, opened 2016-04-21
allow a relay to be created with a user-defined ID

imbriaco:
Currently Relays require an ID to be set at startup time, but that ID is not known until after the Relay resource is created on Cog. This change allows the ID for a Relay to be included in the API call for the initial relay creation, which will make it easier for users to bootstrap an entire Cog cluster from scratch since a relay ID can be defined by the user in advance..

* Allow `POST /v1/relays` to include an ID. 
* Add tests to ensure that we don't allow the ID to be changed after create.

### DIFF
--- a/lib/cog/models/relay.ex
+++ b/lib/cog/models/relay.ex
@@ -24,9 +24,22 @@ defmodule Cog.Models.Relay do
     model
     |> Repo.preload(:groups)
     |> cast(params, @required_fields, @optional_fields)
+    |> allow_user_defined_id_on_insert(params)
     |> validate_presence_on_insert(:token)
     |> encode_token
     |> unique_constraint(:name)
+  end
+
+  def allow_user_defined_id_on_insert(changeset, params) do
+    case {get_field(changeset, :id), Map.get(params, "id")} do
+      {_, nil} ->
+        changeset
+      {nil, _} ->
+        changeset |> cast(params, [], [:id])
+      _ ->
+        changeset
+        |> add_error(:id, "cannot modify ID")
+    end
   end
 
   def validate_presence_on_insert(changeset, :token) do
--- a/lib/cog/models/relay.ex
+++ b/lib/cog/models/relay.ex
@@ -18,24 +18,24 @@ defmodule Cog.Models.Relay do
   end
 
   @required_fields ~w(name)
-  @optional_fields ~w(token enabled description)
+  @optional_fields ~w(id token enabled description)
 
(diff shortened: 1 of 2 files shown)


lib/cog/models/relay.ex
```diff
@@ -24,11 +24,24 @@ defmodule Cog.Models.Relay do
...
     |> Repo.preload(:groups)
     |> cast(params, @required_fields, @optional_fields)
+    |> allow_user_defined_id_on_insert(params)
     |> validate_presence_on_insert(:token)
     |> encode_token
     |> unique_constraint(:name)
   end
 
+  def allow_user_defined_id_on_insert(changeset, params) do
+    case {get_field(changeset, :id), Map.get(params, "id")} do
```
vanstee: If we add `:id` to `@optional_fields` can we use `get_change` here instead of passing in params? Sounded like you may have already tried that yesterday, but just wanted to make sure.

vanstee: This seemed to work:

```elixir
def allow_user_defined_id_on_insert(changeset) do
  case {Map.get(changeset.model, :id), get_change(changeset, :id)} do
    {nil, _changed_id} ->
      changeset
    {id, id} ->
      changeset
    _ ->
      changeset
      |> add_error(:id, "cannot modify ID")
  end
end
```

mpeck: @vanstee will `get_change(changeset, :id)` return nil if we are doing an update? I don't think the id is included in changes on update. If that's the case I think we would get a "cannot modify ID" failure. 

Would `get_field(changeset, :id)` work better?

vanstee: Ah I think you're right. Looks like I forgot to test updating without an id present.

vanstee: Gonna go ahead and make this change and merge since it's block the cogctl peice.

imbriaco: Yeah, the reason I didn't include it in `@optional_fields` is that it's semantically different in that it's not actually optional if you're doing an update rather than a create. I'm not married to the way I did it, I was just trying to be more explicitly intention revealing.

vanstee: Ah good point. `@optional_fields` is poorly named. It really means `@whitelisted_fields`. Another reason to switch was to avoid having to check for both `:id` and `"id"` in params. I pushed the change so I'll just merge that. This will change with Ecto 2.0 anyway since they handle whitelisting and requiring params differently.

PR #561, opened 2016-04-21
Rename trigger_timeout_buffer config flag

christophermaier:
Noticed we still had an environment variable named `"SLACK"` and not
`"BUFFER"`. Also removed the `_sec` suffix.


vanstee:
LGTM

mpeck:
:+1:

PR #562, opened 2016-04-21
Hardcode and comment out extra env vars

vanstee:
When starting up cog with docker-compose we had a lot of warnings and config that you had to worry about. Now, with this PR, you just have to set `SLACK_API_TOKEN` and you're done. All other config that you might want is currently commented out but is easy to include.

This mainly allows us to simply write `SLACK_API_TOKEN="..." docker-compose up` as the command to run in the docs.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,36 +4,28 @@
 # The following variables must be set in the shell that docker-compose
 # is executed from:
 #
-# POSTGRES_USER, POSTGRES_PASSWORD, SLACK_API_TOKEN
-#
-# In addition, a TAG variable may be defined. If present, it will be appended
-# to the Docker image names that are used for Cog and Relay. It should begin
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -4,36 +4,28 @@
...
   image: postgres:9.5
-  environment:
-    - POSTGRES_USER=${POSTGRES_USER}
-    - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
+  # environment:
+  #   - POSTGRES_USER=${POSTGRES_USER}
+  #   - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
 cog:
-  image: operable/cog${TAG}
+  image: 2ba2e73c6b43
```
vanstee: Forgot to change this back. Fixing.

vanstee:
Closing in favor of @christophermaier's work.

PR #568, opened 2016-04-22
Auto create a relay after starting it via docker

vanstee:


docker-compose-example.yml
```diff
@@ -38,7 +38,7 @@ postgres:
     - POSTGRES_USER=cog
     - POSTGRES_PASSWORD=cog
 cog:
-  image: operable/cog:0.5-dev
+  image: operable/cog:0.4-rc1
```
vanstee: D'oh, I'll need to do some fixing here if we want to merge this.

vanstee:
Closing this in favor of explaining it in the docs.

PR #571, opened 2016-04-22
Unregistered relays bug

mpeck:
If a relay wasn't registered with Cog and it requested it's list of
bundles, Cog would try to return a payload of an empty list. This
would cause Carrier to crash. Now we just ignore messages from
unregistered relays.

### DIFF
--- a/lib/cog/relay/info.ex
+++ b/lib/cog/relay/info.ex
@@ -69,6 +69,9 @@ defmodule Cog.Relay.Info do
         respond(%{bundles: bundles}, reply_to, state)
       nil ->
         ## If we get a nil back then the relay isn't registered with Cog.
+        ## Technically we should never respond with an error, because relays
+        ## should never make it through the BusEnforcer if they aren't registered
+        ## but for completeness it's included here.
         respond(%{error: "Relay with id #{relay_id} was not recognized."}, reply_to, state)
     end
   end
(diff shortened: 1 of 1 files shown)


lib/cog/relay/info.ex
```diff
@@ -46,24 +55,22 @@ defmodule Cog.Relay.Info do
...
+        relay = Repo.preload(relay, [groups: :bundles])
+
+        bundles = get_in(relay.groups, [all, :bundles])
+        |> Enum.map(&Map.take(&1, [:name, :config_file, :enabled]))
 
-    %{bundles: prepared_bundles}
+        respond(%{bundles: bundles}, reply_to, state)
+      nil ->
+        ## If we get a nil back then the relay isn't registered with Cog.
+        respond(%{error: "Relay with id #{relay_id} was not recognized."}, reply_to, state)
```
christophermaier: Do you know what will happen on the relay side of things when it receives this? 

Also, if a proper relay should never get here, we should add that in a comment, as well as an explanation of why.

mpeck: I don't @christophermaier since we never actually hit this point in the code with a real relay. I suppose I could dink with bits to force it to happen, but I'm not sure that's worth doing here.

christophermaier:
👍  :shipit:

PR #572, opened 2016-04-23
Adds support for enabling and Disabling relays

mpeck:
This adds a couple functions to the `Cog.Relay.Relays` module to enable and disable relays.

### DIFF
--- a/lib/cog/relay/relays.ex
+++ b/lib/cog/relay/relays.ex
@@ -33,6 +33,14 @@ defmodule Cog.Relay.Relays do
     GenServer.call(__MODULE__, {:drop_bundle, bundle}, :infinity)
   end
 
+  def enable_relay(relay_id) do
+    GenServer.call(__MODULE__, {:enable_relay, relay_id}, :infinity)
+  end
+
+  def disable_relay(relay_id) do
+    GenServer.call(__MODULE__, {:disable_relay, relay_id}, :infinity)
+  end
+
   @doc """
   Returns the IDs of all Relays currently running `bundle_name`. If no
   Relays are running the bundle, an empty list is returned.
@@ -76,6 +84,8 @@ defmodule Cog.Relay.Relays do
   end
   def handle_call({:relays_running, bundle_name} , _from, state),
     do: {:reply, Tracker.relays(state.tracker, bundle_name), state}
+  def handle_call({:enable_relay, relay_id}, _from, state),
+    do: {:reply, enable_relay(:snapshot, state.tracker, relay_id, []), state}
 
   def handle_info({:publish, @relays_discovery_topic, message}, state) do
     case Poison.decode(message) do
@@ -128,22 +138,23 @@ defmodule Cog.Relay.Relays do
                       false -> :offline
                     end
 
+    enabled_status = case Cog.Repo.get(Cog.Models.Relay, relay_id) do
+                       %{enabled: true} -> :enabled
+                       _ -> :disabled
+                     end
+
     snapshot_status = case Map.fetch!(announcement, "snapshot") do
                         true -> :snapshot
                         false -> :incremental
                       end
 
-    bundle_names = Enum.map(success_bundles, &Map.get(&1, :name)) # Just for logging purposes
-    case {online_status, snapshot_status} do
+    case {online_status, enabled_status} do
       {:offline, _} ->
-        Logger.info("Removed Relay #{relay_id} from active relay list")
-        Tracker.remove_relay(tracker, relay_id)
-      {:online, :incremental} ->
-        Logger.info("Incrementally adding bundles for Relay #{relay_id}: #{inspect bundle_names}")
-        Tracker.add_bundles_for_relay(tracker, relay_id, success_bundles)
-      {:online, :snapshot} ->
-        Logger.info("Setting bundles list for Relay #{relay_id}: #{inspect bundle_names}")
-        Tracker.set_bundles_for_relay(tracker, relay_id, success_bundles)
+        disable_relay(tracker, relay_id)
+      {:online, :disabled} ->
+        disable_relay(tracker, relay_id)
+      {:online, :enabled} ->
+        enable_relay(snapshot_status, tracker, relay_id, success_bundles)
     end
   end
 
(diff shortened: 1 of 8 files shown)


lib/cog/relay/relays.ex
```diff
@@ -120,30 +141,33 @@ defmodule Cog.Relay.Relays do
...
-  defp update_tracker(announcement, tracker, success_bundles) do
+  defp update_tracker(announcement, tracker, success_bundles, internal) do
     relay_id = Map.fetch!(announcement, "relay")
 
     online_status = case Map.fetch!(announcement, "online") do
                       true -> :online
                       false -> :offline
                     end
 
+    enabled_status = cond do
```
christophermaier: I think an `if` would be a bit more straightforward than a `cond` for this.

lib/cog/relay/tracker.ex
```diff
@@ -22,6 +28,36 @@ defmodule Cog.Relay.Tracker do
     do: %__MODULE__{}
 
   @doc """
+  Enables a relay if it exists in the disabled set by removing it from the
+  disabled set. When the list of relays for a bundle is requested, disabled
+  bundles are filtered out.
+  """
+  @spec enable_relay(t, String.t) :: t
+  def enable_relay(tracker, relay_id) do
+    if is_disabled?(tracker, relay_id) do
```
christophermaier: Deleting from a map is idempotent, so you don't even need the `if` here.

lib/cog/relay/tracker.ex
```diff
@@ -81,7 +119,20 @@ defmodule Cog.Relay.Tracker do
   def relays(tracker, bundle_name) do
     tracker.map
     |> Map.get(bundle_name, MapSet.new)
+    |> MapSet.difference(tracker.disabled)
     |> MapSet.to_list
   end
 
+  defp in_tracker?(tracker, relay_id) do
+    relays = Enum.reduce(tracker.map, MapSet.new(), fn({_, relays}, acc) ->
```
christophermaier: Might be able to simplify this a bit and make the whole thing a pipeline by calling `Map.values(tracker.map)` first.

test/controllers/v1/relay_controller_test.exs
```diff
@@ -95,6 +98,72 @@ defmodule Cog.V1.RelayControllerTest do
...
   end
 
+  test "relays are enabled in more than just name", %{authed: requestor} do
+    # We create a relay and add a bundle to it so we can query for it in
+    # 'Cog.Relay.Relays'
+    relay = relay("relay-enable-relay-1", "foobar")
+    bundle = bundle("relay-enable-bundle-1")
+    relay_group = relay_group("relay-enable-group-1")
+    add_relay_to_group(relay_group.id, relay.id)
+    assign_bundle_to_group(relay_group.id, bundle.id)
```
christophermaier: How about making some helper methods that can do all this in one shot, like we now do with `cogctl`.

I definitely think we should still have the fine-grained helpers that you've got here, but I think it'd help the tests to put a lot of that tedious setup logic behind some well-named helpers that can do it all in one shot.

test/controllers/v1/relay_controller_test.exs
```diff
@@ -95,6 +98,72 @@ defmodule Cog.V1.RelayControllerTest do
...
+
+    # We shouldn't see any relays running our bundle yet, because the relay
+    # has not yet announced it's presence.
+    assert Relays.relays_running(bundle.name) == []
+
+    # Relays don't show up as available unless they are online and enabled.
+    # FakeRelay lets us send announcement messages like a real relay, so Cog
+    # will add the relay to the available relays list.
+    FakeRelay.new(relay) |> FakeRelay.get_bundles |> FakeRelay.announce
```
christophermaier: Similarly, I wonder if some of these could be coalesced behind a single function call.

test/controllers/v1/relay_controller_test.exs
```diff
@@ -95,6 +98,72 @@ defmodule Cog.V1.RelayControllerTest do
...
+
+    # This should enable our relay
+    conn = api_request(requestor, :put, "/v1/relays/#{relay.id}",
+                       body: %{"relay" => @update_attrs})
+    # Confirm that the api thinks the relay is enabled
+    updated = json_response(conn, 200)["relay"]
+    assert updated["enabled"] == @update_attrs.enabled
+
+    # Now if we check for relays_running we should see our relay
+    assert Relays.relays_running(bundle.name) == [relay.id]
```
christophermaier: I really like this test 👍

web/controllers/v1/relay_controller.ex
```diff
@@ -46,6 +46,10 @@ defmodule Cog.V1.RelayController do
     changeset = Relay.changeset(relay, relay_params)
     case Repo.update(changeset) do
       {:ok, relay} ->
+        # If the enabled flag has changed we need to enable/disable the relay
```
christophermaier: I think this functionality needs to move out of the controller and closer to the core. That is, whenever we update a relay from _anywhere_, we should be sure that the state of the system is consistent.

One way to do it is to put the database operations (like `Repo.update`, here) behind another Cog-specific interface, which acts as a DDD-style "repository". The only way to update a relay would be to go through that interface, and behind it, we're free to do whatever we want. In this case, we can just make the call to update the status, but in the future it opens up more possibilities (e.g., emitting events that arbitrary system components can listen for and then take action on, etc.).

I started this with triggers  - https://github.com/operable/cog/blob/master/lib/cog/repository/triggers.ex - thinking that each "thing" could perhaps have it's own "repository". That could ultimately change, but it seemed like a decent way to decompose things to start with.

lib/cog/support/model_utilities.ex
```diff
@@ -231,6 +276,22 @@ defmodule Cog.Support.ModelUtilities do
   end
 
   @doc """
+  Creates a relay, relay-group and bundle. Then assigns the bundle and adds the
+  relay to the relay-group
+  """
+  def create_relay_bundle_and_group(name, opts \\ []) do
+    relay = relay("relay-#{name}",
+                  Keyword.get(opts, :token, "sekrit"),
```
christophermaier: Documenting the options each test function takes would be nice

web/controllers/v1/relay_controller.ex
```diff
@@ -1,28 +1,23 @@
 defmodule Cog.V1.RelayController do
   use Cog.Web, :controller
 
-  alias Cog.Models.Relay
-  alias Cog.Queries
-  alias Cog.Repo
+  alias Cog.Repository.Relays
```
christophermaier: What do you think about changing this alias to at least `Repository.Relays`? I'm concerned about readers getting potentially confused by just seeing `Relays` and thinking of `Cog.Relay.Relays`.

christophermaier:
👍

web/controllers/v1/relay_controller.ex
mpeck: That's fair. I was actually thinking the same thing.

mpeck: Actually what do think about using `as:` instead? `alias Cog.Repository.Relays, as: RelaysRepo`

christophermaier: That's cool too 👍

PR #581, opened 2016-04-26
Clean up command documentation

jsteiner:
Fixes indentation for markdown rendering and makes the headers and
formatting more consistent.


vanstee:
We still have a lot of work to do here but this is a great start. Thanks @jsteiner

PR #583, opened 2016-04-27
Bring spanner home

christophermaier:
A big chunk of Spanner is now only used here in Cog; it should live in Cog.


lib/cog/gen_command.ex
```diff
@@ -0,0 +1,251 @@
+defmodule Cog.GenCommand do
```
kevsmith: Should be nested under `Cog.Command`?

lib/cog/bundle/config.ex
```diff
@@ -0,0 +1,177 @@
...
+    invocation name, the Elixir module that implements it, and the
+    various options the command may take
+  - A list of permissions the bundle will create
+  - A list of initial rules for the commands in the bundle, using the
+    bundle permissions.
+
+  ## Example
+
+      %{bundle: %{name: "foo"},
+        commands: [%{module: "Spanner.Commands.AddRule",
```
vanstee: Do we want to update these example modules? Might be confusing.

lib/cog/command/request.ex
```diff
@@ -0,0 +1,80 @@
...
+        {:ok, config}
+      {:error, error} ->
+        err = "Unable to read the command config file for the command '#{request.command}'. #{inspect error}"
+        Logger.error(err)
+        {:error, err}
+    end
+  end
+
+  defp open_config(request) do
+    case Application.get_env(:spanner, :command_config_root) do
```
vanstee: Are we going to leave `:spanner` as the config for now?

vanstee:
LGTM

lib/cog/bundle/config.ex
christophermaier: D'oh

kevsmith:
👍

PR #586, opened 2016-04-27
Initial service tokens implementation

christophermaier:


lib/cog/command/service/tokens.ex
```diff
@@ -0,0 +1,82 @@
+defmodule Cog.Command.Service.Tokens do
```
kevsmith: Wondering if `Command` and `Service` should be peers instead of nested...

`Cog.Command.Service.Tokens` vs `Cog.Service.Tokens`

PR #587, opened 2016-04-27
Create a memory service process

christophermaier:


test/cog/command/service/memory_test.exs
```diff
@@ -0,0 +1,99 @@
...
+
+    Memory.replace(token, "bbq", "valentina's tex mex")
+
+    send(pid, :crash)
+    :timer.sleep(500)
+
+    assert {:error, :unknown_key} = Memory.fetch(token, "bbq")
+  end
+
+  defp fake_executor do
```
christophermaier: Looks like maybe we need a helper module for this now?

vanstee:
This creates a memory service with the following operations: `fetch`, `accum`, `join`, `replace`, `delete`. You can read the docs for the semantics of each. We'll expose this via API routes in a future PR.

lib/cog/command/service/memory.ex
```diff
@@ -0,0 +1,168 @@
...
+        :ets.match_delete(tid, {{token, :_}, :_})
+        :ets.delete(tid, monitor_ref)
+      [] ->
+        Logger.warn("Unknown monitor ref #{inspect monitor_ref} for pid #{inspect pid} going down for #{inspect reason}")
+    end
+
+    {:noreply, state}
+  end
+
+  defp monitor_executor(token, tid) do
```
christophermaier: I think we're going to stack up monitors with this code, one for each operation that gets called in the pipeline. We should probably just monitor the process if we're not already monitoring it.

vanstee: D'oh TIL:

```
Making several calls to monitor/2 for the same Item and/or Type is not an error; it results in as many independent monitoring instances.
```

Thanks @christophermaier. Fixin'

christophermaier:
👍

PR #591, opened 2016-04-28
Wait until DB responds after processing bundle announcement

kevsmith:
Prevents a non-obvious failure at startup if the database is slow to respond.


imbriaco:
:+1:

PR #595, opened 2016-04-28
Add /v1/users/me route

imbriaco:
This adds a `/v1/users/me` route to return a standard user resource for the user associated with the token that was used to make the API call. This is helpful since the user may not know their UUID in advance, so it's difficult to make use of the existing `/v1/users/:id` route.


kevsmith:
👍

PR #599, opened 2016-04-29
Allow users to update  their own data.

### LINKED ISSUE #596: Allow users to manage their own chat handles
Currently, `Cog.V1.ChatHandleController` requires the `operable:manage_users` permission. This controller should be updated much like `Cog.V1.UserController` was to allow users to create/update/delete chat handles that are associated with their own user ID.

kevsmith:
This commit moves the self updating check logic out of v1.UserController
and into Plug.Authorization where it can be more easily reused.

Fixes #596

### DIFF
--- a/web/plugs/authorization.ex
+++ b/web/plugs/authorization.ex
@@ -9,9 +9,32 @@ defmodule Cog.Plug.Authorization do
   alias Cog.Models.Permission
   alias Cog.Repo
 
+  @typedoc """
+  Namespaced permission name a user must have in order to be authorized
+  """
+  @type permission_opt :: {:permission, String.t}
+
+  @typedoc """
+  Boolean flag indicating whether or not 'self updates' are authorized
+  'Self update' is defined as then invocation of `:show` or `:update` action
+  when the path parameter `id` matches the currently authenticated user.
+  """
+  @type self_updates_opt :: {:allow_self_updates, bool}
+
+  @type auth_options :: [permission_opt | self_updates_opt]
+
   @doc """
-  Requires a `:permission` option, which should be the namespaced name
-  of a permission that a user must have in order to be authorized.
+  Options:
+    `:permission` - Fully namespaced permission name required for authorization
+    `:allow_self_updates` - Permit user to edit their own data. Defaults to false
+                            if unset.
+
+  Example:
+
+  ```
+  plug Cog.Plug.Authorization, [permission: "operable:manage_users",
+                                allow_self_updates: true]
+  ```
   """
   def init(opts) do
     permission = Keyword.fetch!(opts, :permission)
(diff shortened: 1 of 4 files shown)


web/plugs/authorization.ex
```diff
@@ -19,27 +19,51 @@ defmodule Cog.Plug.Authorization do
...
+
+  # NOTE: This code assumes the id path parameter is the
+  # the user's ID. We should consider making this configurable
+  # in the future. For example, this would allow users to update
+  # related data with paths such as `/v1/users/:user_id/profiles/:id`.
+  defp self_updating?(conn) do
+    conn.private.phoenix_action in [:update, :show] and
+      conn.assigns.user.id == conn.params["id"] or
+      conn.params["id"] == "me"
+  end
```
christophermaier: Some parentheses here would be good; hard to tell if it's `(X and Y) or Z`, `X and (Y or Z)`

web/plugs/authorization.ex
```diff
@@ -19,27 +19,51 @@ defmodule Cog.Plug.Authorization do
...
     permission = name_to_permission(permission_name)
     case User.has_permission(authenticated_user, permission) do
       true ->
         conn
       false ->
-        conn
-        |> put_status(:forbidden)
-        |> json(%{error: "Not authorized."})
-        |> halt
+        if Keyword.get(opts, :allow_self_updates, false) do
```
christophermaier: Adding a blurb about `:allow_self_updates` in the docstring for `init/1` would be helpful.

web/plugs/authorization.ex
```diff
@@ -19,27 +19,51 @@ defmodule Cog.Plug.Authorization do
...
+          if self_updating?(conn) do
+            conn
+          else
+            forbid_access(conn)
+          end
+        else
+          forbid_access(conn)
+        end
```
vanstee: Looks like you can collapse these clauses.

```elixir
authorized = User.has_permission(authenticated_user, permission) ||
               Keyword.get(opts, :allow_self_updates, false) &&
               self_updating?(conn)

if authorized do
  conn
else
  forbid_access(conn)
end
```

vanstee:
LGTM

christophermaier:
👍  modulo a few small comments

web/plugs/authorization.ex
```diff
@@ -9,31 +9,58 @@ defmodule Cog.Plug.Authorization do
...
     permission = name_to_permission(permission_name)
-    case User.has_permission(authenticated_user, permission) do
-      true ->
-        conn
-      false ->
-        conn
-        |> put_status(:forbidden)
-        |> json(%{error: "Not authorized."})
-        |> halt
+    if User.has_permission(authenticated_user, permission) || self_updating?(opts, conn) do
```
vanstee: :+1:

christophermaier:
👍

PR #600, opened 2016-04-29
Changed invalid config message to just "Invalid config."

mpeck:
The message about the config file didn't make sense anymore, since you
can either send a file or pass the config directly.

### DIFF
--- a/web/controllers/v1/bundles_controller.ex
+++ b/web/controllers/v1/bundles_controller.ex
@@ -139,7 +139,7 @@ defmodule Cog.V1.BundlesController do
   end
 
   defp error(:no_config) do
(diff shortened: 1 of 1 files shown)


web/controllers/v1/bundles_controller.ex
```diff
@@ -139,7 +139,7 @@ defmodule Cog.V1.BundlesController do
   end
 
   defp error(:no_config) do
-    {:bad_request, %{error: "'bundle_config' not specified. Make sure to use a multipart form and to send your file in the 'bundle_config' field."}}
+    {:bad_request, %{error: "Invalid config."}}
```
kevsmith: I think `"Missing bundle config."` would be a clearer error message given this function head is matching on `:no_config`.

kevsmith:
👍 Modulo above comment

PR #601, opened 2016-04-29
Deal with existing state when starting a memory service

vanstee:
* Remonitor or delete related data for existing processes on init
* Periodically check all keys and remove anything that could have snuck in between process restarts

### DIFF
--- a/lib/cog/command/service/memory.ex
+++ b/lib/cog/command/service/memory.ex
@@ -145,7 +145,7 @@ defmodule Cog.Command.Service.Memory do
   end
 
   defp schedule_dead_process_cleanup(interval) do
-    Logger.info ("Scheduling dead process cleanup for #{round(interval / 1000)} from now")
+    Logger.info ("Scheduling dead process cleanup for #{round(interval / 1000)} seconds from now")
     Process.send_after(self(), :dead_process_cleanup, interval)
   end
 
--- a/lib/cog/command/service/memory.ex
+++ b/lib/cog/command/service/memory.ex
@@ -110,12 +110,9 @@ defmodule Cog.Command.Service.Memory do
   end
 
   def handle_info({:DOWN, _monitor_ref, :process, pid, reason}, state) do
-    Logger.debug("Process #{inspect pid} went down (#{inspect reason}); removing its memory storage")
-
     case ets_lookup(state.monitor_table, pid) do
       {:ok, token} ->
-        :ets.match_delete(state.memory_table, {{token, :_}, :_})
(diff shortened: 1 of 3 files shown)


lib/cog/command/service/memory.ex
```diff
@@ -59,100 +61,139 @@ defmodule Cog.Command.Service.Memory do
...
+
+    {:noreply, state}
+  end
+
+  defp account_for_existing_processes(monitor_table, memory_table) do
+    ets_iterate(monitor_table, fn pid, token ->
+      case Process.alive?(pid) do
+        true ->
+          Logger.debug("Remonitoring #{inspect pid} for token #{inspect token}")
+          :erlang.monitor(:process, pid)
```
christophermaier: Don't you need to keep track of this monitor?

lib/cog/command/service/memory.ex
```diff
@@ -59,100 +61,136 @@ defmodule Cog.Command.Service.Memory do
...
+        false ->
+          cleanup_process(monitor_table, memory_table, pid, token)
+      end
+    end)
+  end
+
+  defp schedule_dead_process_cleanup(interval) do
+    Logger.info ("Scheduling dead process cleanup for #{round(interval / 1000)} from now")
```
christophermaier: Should add the units in the log message

lib/cog/command/service/memory.ex
vanstee: I don't think so. The pid is already in the ets table and we're calling this before init returns so all pids in the ets table should be monitored and we can use that to make sure we're not double monitoring.

christophermaier: Ah, yeah... I just got to that part below. 👍

test/cog/command/service/memory_test.exs
```diff
@@ -6,6 +6,30 @@ defmodule Cog.Command.Service.MemoryTest do
...
+  end
+
+  setup do
+    memory_table  = :ets.new(:test_memory_table,  [:public])
+    monitor_table = :ets.new(:test_monitor_table, [:public])
+
+    {:ok, pid} = Memory.start_link(memory_table, monitor_table)
+
+    on_exit(fn ->
+      Process.exit(pid, :kill)
```
christophermaier: Is this necessary if you've `start_linked` the process?

christophermaier:
👍

test/cog/command/service/memory_test.exs
vanstee: Ah you're right.

PR #602, opened 2016-05-02
Tweak transaction logic in tests

christophermaier:
Current Phoenix practice seems to be starting up a test transaction in
`test_helper.exs` and strategically rolling it back in specific test
cases as required.

The tests as they currently exist run fine, but in some I'm working on
in another branch, we can run into issues if we use our `ConnCase`
without also using our `ModelCase`, since the latter actually starts a
test transaction, but the former will roll one back.


mpeck:
Nice :+1:

PR #603, opened 2016-05-02
Service metadata API + Memory Service

### LINKED ISSUE #358: HTTP-based service endpoint for commands
# Cog Services API

Cog services will be managed by a separate Phoenix endpoint named `Cog.PlatformServicesEndpoint`. Building the API as a separate endpoint instead of a separate Phoenix app was chosen to avoid packaging and reuse issues with Cog (since it was never designed to be a reusable library) and also reduce the coding effort. An endpoint-based implementation strikes a good balance between isolation, modularity, and speed of development without constraining future choices.

The endpoint will contain its own HTTP router with the majority of application logic contained in controllers and views. The root URL of the services API will be exported to the command runtime environment under the environment variable `COG_SERVICES_ROOT`. It shall be a valid URL starting with either `http://` or `https://`, depending on user configuration, followed by the host name or IP address and listening port number. Examples: `http://192.168.2.22:4005`, `https://mycog.somecorp.com:4100`.

## Authentication and Authorization

Cog will generate a unique pipeline token upon pipeline initialization. The token will correlate a pipeline to the calling user much like REST API tokens correlate REST API sessions to calling users. Pipeline tokens have a finite lifespan. Their expiration can be triggered by either the completion (regardless of status) of the owning pipeline or after 1.5x the maximum execution interval for a pipeline. As of 4/18/2016, this translates to a maximum lifespan of 90 seconds.

The token will be exported to the command runtime environment under the environment variable `COG_PIPELINE_TOKEN`. All calls to the Cog Service API _must_ include the token in their `Authorization` HTTP header like so: `Authorization: pipeline <token>`. Requests without a properly formed `Authorization` header or referencing an expired or invalid pipeline token will receive `401 Unauthorized`.

## Routes

The Services API provides an informational API to assist monitoring, service discovery, and service enumeration. These endpoints are listed below.

### /services/meta

Returns a JSON object describing Cog version, Cog Services API version, and basic information about each hosted service. The below JSON document can be used as a normative reference.

#### Request
```http
GET /services/meta HTTP/1.1
Authorization: pipeline 123456789
```

#### Response
```http
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: ...

{"info": 
  {
    "cog_version": "0.5.0",
    "cog_services_api_version": "1",
    "services": [{"name": "exec",
                  "version": "1.0"},
                 {"name": "memory",
                  "version": "1.0"}]
  }
}
```

### /services/meta/deployed/{service_name}

Returns a Swagger 2.0 compliant JSON description of the named service's API. A 404 status code shall be returned if the requested service doesn't exist.

christophermaier:
Fixes #358

### DIFF
--- a/priv/swagger/services/memory.yaml
+++ b/priv/swagger/services/memory.yaml
@@ -1,0 +1,97 @@
+swagger: '2.0'
+
+info:
+  version: "1.0.0"
+  title: Cog Memory Service
+
+basePath: "/v1/services/memory/1.0.0"
+
+definitions:
+  memory_value:
+    description: |
+      A value to be stored in memory, added to memory, or retrieved
+      from memory.
+
+      Though this is formally declared as a `string` type, it can
+      actually be any valid JSON value (object, list, number, etc.)
+    type: string
+
+paths:
+  /{key}:
+    get:
+      description: |
+        Retrieve the current value stored under `key`
+      parameters:
+        - name: key
+          in: path
+          description: the memory key to retrieve
+          required: true
+          type: string
+      responses:
+        "200":
+          description: The current value stored under `key`
+          schema:
+            $ref: '#/definitions/memory_value'
+        "404":
+          description: The key was not found
+    put:
+      description: |
+        Set a value for `key`
+      parameters:
+        - name: key
+          in: path
+          description: the key to operate on
+          required: true
+          type: string
+        - name: data
+          in: body
+          description: Value to set for `key`
+          required: true
+          schema:
+            $ref: '#/definitions/memory_value'
+      responses:
+        "200":
+          description: The resulting value of `key`
+    post:
+      description: |
+        Perform a memory operation
+      parameters:
+        - name: key
+          in: path
+          description: the key to operate on
+          required: true
+          type: string
+        - name: operation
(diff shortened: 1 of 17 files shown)


priv/swagger/services/memory.yaml
```diff
@@ -0,0 +1,97 @@
+swagger: '2.0'
```
kevsmith: Have we tried generating a client from this doc to verify it's accurate?

web/controllers/v1/memory_service_controller.ex
```diff
@@ -0,0 +1,85 @@
...
+  use Cog.Web, :controller
+  require Logger
+
+  plug Cog.Plug.ServiceAuthentication
+  import Cog.Plug.Util, only: [get_service_token: 1]
+
+  alias Cog.Command.Service.Memory
+
+  def show(conn, %{"key" => key}) do
+    service_token = get_service_token(conn)
```
kevsmith: Feels like this check should be done in a plug for two reasons:

1. DRY up duplicate token checking logic
2. Return a more accurate status code (400/401/422) instead of 500 in the case of an invalid token.

PR #605, opened 2016-05-02
Better handle edge cases where services don't cleanup after pipeline processes

### LINKED ISSUE #594: Reap ets entries from services even after process crashes
Right now if the token or memory service crash, they'll no longer receive DOWN info calls which we rely on to clean up after pipeline executor processes that have finished or crashed. We need to do two things to fix this:

* [ ] Store the executor pids in the ets table instead of the monitor pid, and re-monitor existing pids on startup in the case of a service process crash
* [ ] Run a reaper that periodically checks all the pids we have stored in the ets table to make sure that they're still alive. If not, clean up after those entries since we missed them in between a process crash and a restart.

vanstee:
To better clean up dead pipeline processes we now do two things (implemented in the shared `PipelineMonitor` module).

1. On process init, look at all processes in the monitor table. Monitor for the alive pids and cleanup after the dead pids.
2. Periodically (every 30 seconds) look at all pids we know about and cleanup after any dead pids.

Closes #594

### DIFF
--- a/lib/cog/command/service/memory.ex
+++ b/lib/cog/command/service/memory.ex
@@ -18,6 +18,10 @@ defmodule Cog.Command.Service.Memory do
 
   defstruct [:memory_table, :monitor_table]
 
+  @doc """
+  Starts the #{inspect __MODULE__} service. Accepts two public ets table
+  ids: one for storing tokens and keeping track of monitored pids.
+  """
   def start_link(memory_table, monitor_table),
     do: GenServer.start_link(__MODULE__, [memory_table, monitor_table], name: __MODULE__)
 
--- a/lib/cog/command/service/memory.ex
+++ b/lib/cog/command/service/memory.ex
@@ -10,10 +10,11 @@ defmodule Cog.Command.Service.Memory do
   """
 
   use GenServer
-  alias Cog.Command.Service
+  import Cog.Command.Service.PipelineMonitor
+  alias Cog.ETS
   require Logger
 
-  @dead_process_cleanup_interval 30000 # 30 seconds
+  @dead_pipeline_cleanup_interval 30000 # 30 seconds
 
   defstruct [:memory_table, :monitor_table]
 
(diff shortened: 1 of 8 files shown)


lib/cog/command/service/memory.ex
```diff
@@ -65,54 +70,56 @@ defmodule Cog.Command.Service.Memory do
...
+    account_for_existing_pipelines(monitor_table, memory_table, &{&1, :_})
+    schedule_dead_pipeline_cleanup(@dead_pipeline_cleanup_interval)
 
     state = %__MODULE__{memory_table: memory_table, monitor_table: monitor_table}
     {:ok, state}
   end
 
   def handle_call({:fetch, token, key}, _from, state) do
-    result = ets_lookup(state.memory_table, {token, key})
+    result = ETS.lookup(state.memory_table, {token, key})
```
kevsmith: Can we change the module name to `EtsWrapper` or something? I think `ETS` will be too confusing over time.

vanstee: Yep. Good idea.

lib/cog/ets.ex
```diff
@@ -0,0 +1,52 @@
...
+    :ets.match_delete(table, query)
+  end
+
+  def each(table, fun) do
+    :ets.safe_fixtable(table, true)
+    unsafe_each(table, fun)
+    :ets.safe_fixtable(table, false)
+  end
+
+  def unsafe_each(table, fun) do
```
kevsmith: Where do we need iterate over an unfixed table?

lib/cog/ets.ex
```diff
@@ -0,0 +1,52 @@
...
+    do_each(table, :ets.first(table), fun)
+  end
+
+  defp do_each(_table, :'$end_of_table', _fun),
+    do: :ok
+  defp do_each(table, key, fun) do
+    case lookup(table, key) do
+      {:ok, value} ->
+        fun.(key, value)
+      _error ->
```
kevsmith: Should probably log this error since it aborts iteration.

lib/cog/ets.ex
vanstee: Ah we're not anymore. Removing.

kevsmith:
Are tests for `Cog.Memory.Service` and `Cog.Service.Command.PipelineMonitor` part of another commit? Want to make sure we're testing for races and edge conditions.

vanstee:
@kevsmith Yep those already exist as this work was mostly pulling that work out into a reusable module (`PipelineMonitor`).

https://github.com/operable/cog/blob/master/test/cog/command/service/memory_test.exs#L95-L155

Issue #604, opened 2016-05-02
Upsert chat handles


christophermaier:
Currently, it's only really possible for a user to have one active chat handle at a time, governed by the chat adapter that Cog is currently configured to use.

In order to determine whether or not a user is updating or creating a chat handle, one approach is to add a `:show` action to the chat handles controller (currently, the `:index` action is used, which is not ideal).

However, since there can really only be one chat handle active at a time, we can simplify things by merging the current `:create` and `:update` actions into on (effectively) "upsert" operation.

To prevent confusion, we'll require the chat provider name in the API call, and return an error if that name doesn't correspond to the currently-configured chat adapter (until we can relax the "internal ID lookup logic" when chat handles are created).

Also, the database schema currently allows a single Cog user to have multiple chat handles from the same provider. That's not intentional, and can also be fixed in the course of this work.

imbriaco:
:+1:

jsteiner:
👍  Sounds good to me

PR #606, opened 2016-05-02
A user should have one chat handle per provider

christophermaier:
Previously, a Cog user could actually have an unlimited number of chat
handles for a given provider (e.g., 5 different Slack handles mapping to
a single Cog user). This was an oversight that is now corrected.

There is already a unique constraint on `{provider_id, handle}`. By
replacing the constraint on `{user_id, provider_id, handle}` with the
simpler `{user_id, provider_id}`, we get the effect we want.

The controller tests were implicitly depending on the ability to
assign multiple handles to each user. They have been minimally
refactored to accommodate this index change, under the assumption that
they'll change more extensively with work for #604.


kevsmith:
LGTM

PR #609, opened 2016-05-03
Chat handle creation is now an "upsert"

### LINKED ISSUE #604: Upsert chat handles

christophermaier:
Previously it was a bit awkward to know whether or not you needed to
create or update a chat handle for a user. It turns out we can simplify
things, because a user can only ever have a single chat handle for a
given chat provider. Whether the user already has a chat handle or not,
we can just say we're "setting" the chat handle and basically do an
"upsert" operation on the backend.

This removes the `PUT` operation on the chat handle controller. Now, to
both create and update a user's chat handle, send a `POST` to
`/v1/users/:id/chat_handles`.

This also puts a guard in against the situation where a user tries to
set a chat handle for a provider that differs from the
currently-configured Cog chat adapter. Previously, this would crash Cog;
now it's just an error tuple.

This also pulls this chat handle logic out of the controller and into a
chat handle "repository", in keeping with the new work to consolidate
domain functionality in the core of the system.

Fixes #604

### DIFF
--- a/lib/cog/repository/chat_handles.ex
+++ b/lib/cog/repository/chat_handles.ex
@@ -20,7 +20,7 @@ defmodule Cog.Repository.ChatHandles do
           result = user_id
           |> find_handle_for(provider_name)
           |> ChatHandle.changeset(params)
-          |> upsert
+          |> Repo.insert_or_update
 
           case result do
             {:ok, chat_handle} ->
@@ -54,17 +54,6 @@ defmodule Cog.Repository.ChatHandles do
     join: p in assoc(ch, :chat_provider),
     where: ch.user_id == ^user_id,
     where: p.name == ^provider_name
-  end
-
-  # TODO: When we can do database upserts, let's actually do a real
-  # one here
-  defp upsert(changeset) do
-    if changeset.model.id == nil do
-      # NOTE: for Ecto 2.0, it will be `data`, not `model`
-      Repo.insert(changeset)
-    else
-      Repo.update(changeset)
-    end
   end
 
   # We track the chat provider's "internal user ID" for a given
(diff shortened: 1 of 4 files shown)


jsteiner:
👍  good stuff

kevsmith:
LGTM

lib/cog/repository/chat_handles.ex
```diff
@@ -0,0 +1,101 @@
...
+  # consult the chat provider to obtain this information.
+  #
+  # One side effect of how things are currently structured is that we
+  # only allow handles to be created or edited if the currently
+  # running chat adapter is the one for the chat provider in
+  # question. That is, if you're running Cog with the Slack adapter,
+  # you can _only_ create or update Slack chat handles.
+  defp chat_handle_params(provider_name, handle, user_id) do
+    {:ok, adapter} = Cog.chat_adapter_module
+    if String.downcase(provider_name) == adapter.name do
```
vanstee: This is super minor, but it's kinda a bummer that we have to `downcase` the provider_name twice. Might be worth throwing it in a variable real quick.

christophermaier: 😫  Aargh... one sec.

vanstee:
Looks good. Curious about Ecto's upsert support but this should be fine.

PR #610, opened 2016-05-03
Make dead token clean up less noisy

kevsmith:
I think this still logs all relevant info without making Cog logs too noisy. Since we log whenever a stale token is cleaned up we shouldn't lose any visibility.


vanstee:
:+1: Noticed this as well. Thanks for cleaning it up.

Issue #611, opened 2016-05-03
Render tables in templates


mpeck:
Often times you want the data returned by commands to be presented as a table. For example, when you request a list of things. Currently there isn't an easy way to generate a table without piping the output of the command into the table function. And commands can't easily render tables on there own without polluting their data with additional rendering info. It would be nice if we could just generate a table in the commands template.

Issue #513, opened 2016-04-12
Add administration commands for relays and relay groups


imbriaco:
The following administration activities for relays and relay_groups should be supported in the commands within the `operable` bundle. The the maximum extent possible, these should mirror the same actions in `cogctl`:

* List relays grouped by relay_group.
* Update name and/or description for a relay.
* Create/Rename/Delete relay_groups.
* Add/Remove relays to relay_groups.
* Assign/Unassign bundles to relay_groups.

Note: Relay creation and token update should not be exposed in the chat commands for now so that we don't leak the token.

mpeck:
# Commands
## relays
```
Lists relays.
Usage: relays [--by-relay-group]

Examples:
me> relays
cog>  NAME     STATUS    CREATED AT             ID
      relay1   enabled   2016-04-27T13:23:33Z   e7ddd6e9-9ad1-4a13-83a2-2c993478654e
      relay2   enabled   2016-04-27T13:23:33Z   0e2a1a50-efea-487c-b1a9-bd2cb62d8df3
      relay3   disabled  2016-04-27T13:23:33Z   0872b713-9eef-4559-9ae2-4af55d2690ba
  
me> relays --by-relay-group
cog>  NAME     RELAY GROUP   STATUS    CREATED AT             ID
      relay1   group1        enabled   2016-04-27T13:23:33Z   e7ddd6e9-9ad1-4a13-83a2-2c993478654e
      relay3   group1,group2 enabled   2016-04-27T13:23:33Z   0872b713-9eef-4559-9ae2-4af55d2690ba
      relay2   group2        disabled  2016-04-27T13:23:33Z   0e2a1a50-efea-487c-b1a9-bd2cb62d8df3
```

JSON output
```
[{"name": "relay1",
  "status": "enabled",
  "created_at": "2016-04-27T13:23:33Z",
  "id": "e7ddd6e9-9ad1-4a13-83a2-2c993478654e",
  "description": "",
  "relay_groups": ["group1"], // only available if --by-relay-group is passed},
  ...]
```
As far as I can tell there is no way to list relays grouped by relay group in cogctl. Perhaps that's something we should add?

## relays update
```
Updates a relay.
Usage: relays update <relay_name> [--name][--description]

Examples:
me> relays update relay1 --name bigdaddy1
cog>  ok
```
JSON Output
```
{"status": "ok",
 "relay": { ... relay model ... }}
```
[...]

PR #613, opened 2016-05-03
Relay commands

mpeck:
Adds support for listing relays and updating their name and/or description.

Would really appreciate thoughts on the help bits. I rolled usage info into the command in order to get more control.

reference #513 

Note: I couldn't get the display output exactly like I wanted it. I think we will need a few updates to FuMachu before we can easily render tables without piping into the table command. Created a ticket for that here, #611.


kevsmith:
Couple of UX issues I noticed using the relay command:

- `relay list` when no relays are installed returns "Pipeline executed successfully but no output was returned". The message should be "No relays configured".
- `relay update` should allow the user to update the relay's token too.

kevsmith:
How will the user get help for the subcommands?

lib/cog/commands/relay/list.ex
```diff
@@ -0,0 +1,72 @@
...
+      if Helpers.flag?(options, "group") do
+        Map.put(relay_map, "relay_groups", Enum.map(relay.groups, &generate_group_map/1))
+        |> Map.put("_show_groups", true)
+      else
+        relay_map
+      end
+    end)
+  end
+
+  defp verbose_relay(relay) do
```
kevsmith: I'd unify `verbose_relay/1` and `standard_relay/1` into a single function that returned all the data. You can then handle what the user sees in the output template without impacting usefulness within a pipeline.

mpeck:
@kevsmith the token update bit was specifically called out to not expose in the original issue. Is that no longer the case?

mpeck:
And also, I'm glad you pointed out the "Pipeline executed successfully but no output was returned" thing. We get that from the executor. If you return an empty list the executor bypasses templating and renders that message. Is that desired behavior? It does kinda suck since you can't handle 'no results' cases in your template.

lib/cog/commands/relay/list.ex
mpeck: So I actually thought about that @kevsmith, and originally that's how I planned on doing it. But there isn't a way to easily get additional metadata into the template without polluting the record. I could just stick flags in each object to turn things on and off, that's actually what I had to do with the relay groups bit to get it to display properly, but that seems like a generally bad idea. Do you have any thoughts?

kevsmith:
@mpeck Thanks for pointing out the token bit in the original issue. I _assumed_ we'd want parity with `cogctl` but it makes total sense we wouldn't.

kevsmith:
Re: no output -- possibly we could handle with a template by emitting `[{relays: []}]` and specifying an empty result template?

PR #615, opened 2016-05-04
Relay group commands

mpeck:
Implements commands to manage relay groups. The relay-group commands use a few bits from relay commands, for generating relay json and whatnot, which necessitated a couple tweaks to relays. I used a basically the same pattern in both the relay-group and relay commands. If it proves to be a decent pattern, I think we can potentially consolidate some of these bits in gen_command. But that's for another PR.

I also implemented a repository for relay groups. There is a fair bit of logic required for adding/removing/assigning/unassigning bundles and relays to relay groups. Instead of just copying it to the relay-group commands I created a repository. We will still need to update the relay-group and relay-group membership controllers to leverage the repository though.

resolves #613


kevsmith:
Based on a quick read this looks OK to me.

Issue #621, opened 2016-05-04
Remove enforcing commands


mpeck:
unenforcing commands are no longer a thing. All commands will now require rules in order to invoke. We just need to remove all references to unenforcing commands and add the new "allow" rule. For now we can leave the 'enforcing' field in the database. We may still use it internally for optimizations, but there should be no user facing mentions of enforcing commands.

parent #399
depends on #620

mpeck:
- [ ] update commands with the new "allow" syntax.
- [ ] remove enforcing from gen_command

unenforcing commands: alias, echo, filter, greet, help, max, min, raw, seed, sleep, sort, sum, table, thorn, unique, wc, which

PR #645, opened 2016-05-05
change 'enforcing: false' commands to 'allowed' commands

mpeck:
The removes references to enforcing commands and adds the rule `when command is operable:<command> allow` for all our internal commands.

reference #621


kevsmith:
LGTM

PR #648, opened 2016-05-05
Nest chat_handles under users

jsteiner:
We needed a way for non-admin users to view their own chat handles. We
could have opened up chat_handle#index to the current_user, but I think
this is a more desirable option since it's a bit easier from an
implementation side in Cog, and it prevents having to make a second
request just to get the associated chat handle.

### DIFF
--- a/lib/cog/queries/chat_handle.ex
+++ b/lib/cog/queries/chat_handle.ex
@@ -1,8 +1,0 @@
-defmodule Cog.Queries.ChatHandle do
-  use Cog.Queries
-
-  def for_user_id(query \\ ChatHandle, user_id) do
-    from ch in query,
-    where: ch.user_id == ^user_id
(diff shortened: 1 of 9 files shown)


jsteiner:
Should I delete https://github.com/operable/cog/blob/master/web/controllers/v1/chat_handle_controller.ex#L15-L21 as well? This essentially replaces the need for that, but I wasn't sure if it needs to stick around for any reason.

imbriaco:
@jsteiner I think we can lose the function for listing a user chat handle in the chat handle controller. I don't see why we'd need that for anything else.

jsteiner:
@imbriaco updated

imbriaco:
Note: A Contributor License Agreement was not required for this PR because this work was done under contract.

PR #652, opened 2016-05-05
Rename stage_pos to invocation_step

vanstee:


mpeck:
:+1: I assume that this means the env var for foreign commands will be `COG_INVOCATION_STEP`?

PR #653, opened 2016-05-05
remove enforcing from gen_command and install

mpeck:
reference #621


vanstee:
Do we need to add that allow rule for all the embedded bundle commands?

mpeck:
That's already merged @vanstee https://github.com/operable/cog/pull/645

vanstee:
Gotcha. LGTM then.

PR #654, opened 2016-05-05
Include services_root in command request

vanstee:


lib/cog/command/pipeline/executor.ex
```diff
@@ -660,4 +662,8 @@ defmodule Cog.Command.Pipeline.Executor do
       end
     end
   end
+
+  defp services_root do
+    ServiceEndpoint.url <> "/v1/services"
```
mpeck: Instead of pulling the root out of the request, shouldn't the commands specify which endpoint to use? At least in regards to the version. Otherwise wouldn't all the commands break when we add '/v2/services`? I assume at some point we could have commands using different versions of the services endpoint.

vanstee: Hrm, that's a good question. The weirder part of this is that there's an api version for the services endpoint, but then also a version for each service. I guess both of those might need to be changed. I'll remove the services bit and allow each command to provide the version.

I wish we were including the version as a param in the Accepts header to avoid stuff like this. But for now I'll just remove it.

PR #659, opened 2016-05-06
Switch once commands to be multi commands

### LINKED ISSUE #625: Rework internal 'once' commands
All commands will become multi commands. That is, they will be run once for each result from the previous stage of the pipeline. Commands will depend on the new memory service to accumulate any state.

Commands to be reworked:
 - [ ] table
 - [ ] sort
 - [ ] unique
 - [ ] sleep

parent #399 
depends on:
* #618
* #619

vanstee:
All commands now use `invocation_step` and the memory service so they can be run as multi commands. Next step will be to remove the `execution` field on commands altogether.

Closes #625


kevsmith:
LGTM

Issue #623, opened 2016-05-04
bundle config cleanup


mpeck:
With the simplifications to command calling conventions we need to update bundle configs.

* remove execution field, all commands will now be multi commands
* remove enforcing field, all commands will now be enforcing
* require rules, since commands can no longer be unenforcing at least one rule is required.

parent #399

PR #661, opened 2016-05-06
update bundles controller for config validator changes

mpeck:
updates bundles controller to handle the new response signature of `Spanner.Config.validate`. Now if warnings are generated they are passed back to the user.

depends on https://github.com/operable/spanner/pull/70
reference https://github.com/operable/cog/issues/623

### DIFF
--- a/test/controllers/v1/bundles_controller_test.exs
+++ b/test/controllers/v1/bundles_controller_test.exs
@@ -56,6 +56,26 @@ defmodule Cog.V1.BundlesControllerTest do
     end
   end)
 
+  test "accepts an upgradable config", %{authed: requestor} do
+    config_path = make_config_file("old_config.yaml", contents: old_config)
+    upload = %Plug.Upload{path: config_path, filename: "old_config.yaml"}
+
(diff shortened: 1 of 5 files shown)


web/controllers/v1/bundles_controller.ex
```diff
@@ -41,12 +41,17 @@ defmodule Cog.V1.BundlesController do
 
   def create(conn, %{"bundle" => params}) do
     case install_bundle(params) do
-      {:ok, bundle} ->
+      {:ok, bundle, warnings} ->
+        view = if length(warnings) > 0 do
+          "show_with_warnings.json"
+        else
+          "show.json"
+        end
```
vanstee: Could you do this inside the view instead and conditionally include a `warnings` key in the result?

test/controllers/v1/bundles_controller_test.exs
```diff
@@ -202,7 +202,7 @@ defmodule Cog.V1.BundlesControllerTest do
     """
     ---
     # Format version
-    cog_bundle_version: 2
+    cog_bundle_version: 3
```
kevsmith: Do we have tests for upgrading v2 configs via the controller?

Issue #612, opened 2016-05-03
Ensure embedded commands are compatible with Simplified Calling Convention


kevsmith:
Depends on #399

mpeck:
This feels like a good opportunity to verify that our usage info is at least roughly formatted the same, and that we have a few tests for each command.

I think the best path is to PR this in batches:
- [ ] batch1 - alias, bundle, echo, filter
- [ ] batch2 - group, help, max, min, permissions
- [ ] batch3 - raw, relay, relay_group, role
- [ ] batch4 - rules, seed, sleep, sort
- [ ] batch5 - sum, table, unique, wc, which

Batch grouping are subject to change depending on the number of changes. I'll update the issue accordingly.

PR #664, opened 2016-05-10
updates to usage info and tests for alias, bundle, echo, filter

mpeck:
This standardizes usage info and adds tests when necessary for alias, bundle, echo, filter.

reference #612

### DIFF
--- a/lib/cog/commands/alias.ex
+++ b/lib/cog/commands/alias.ex
@@ -6,18 +6,18 @@ defmodule Cog.Commands.Alias do
   @moduledoc """
   Manages aliases
 
-  USAGE:
+  USAGE
     alias <subcommand>
 
 
-  SUBCOMMANDS:
+  SUBCOMMANDS
     new <alias-name> <alias-definition>             creates a new alias visible to the creating user.
     mv <alias-name> <site | user>:[new-alias-name]  moves aliases between user and site visibility. Optionally renames aliases.
     rm <alias-name>                                 Removes aliases
     ls [pattern]                                    Returns the list of aliases optionally filtered by pattern. Pattern support basic wildcards with '*'.
 
 
-  EXAMPLES:
+  EXAMPLES
     alias new my-awesome-alias "echo \"My Awesome Alias\""
     > user:my-awesome-alias has been created
 
--- a/lib/cog/commands/alias.ex
+++ b/lib/cog/commands/alias.ex
@@ -6,31 +6,28 @@ defmodule Cog.Commands.Alias do
   @moduledoc """
   Manages aliases
 
-  ## Usage
-
+  USAGE:
     alias <subcommand>
 
 
-  ## Subcommands
+  SUBCOMMANDS:
+    new <alias-name> <alias-definition>             creates a new alias visible to the creating user.
+    mv <alias-name> <site | user>:[new-alias-name]  moves aliases between user and site visibility. Optionally renames aliases.
+    rm <alias-name>                                 Removes aliases
+    ls [pattern]                                    Returns the list of aliases optionally filtered by pattern. Pattern support basic wildcards with '*'.
 
-    • new <alias-name> <alias-definition> -- creates a new alias visible to the creating user.
-    • mv <alias-name> <site | user>:[new-alias-name] -- moves aliases between user and site visibility. Optionally renames aliases.
-    • rm <alias-name> -- Removes aliases
-    • ls [pattern] -- Returns the list of aliases optionally filtered by pattern. Pattern support basic wildcards with '*'.
 
-
-  ## Example
-
-    @cog alias new my-awesome-alias "echo \"My Awesome Alias\""
+  EXAMPLES:
+    alias new my-awesome-alias "echo \"My Awesome Alias\""
     > user:my-awesome-alias has been created
 
-    @cog alias mv my-awesome-alias site:awesome-alias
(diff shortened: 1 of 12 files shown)


lib/cog/commands/alias.ex
```diff
@@ -6,31 +6,38 @@ defmodule Cog.Commands.Alias do
...
   ## Subcommands
-  * new <alias-name> <alias-definition> -- creates a new alias visible to the creating user.
-  * mv <alias-name> <site | user>:[new-alias-name] -- moves aliases between user and site visibility. Optionally renames aliases.
-  * rm <alias-name> -- Removes aliases
-  * ls [pattern] -- Returns the list of aliases optionally filtered by pattern. Pattern support basic wildcards with '*'.
+
+    • new <alias-name> <alias-definition> -- creates a new alias visible to the creating user.
+    • mv <alias-name> <site | user>:[new-alias-name] -- moves aliases between user and site visibility. Optionally renames aliases.
+    • rm <alias-name> -- Removes aliases
+    • ls [pattern] -- Returns the list of aliases optionally filtered by pattern. Pattern support basic wildcards with '*'.
```
vanstee: I think we could probably just get rid of the dots here. Although, they do look pretty fancy.

lib/cog/commands/alias.ex
```diff
@@ -6,31 +6,35 @@ defmodule Cog.Commands.Alias do
   @moduledoc """
   Manages aliases
 
-  ## Subcommands
-  * new <alias-name> <alias-definition> -- creates a new alias visible to the creating user.
-  * mv <alias-name> <site | user>:[new-alias-name] -- moves aliases between user and site visibility. Optionally renames aliases.
-  * rm <alias-name> -- Removes aliases
-  * ls [pattern] -- Returns the list of aliases optionally filtered by pattern. Pattern support basic wildcards with '*'.
+  USAGE:
```
vanstee: Super nit-pick mode. Man pages seem to leave out the `:`. Any thoughts?

lib/cog/commands/alias.ex
```diff
@@ -6,31 +6,35 @@ defmodule Cog.Commands.Alias do
...
+    alias <subcommand>
 
-  ## Example
 
-      @bot #{Cog.embedded_bundle}:alias new my-awesome-alias "echo \"My Awesome Alias\""
-      > user:my-awesome-alias has been created
+  SUBCOMMANDS:
+    new <alias-name> <alias-definition>             creates a new alias visible to the creating user.
+    mv <alias-name> <site | user>:[new-alias-name]  moves aliases between user and site visibility. Optionally renames aliases.
```
vanstee: Does the colon need to be included in the second optional piece there?

lib/cog/commands/alias.ex
mpeck: I'm indifferent honestly. As long as we are consistent. I'll remove it.

lib/cog/commands/alias/new.ex
```diff
@@ -6,11 +6,12 @@ defmodule Cog.Commands.Alias.New do
...
-  usage:
-  alias new <alias-name> <alias-definition>
+  USAGE:
+    alias new <alias-name> <alias-definition>
 
-  example:
-  alias new my-awesome-alias "echo \"My awesome alias\""
+  EXAMPLE:
+    alias new my-awesome-alias "echo \"My awesome alias\""
+    > Success, 'user:my-awesome-alias' has been created"
```
vanstee: Looks like an extra quote on the end there.

lib/cog/commands/bundle.ex
```diff
@@ -29,46 +26,49 @@ defmodule Cog.Commands.Bundle do
...
   error.
 
-  ## Subcommands
+  USAGE:
+    bundle <subcommand>
+
 
-  * `status`
+  SUBCOMMANDS:
+    `status`
```
vanstee: You can probably remove the "`" since this will all be inside of a "```" anyway.

lib/cog/commands/echo.ex
```diff
@@ -4,11 +4,12 @@ defmodule Cog.Commands.Echo do
   @moduledoc """
   Repeats whatever it is passed.
 
-  ## Example
-
-      @bot #{Cog.embedded_bundle}:echo this is nifty
-      > this if nifty
+  USAGE:
+    echo [args]
```
vanstee: I like using that `...` to differentiate between 1 arg and multiple. It's pretty obvious here but, it's still nice to be consistent.

lib/cog/commands/alias.ex
mpeck: I'm trying to not change the content of the usage info right now. I'm just reformatting. Honestly I think all of this should be removed from here and moved into the subcommand module.

But to answer your question, yes. `mv` moves the alias to either the `site` or `user` visibility and can optionally change the name of the alias. So if you are moving visibility and changing the name it would require the colon. But I agree, I don't think it's particularly clear.

mpeck:
I think I've covered everything here @vanstee. Any comments before I merge?

vanstee:
:+1: We might want to do a one-over for all of the status messages we return in a future PR. I tend to prefer short messages like `Created alias user:jimmy-johns` omitting any `Successfully` or any unnecessary quotes. But that's not necessary for this PR.

mpeck:
Yeah, I'm just trying to get things tested and our usage info standardized for this.

PR #666, opened 2016-05-10
Usage updates and tests for group, help, max, min, permissions

mpeck:
Updates usage to roughly follow the same pattern and adds tests where necessary.

reference #612

### DIFF
--- a/lib/cog/commands/group.ex
+++ b/lib/cog/commands/group.ex
@@ -2,22 +2,20 @@ defmodule Cog.Commands.Group do
   @moduledoc """
   This command allows the user to manage groups of users.
 
-  ## Usage
+  USAGE:
+    group --create <groupname>
+    group --drop <groupname>
+    group --add --user=<username> <groupname>
+    group --remove --user=<username> <groupname>
+    group --list
 
-      group --create <groupname>
-      group --drop <groupname>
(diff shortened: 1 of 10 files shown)


lib/cog/commands/group.ex
```diff
@@ -2,22 +2,33 @@ defmodule Cog.Commands.Group do
...
-
-  ## Example
-
-      @bot operable:group --create ops
-      @bot operable:group --create engineering
-      @bot operable:group --add --user=bob ops
-      @bot operable:group --remove --user=bob ops
-      @bot operable:group --drop ops
-      @bot operable:group --list
+  USAGE:
```
vanstee: Super nitpick mode. I think the others exclude the `:`. Might be nice to stay consistent.

lib/cog/commands/group.ex
```diff
@@ -2,22 +2,33 @@ defmodule Cog.Commands.Group do
...
+    group --create <groupname>
+    group --drop <groupname>
+    group --add --user=<username> <groupname>
+    group --remove --user=<username> <groupname>
+    group --list
+
+  EXAMPLE:
```
vanstee: Also nitpicky `s/EXAMPLE/EXAMPLES`

PR #667, opened 2016-05-10
Usage and tests for raw, relay, relay_group, role

mpeck:
Updates usage and adds test where necessary for raw, relay, relay_group and role.

reference #612

### DIFF
--- a/lib/cog/commands/relay_group.ex
+++ b/lib/cog/commands/relay_group.ex
@@ -5,17 +5,19 @@ defmodule Cog.Commands.RelayGroup do
 
   @moduledoc """
   Manages relay groups
-  Usage relay-group <subcommand> [flags]
 
-  Subcommands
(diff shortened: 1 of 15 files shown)


lib/cog/commands/relay_group.ex
```diff
@@ -5,17 +5,19 @@ defmodule Cog.Commands.RelayGroup do
...
+
+  SUBCOMMANDS
+    info      Get info about one or more relay groups
+    create    Creates a relay group
+    rename    Renames a relay group
+    delete    Deletes a relay group
+    add       Adds relays to relay groups
+    remove    Remove relays from relay groups
+    assign    Assigns bundles to relay groups
+    unassign  Un-assigns bundles from relay groups
```
vanstee: Nice, this looks really good.

PR #668, opened 2016-05-10
Partially revert "removes the thorn and greet commands"

christophermaier:
This partially reverts commit 142f79e6b3478f007f4919ef77c87b0b4f2f6554.

Until bundle upgrades are in place, removing these commands actually
prevents the embedded bundle from starting, which blocks the entire
system from coming up.


imbriaco:
LGTM to merge once tests finish.

PR #669, opened 2016-05-10
Usage updates and tests for rules, seed, sleep, sort

mpeck:
Some usage and test updates for rules, seed, sleep and sort.

Note: I added a test for sleep, but I'm sure how useful it is.

reference #612


mpeck:
These changes were included in another PR #670

PR #670, opened 2016-05-10
updates to usage and tests for sum, table, unique, wc, which

mpeck:
Updates usage and adds tests when necessary for sum, table, unique, wc and which.

reference #612


vanstee:
LGTM. Thanks for this @mpeck

PR #683, opened 2016-05-11
Rework min and max

### LINKED ISSUE #673: Redesign `max` command
`max` isn't currently very useful because it only allows you to find the max among a list of args. Instead we should be finding the max given the items in the input list. We should also allow variable substitution to be used for finding the max via a single arg provided to each invocation of `max`.

```
USAGE

max [$variable]

EXAMPLES

seed '[{"a": 2}, {"a": 1}, {"a": 3}]' | max
> {"a": 3}

seed '[{"a": 2}, {"a": 1}, {"a": 3}]' | max $a
> 3

seed '[{"a": {"b": 2}}, {"a": {"b": 1}}, {"a": {"b": 3}}]' | max $a.b
> 3

### LINKED ISSUE #674: Redesign `min` command
Similar to `max`, `min` can currently only be used to find the minimum among a list of args. We should instead find the min in the input list and allow variable substitution.

```
USAGE

min [$variable]

EXAMPLES

seed '[{"a": 2}, {"a": 1}, {"a": 3}]' | min
> {"a": 1}

seed '[{"a": 2}, {"a": 1}, {"a": 3}]' | min $a
> 1

seed '[{"a": {"b": 2}}, {"a": {"b": 1}}, {"a": {"b": 3}}]' | min $a.b
> 1
```

### PR DESCRIPTION
**Pull request by vanstee**
Changed the args a little bit after talking to @mpeck since you'll probably always want to select an entire input item and not just a value.

Closes #673 
Closes #674

### DIFF
--- a/test/integration/commands/max_test.exs
+++ b/test/integration/commands/max_test.exs
@@ -1,34 +1,25 @@
-defmodule Integration.Commands.MaxTest do
+defmodule Integration.MaxTest do
   use Cog.AdapterCase, adapter: "test"
 
   setup do
-    user = user("vanstee", first_name: "Patrick", last_name: "Van Stee")
+    user = user("jfrost", first_name: "Jim", last_name: "Frost")
     |> with_chat_handle_for("test")
 
     {:ok, %{user: user}}
   end
 
-  test "Max from a set of integers", %{user: user} do
-    response = send_message(user, "@bot: operable:max 49 9 2 2")
-
-    [payload] = decode_payload(response)
-
-    assert payload.max == 49
+  test "basic max", %{user: user} do
+    response = send_message(user, ~s(@bot: seed '[{"a": 1}, {"a": 3}, {"a": 2}]' | max))
+    assert_payload(response, %{a: 3})
   end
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/integration/commands/max_test.exs [hunk 1]
```diff
@@ -1,34 +1,25 @@
...
 
-  test "Max from a set of words", %{user: user} do
-    response = send_message(user, "@bot: operable:max apple ball car zebra")
-
-    [payload] = decode_payload(response)
-
-    assert payload.max == "zebra"
+  test "max by complex key", %{user: user} do
+    response = send_message(user, ~s(@bot: seed '[{"a": {"b": 1}}, {"a": {"b": 3}}, {"a": {"b": 2}}]' | max a.b))
+    assert_payload(response, %{a: %{b: 3}})
```
mpeck: What happens when you use a key that does not exist? Maybe worth a test?

mpeck:
:+1: Looks good, just the one comment.

test/integration/commands/max_test.exs
vanstee: Good call. I'll fix that up.

PR #684, opened 2016-05-11
Tweaks to relay-group command

### LINKED ISSUE #687: relay-group update
the relay group command is mostly how we want it. We just need to put member actions, adding and removing bundles and relays, under their own sub-subcommand. The usage info needs to be tweaked a bit as well.

mpeck:
This updates relay-group USAGE info and sticks member functions, ie adding and removing bundles and relays, under their own subcommand.

resolves #687


imbriaco:
As discussed in Slack, I think we want to go with a calling style that puts all of the subcommands at the beginning of the command line for things like managing group members. For example:

```
relay-group member add <group-name> <relay1> .. <relayN>
relay-group member remove <group-name> <relay1> .. <relayN>
```

The primary motivator is to allow us to standardize on a style of a subcommand such that you could write commands with names like `relay-group-member-add` and have Cog infer subcommands from the name. That's a future optimization, but we may as well design in that direction in the mean time.

It's entirely possible for commands to adopt their own subcommand style based on parsing positional arguments themselves but our opinionated view (and the commands we write) should structure calling convention to have subcommands to the left and arguments to the right for consistency.

mpeck:
@imbriaco I should have been clearer. This follows the `relay-group member add <group name> <relay1> .. <relayN>` style. I switched it this morning.

mpeck:
relay-group
relay-group-member
relay-group-member-add

imbriaco:
Awesome, looks great. Sorry, I didn't see the commit from this morning in the history.

Issue #671, opened 2016-05-11
Redesign `group` command


vanstee:
Group was previously built to require long flags for options. Let's convert these flags to subcommands and bias towards args instead of using options. Also we'll restructure the output to be more easily used in a pipeline by always returning either an empty list, a list of groups, or a status message about the change made.

```
USAGE

group <subcommand>

group create <group-name>
group drop <group-name>
group add <group-name> <username>
group remove <group-name> <username>
group list

EXAMPLES

group create ops
> Created group ops

group drop ops
> Dropped group ops

group add ops vanstee
> Added user vanstee to group ops

group remove ops vanstee
> Removed user vanstee from group ops

group list
> NAME  USERS
  ops   vanstee imbriaco kevsmith chris mpeck
  ux    shelton
```

christophermaier:
Looks sooooooo much nicer 👍

imbriaco:
There are two different usage patterns that we're trying to collapse into this command. Managing existence of a group (create, delete) and managing the details of that group (show, add/remove users). 

I wonder if there might be a better model by thinking about this a little differently. Would love thoughts on the following possible UI:

**Group Existence:**
```
group list
group create <group-name>
group delete <group-name>
```

**Individual Group Management:**
```
group <group-name> add <user1> .. <userN> 
group <group-name> remove <user> .. <userN>
```

kevsmith:
I like @imbriaco's syntax ideas. I wonder if regularizing the syntax makes it easier to implement and document?

```
group list
group <group-name> create
group <group-name> delete
group <group-name> add <user1> .. <userN>
group <group-name> remove <user1> .. <userN>
```

imbriaco:
After discussing in the office, @kevsmith and I are both liking: 

```
group members <group-name> add <user1> .. <userN>
group members <group-name> remove <user1> .. <userN>
```

It keeps the `command subcommand target` flow, though admittedly there's a second subcommand tacked on after the target. I really like the way it reads though.

christophermaier:
A benefit of the `group members <group-name> etc.` syntax is that you don't have to carve out exceptions for group names. That is, if the scheme was like

```
group list
group create <group-name>
group delete <group-name>
group <group-name> add <user1> .. <userN> 
group <group-name> remove <user> .. <userN>
```

you might run into issues if someone created groups named `list`, `create`, or `delete`. Granted, this is probably not likely to happen in real life, but it'd be a pain for anyone that happened to hit it.

[...]

PR #688, opened 2016-05-12
Create group repository

mpeck:
This creates a groups repository and updates the group and group membership controllers accordingly. This is a prereq for https://github.com/operable/cog/issues/671

### DIFF
--- a/lib/cog/repository/groups.ex
+++ b/lib/cog/repository/groups.ex
@@ -6,6 +6,10 @@ defmodule Cog.Repository.Groups do
 
   alias Cog.Repo
   alias Cog.Models.Group
+  alias Cog.Models.User
+  alias Cog.Models.Role
+  import Ecto.Query, only: [from: 1, from: 2]
+
 
   @doc """
   Creates a new user group given a map of attributes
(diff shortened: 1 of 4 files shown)


lib/cog/repository/groups.ex
```diff
@@ -0,0 +1,215 @@
...
+    |> Group.changeset(attrs)
+    |> Repo.insert
+  end
+
+  @doc """
+  Retrieves all user groups
+  """
+  @spec all :: [%Group{}]
+  def all,
+    do: Repo.all(Group) |> Repo.preload([:direct_user_members, :direct_group_members, :roles])
```
christophermaier: I don't think you need `direct_group_members` anymore, since groups can't be members of groups.

If you can remove this and the tests still pass, go ahead and take it out. I know there are some places in our test setup code that still might take advantage of this. If that's the case, can you make a note to that effect here, so we can take out the preload once we tighten the tests up / remove all the outdated authz infrastructure?

lib/cog/repository/groups.ex
```diff
@@ -0,0 +1,215 @@
...
+  @doc """
+  Retrieves a single user group based on the id. The given id must be
+  a valid UUID.
+  """
+  @spec by_id(String.t) :: {:ok, %Group{}} | {:error, Ecto.Changeset.t} | {:error, Atom.t}
+  def by_id(id) do
+    if Cog.UUID.is_uuid?(id) do
+      case Repo.get(Group, id) do
+        %Group{} = group ->
+          group = Repo.preload(group, [:direct_user_members, :direct_group_members, :roles])
```
christophermaier: Same deal with `direct_group_members` here.

Actually, it might be nice to extract the preloads to a separate private function that we can call in the various places we retrieve groups to easily ensure that we're being consistent.

lib/cog/repository/groups.ex
christophermaier: Ah, nevermind... I see you're just pulling the logic from the controllers into a repository. There's still a lot of that authz infrastructure we need to yank out, so we can just leave this until then.

christophermaier:
Looks good; ignore my comments :)

PR #690, opened 2016-05-12
Expire channel in cache when channel events are received

### LINKED ISSUE #676: Allow the bot to post messages to rooms the invoking user is not a member of
Currently, if you try to redirect to a room you are not a member of you receive the following error even though the bot is a member of that room:

screen shot 2016-05-11 at 1 29 52 pm

This should not be an error and the bot should be able to write to any room they are a member of regardless of whether the invoking user is a member.

vanstee:
Closes #676 

We should also look into using the cached channels that the slack lib provides instead of keeping our own.

### DIFF
--- a/lib/cog/adapters/slack/api.ex
+++ b/lib/cog/adapters/slack/api.ex
@@ -70,6 +70,10 @@ defmodule Cog.Adapters.Slack.API do
     end
   end
 
+  def expire_channel(channel_id) when is_binary(channel_id) do
+    GenServer.call(@server_name, {:expire_channel, channel_id}, :infinity)
+  end
+
   def init(config) do
     token = config[:api][:token]
     cache_ttl = config[:api][:cache_ttl] || @default_ttl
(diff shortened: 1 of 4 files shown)


lib/cog/adapters/slack/api.ex
```diff
@@ -288,6 +297,10 @@ defmodule Cog.Adapters.Slack.API do
     end
   end
 
+  def expire_key(key) do
```
christophermaier: Can this be private?

vanstee: Ah yes it can. Fixing.

lib/cog/adapters/slack/rtm_connector.ex
```diff
@@ -112,6 +114,17 @@ defmodule Cog.Adapters.Slack.RTMConnector do
     end
     handler.()
   end
+  def handle_message(%{type: channel_event, channel: channel}, _slack, state) when channel_event in @channel_events do
+    channel_id = case channel do
+      channel when is_binary(channel) ->
+        channel
+      %{id: id} ->
```
christophermaier: What determines if `channel` is a binary or a map? Whether it's a "channel" event or a "group" event? If so, a comment might be useful here, just for reference.

christophermaier:
Looks good!

lib/cog/adapters/slack/rtm_connector.ex
vanstee: There's not really a method to it from what I can tell. Some events give you a map, and others an id. If there was a pattern I'd call it out, but a comment like "Some events give you a map, and others an id" is kinda just reiterating what the code already says.

christophermaier: Fair enough. Yay for gradually evolving APIs!

PR #692, opened 2016-05-13
Rework help command

### LINKED ISSUE #672: Redesign `help` command
Help is already on the right track. I want to change just a couple of things.

We need to make sure we're always returning a list of commands here. Although with this change, we wont be able to prefix the list with a string, but I think that's ok for now.

Also we should add short flags for `--all` (`-a`) and `--disabled` (`-d`).

vanstee:
`help` now returns a list of commands or a single command when running `help` or `help <command>`. However, this does limit what we can print since templates are rendered per item in the list and not for the entire response. For instance description at the top of the old help output ("I know about these commands:") is no longer included in the output.

Also any errors encountered are returned as errors rather than returning a string of output.

Closes #672

### DIFF
--- a/lib/cog/commands/help.ex
+++ b/lib/cog/commands/help.ex
@@ -1,29 +1,54 @@
 defmodule Cog.Commands.Help do
-  use Cog.Command.GenCommand.Base, bundle: "#{Cog.embedded_bundle}"
-
-  @moduledoc """
-  Get help on all installed Cog bot commands. By default only enabled
-  commands are listed.
-
-  USAGE:
-    help [OPTIONS] [ARGS]
-
-  OPTIONS:
-    -d, --disabled    list all disabled commands
-    -a, --all         list all know commands, enabled and disabled
-
-  ARGS:
-    command_name      list specific help for a command
-  """
+  use Cog.Command.GenCommand.Base,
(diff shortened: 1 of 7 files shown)


mpeck:
So is there a reason that we don't just return one result that contains a list of commands? I know that does limit what could be done with the help command in pipelines, but is it really important in this case?

vanstee:
@mpeck I guess it's really not super important, but it does seem to break the convention we have for returning sets of data. This is also the only place we return the commands installed, so I was considering that as a possibility. For instance `help | filter -p name -m "special-command"` to see if something is installed makes sense.

mpeck:
Yeah, I get ya. I thought about that. But then I thought that it might be more useful to use something like the bundle command for things like that. I'm not really arguing one way or the other. Generally I'm all about consistency. This command just feels a bit different. I mean honestly if we were being totally consistent with it, should we get a list of help objects instead of just a list of command names?

But like I said, I'm not super passionate about it. I just thought that it might be easier to format the output if we broke the rule in this very specific case. Of course, that could be a terrible idea.

vanstee:
I think a list of command objects make sense, because they contain the help documentation. Either way, it does show that we need a better way to format responses.

PR #693, opened 2016-05-13
Fix key in table template

vanstee:
The `table` command now returns `{"table" => ... table string ...}` so we need to print that from the template. See https://github.com/operable/cog/blob/master/lib/cog/commands/table.ex#L54


christophermaier:
👍

PR #696, opened 2016-05-16
Don't crash if an unknown id is passed to delete

mpeck:
If an unknown or bad id was passed to delete we would never render a response. Now we will.

### DIFF
--- a/web/controllers/v1/group_controller.ex
+++ b/web/controllers/v1/group_controller.ex
@@ -46,23 +46,25 @@ defmodule Cog.V1.GroupController do
   end
 
   def update(conn, %{"id" => id, "group" => group_params}) do
-    with {:ok, group} <- Groups.by_id(id) do
-      case Groups.update(group, group_params) do
-        {:ok, updated} ->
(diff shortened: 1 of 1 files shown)


web/controllers/v1/group_controller.ex
```diff
@@ -67,19 +67,21 @@ defmodule Cog.V1.GroupController do
...
-          |> put_status(:not_found)
-          |> json(%{errors: "Group not found"})
-        {:error, :bad_id} ->
-          conn
-          |> put_status(:bad_request)
-          |> json(%{errors: "Bad ID format"})
-      end
+    results = with {:ok, group} <- Groups.by_id(id) do
+      Groups.delete(group)
+    end
```
vanstee: This is probably ok, but I wonder if we could do a delete by id and check the result to see if we actually deleted anything or not, rather than doing a select followed by a delete.

mpeck: The repository takes a model and not an id. We could change that to accept an id instead, but I feel that there was a reason it was done this way. I just don't remember what it was.

PR #697, opened 2016-05-16
User's repository

mpeck:
Moves user bits into a users repository

prereq for #671

### DIFF
--- a/lib/cog/repository/users.ex
+++ b/lib/cog/repository/users.ex
@@ -1,0 +1,78 @@
+defmodule Cog.Repository.Users do
+  @moduledoc """
+  Behavioral API for interacting with users. Prefer these
+  functions over direct manipulation with 'Cog.Repo'.
+  """
+
+  alias Cog.Repo
+  alias Cog.Models.User
+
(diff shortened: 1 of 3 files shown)


lib/cog/repository/users.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  @doc """
+  Deletes a user
+  """
+  @spec delete(%User{}) :: {:ok, %User{}} | {:error, :not_found}
+  def delete(%User{}=user) do
+    try do
+      Repo.delete(user)
+    rescue
+      Ecto.StateModelError ->
```
vanstee: Is this supposed to be `StaleModelError`? Also surprised that the `delete` func raises like this. Did you mean `delete!`?

mpeck: Actually we may not need this. I copied it from the triggers repository. I think it only get's triggered if you try to delete the same model twice.

mpeck: I'll just drop it for now.

christophermaier:
👍

PR #708, opened 2016-05-18
Group command rework

### LINKED ISSUE #671: Redesign `group` command

mpeck:
Resolves #671


lib/cog/commands/group.ex
```diff
@@ -1,170 +1,190 @@
...
-  defp drop_group([name | _]) do
-    case Repo.get_by(Group, name: name) do
-      %Group{}=group ->
-        Repo.delete!(group)
-        MessageTranslations.translate_success(%{action: :drop, entity: group})
-      nil -> MessageTranslations.translate_error("group", {:does_not_exists, name})
+      invalid ->
+        suggestion = Enum.max_by(["list", "create", "delete", "member", "role", "info"],
+                                 &String.jaro_distance(&1, invalid))
+        show_usage(error({:unknown_subcommand, invalid, suggestion}))
```
vanstee: kewl

lib/cog/commands/group.ex
```diff
@@ -1,170 +1,190 @@
...
+        {:reply, req.reply_to, template, data, state}
+      {:ok, message} ->
+        {:reply, req.reply_to, message, state}
+      {:error, err} ->
+        {:error, req.reply_to, Helpers.error(err), state}
     end
   end
 
+  defp error(:required_subcommand),
+    do: "You must specify a subcommand. Please specify one of, 'list', 'create', 'delete' or 'member'"
```
vanstee: Is this still used now that we default to `list`?

lib/cog/commands/group.ex
```diff
@@ -1,170 +1,190 @@
...
+    #%{input | errors: [:missing_group | errors]}
+  #end
+  #defp validate_group(%{req: %{args: [name | _]}, errors: errors}=input) do
+    #case Repo.get_by(Group, name: name) do
+      #%Group{}=group -> %{input | classification: group}
+      #nil -> %{input | errors: [{:unrecognized_group, name} | errors]}
+    #end
+  #end
+
+#end
```
vanstee: Looks like leftover code.

lib/cog/commands/group/create.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Cog.Commands.Group.Create do
+  require Cog.Commands.Helpers, as: Helpers
+  alias Cog.Repository.Groups
+
+  Helpers.usage """
+  Creates new user groups.
+
+  USAGE
+    group create [FLAGS] <group_name>
```
vanstee: I've been using dasherized names (like `group-name`) but this is fine.

lib/cog/commands/group/create.ex
```diff
@@ -0,0 +1,40 @@
...
+  ARGS
+    group_name    The name of the user group to create
+
+  FLAGS
+    -h, --help    Display this usage info
+  """
+
+  @spec create_group(%Cog.Command.Request{}, List.t) :: {:ok, String.t, Map.t} | {:error, any()}
+  def create_group(req, arg_list) do
+    if Helpers.flag?(req.options, "help") do
```
vanstee: The helper is nice, but it might actually be a bit nicer to use pattern matching for this stuff. `def create_group(%{options: %{"help" => true}}, _args)` is clear enough to me at least.

lib/cog/commands/group.ex
mpeck: Ah, good point, I'll get rid of it.

lib/cog/commands/group.ex
mpeck: Geez, that'll teach me to not scroll to the bottom :|

lib/cog/commands/group/create.ex
mpeck: We should be consistent. I'll change to using dashes.

lib/cog/commands/group/list.ex
```diff
@@ -0,0 +1,37 @@
...
+  """
+
+  @spec list_groups(%Cog.Command.Request{}, List.t) :: {:ok, String.t, Map.t} | {:error, any()}
+  def list_groups(req, _arg_list) do
+    if Helpers.flag?(req.options, "help") do
+      show_usage
+    else
+      case Groups.all do
+        [] ->
+          {:ok, "Currently, there are no groups in the system."}
```
vanstee: This is kinda a bummer, but I get why we need to do this.

lib/cog/commands/group.ex
```diff
@@ -1,170 +1,73 @@
...
-    group --list
-
-  EXAMPLE:
-    group --create ops
-    > The group 'ops' has been created.
-
-    group --create engineering
-    > The group 'engineering' has been created.
+  use Cog.Command.GenCommand.Base, bundle: Cog.embedded_bundle
+  require Cog.Commands.Helpers, as: Helpers
```
vanstee: Is this the same as `alias Cog.Commands.Helpers`?

mpeck: No, in order to include macros you have to use `require`

lib/cog/commands/group/list.ex
mpeck: Yeah, when we level up the error system these will go away.

lib/cog/commands/group.ex
vanstee: Oh nice TIL

PR #709, opened 2016-05-18
Redesign rules

vanstee:
Rules now no longer uses the flag syntax and was switched to a much more succinct subcommand version.

```
rule list
rule create ec2:destroy site:ops
rule create "when command is ec2:destroy must have site:ops"
rule drop -c ec2:destroy
```


lib/cog/commands/rule.ex
```diff
@@ -0,0 +1,80 @@
...
+  FLAGS
+    -h, --help  Display this usage info
+
+  SUBCOMMANDS
+    list    List all rules (default)
+    create  Add a rule
+    drop    Drop a rule
+  """
+
+  option "help", type: "bool", short: "h"
```
mpeck: The `Helpers.usage` macro takes care of the help option and the show_usage function.

lib/cog/commands/rule/drop.ex
```diff
@@ -0,0 +1,79 @@
+defmodule Cog.Commands.Rule.Drop do
+  @moduledoc """
+  Drops rules by id or all rules for a specific command.
+
+  USAGE
+    rule drop [FLAGS] [OPTIONS] [<id> ...]
```
mpeck: I've been specifying lists of required args like `<arg-list ...>`

lib/cog/commands/rule/drop.ex
```diff
@@ -0,0 +1,79 @@
...
+  USAGE
+    rule drop [FLAGS] [OPTIONS] [<id> ...] 
+
+  FLAGS
+    -h, --help  Display this usage info
+
+  OPTIONS
+    -c, --command  Drops rules belonging to the command
+
+  ARGS
```
mpeck: Not that it matters, but I've been putting `ARGS` at the top. My thinking was that `ARGS` are more important since they're often required.

lib/cog/commands/rule/create.ex
```diff
@@ -0,0 +1,68 @@
...
+  defp do_create(rule) do
+    with :ok <- validate_permissions(rule),
+         {:ok, rule} <- create_rule(rule),
+         do: {:ok, rule}
+  end
+
+  defp show_usage(error \\ nil),
+    do: {:ok, "usage", %{usage: @moduledoc, error: error}}
+
+  defp validate_permissions(rule) do
```
mpeck: I wonder if rules would be worth putting into a "repository"? Not suggesting that we do that here, but I think we do things like this in other places. It might be worth consolidating.

christophermaier: `Cog.RuleIngestion` is effectively our rule repository (and should probably be renamed to reflect that). I think that this logic already lives in the installation process, actually. If it doesn't, we should add it, and make sure that we can get the kind of error reporting out of it that we want here.

lib/cog/models/command.ex
```diff
@@ -50,14 +50,53 @@ defmodule Cog.Models.Command do
...
+    case parse_qualified_name(name) do
+      {:ok, command} ->
+        {:ok, command}
+      {:error, _} ->
+        parse_ambiguous_name(name)
+    end
+  end
+
+  def parse_qualified_name(name) do
+    case String.split(name, ":", parts: 2) do
```
mpeck: I feel like I've seen a function that did this exact thing before. Does that still exist?

lib/cog/commands/rule/create.ex
mpeck: I suspected that, but it looks like there are still several spots where we are using `Repo` directly. Maybe rules should be the next resource we look at for cleaning up it's api?

lib/cog/commands/rule.ex
vanstee: Ah nice I'll use that.

lib/cog/commands/rule/drop.ex
vanstee: :+1:

lib/cog/commands/rule/drop.ex
vanstee: Ah I like that better.

lib/cog/models/command.ex
vanstee: It's actually right above this, but has some weird functionality. I think we still use it in a few places so I didn't rip it out. Let me see if I can without too much trouble.

PR #719, opened 2016-05-24
Bundle API Changes

### LINKED ISSUE #642: Version-related API changes
- [ ] bundle deletion is version aware
- [ ] bundle enable/disable is version aware
- [ ] dry-run option for bundle delete
- [ ] bundle installation changes: update messaging at least
- [ ] API changes

`POST /v1/bundles` - installation (the only way)
`GET /v1/bundles/:name` - return list of all installed versions
`GET /v1/bundles/:name/:version` - details about specific version
`DELETE /v1/bundles/:name/:version` - delete specific version (must be disabled first)
`DELETE /v1/bundles/:name` - if given an explicit list of versions, delete each one. If not, delete _all_ installed versions. Atomic.

There is no `PUT`.

christophermaier:
Various tweaks and refinements to the new bundle API.

Fixes #642


lib/cog/models/bundle_version.ex
```diff
@@ -23,4 +28,18 @@ defmodule Cog.Models.BundleVersion do
     model |> cast(params, @required_fields, [])
   end
 
+  def enabled?(bundle_version) do
+    if Ecto.assoc_loaded?(bundle_version.enabled_version_registration) do
+      if bundle_version.enabled_version_registration do
```
kevsmith: Could this be `bundle_version.enabled_version_registration != nil` instead of the `if`?

web/controllers/v1/bundle_version_controller.ex
```diff
@@ -12,23 +12,31 @@ defmodule Cog.V1.BundleVersionController do
...
     end
   end
 
-  def delete(conn, %{"id" => id}) do
+  def delete(conn, %{"bundle_version_id" => id}) do
     case Repository.Bundles.version(id) do
       %BundleVersion{}=bv ->
         case Repository.Bundles.delete(bv) do
           {:ok, _} ->
             send_resp(conn, 204, "")
```
kevsmith: I thought Phoenix handled sending the correct status code?

web/router.ex
```diff
@@ -45,11 +45,10 @@ defmodule Cog.Router do
     get    "/v1/bundles/:id", V1.BundlesController, :show
     delete "/v1/bundles/:id", V1.BundlesController, :delete
 
-    # TODO: Maybe this is just :show on /bundles/:id?
     get "/v1/bundles/:id/versions", V1.BundlesController, :versions
 
-    # TODO: maybe this is just /v1/bundle_versions/:bundle_version_id
-    resources "/v1/bundles/:id/versions/:bundle_version_id", V1.BundleVersionController, only: [:show, :delete]
+    get    "/v1/bundles/:id/versions/:bundle_version_id", V1.BundleVersionController, :show
```
kevsmith: Why separate lines for each action instead of combining them on one with `only:` ?

web/views/bundles_view.ex
```diff
@@ -1,92 +1,45 @@
...
-  #   |> Enum.reduce(%{}, fn(field, reply) ->
-  #     case render_include(field, resource) do
-  #       nil -> reply
-  #       {key, value} -> Map.put(reply, key, value)
-  #     end
-  #   end)
-  # end
+  ########################################################################
+
+  defp ordered_versions(versions) do
```
kevsmith: Would Elixir's `Version` module make working with versions easier?

web/router.ex
christophermaier: I tried that earlier, but it ended up not working. In retrospect, I suspect that sticking the `:id` key up in the root (instead of having Phoenix tack it onto the generated resource paths) was the issue. I'd bet that changing `:id` to `:bundle_id`, and `:bundle_version_id` to `:id` where it's used would fix it.

web/views/bundles_view.ex
christophermaier: We actually are using Elixir's `Version` for our custom Ecto `VersionTriple` type. Perhaps I should have named this variable `bundle_versions` instead of `versions`, since they're the models.

web/controllers/v1/bundle_version_controller.ex
christophermaier: Phoenix does some automatic response handling if the controller code throws an exception, provided you've supplied the appropriate exception -> status code mappings. I don't think that applies here.

lib/cog/models/bundle_version.ex
christophermaier: Derp, yes.

web/router.ex
christophermaier: Ha, I was right. Fix incoming.

kevsmith:
:shipit:

PR #723, opened 2016-05-25
Override service and trigger base URLs

### LINKED ISSUE #694: Make sure the HTTP API endpoint URL generation is SSL aware
We need to make sure that things like service URLs that are generated can correctly generate https URLs when SSL is configured.

kevsmith:
This commit introduces two new environment variables which can be used
to override the base URLs for services and triggers:

$COG_TRIGGER_URL_BASE
$COG_SERVICE_URL_BASE

Setting either/both of these variables will override the base portion of the corresponding URL. This capability is useful when Cog is unable to detect the correct routable URL to send to clients. For example, firewalls and load balancers interpose themselves between clients and servers and require hosts behind them to carefully build any URLs returned in output so clients are able to access those resources.

Fixes #694

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -148,6 +148,9 @@ config :cog, Cog.ServiceEndpoint,
   check_origin: true,
   render_errors: [accepts: ~w(json)]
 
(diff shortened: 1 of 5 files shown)


imbriaco:
LGTM!

PR #724, opened 2016-05-25
Delete & recreate embedded bundle

### LINKED ISSUE #721: Always auto-upgrade the embedded bundle when MIX_ENV=dev
Current behavior is to only auto-upgrade when the application version changes, but this can be problematic for developers making changes to parts of that bundle.

kevsmith:
This commit causes Cog to delete and rebuild the embedded bundle at boot whenever `Mix.env == :dev`.

Fixes #721

### DIFF
--- a/lib/cog/repository/bundles.ex
+++ b/lib/cog/repository/bundles.ex
@@ -50,6 +50,17 @@ defmodule Cog.Repository.Bundles do
     end
   end
 
+
+  # Deletes the embedded bundle when Mix.env == :dev so Cog can recreate the bundle
+  # at startup.
+  defp __reset_embedded() do
+    if Mix.env == :dev do
+      Logger.info("Dev mode detected. Resetting embedded bundle.")
+      Repo.delete_all(from b in Bundle,
+                   where: b.name == ^Cog.embedded_bundle)
+    end
+  end
+
   @doc """
   Return all known versions for `bundle`. Ensures that all required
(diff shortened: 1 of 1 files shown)


lib/cog/repository/bundles.ex
```diff
@@ -50,6 +50,17 @@ defmodule Cog.Repository.Bundles do
...
   end
 
+
+  # Deletes the embedded bundle when Mix.env == :dev so Cog can recreate the bundle
+  # at startup.
+  defp __reset_embedded() do
+    if Mix.env == :dev do
+      Logger.info("Dev mode detected. Resetting embedded bundle.")
+      Repo.delete_all(from b in Bundle,
+                   where: b.name == ^Cog.embedded_bundle)
```
christophermaier: I think we might need to delete the current bundle version instead of the bundle itself... otherwise I suspect we lose all permissions, and then any users in the system will lose the ability to do anything

lib/cog/repository/bundles.ex
```diff
@@ -50,6 +50,17 @@ defmodule Cog.Repository.Bundles do
     end
   end
 
+
+  # Deletes the embedded bundle when Mix.env == :dev so Cog can recreate the bundle
+  # at startup.
+  defp __reset_embedded() do
+    if Mix.env == :dev do
```
christophermaier: We probably don't need this conditional, since it's also embedded at this function's call site

PR #725, opened 2016-05-25
Fix help to work with bundle verisons

vanstee:


kevsmith:
Some of these queries are gnarly but....I'm in favor of merging so we can get to a running system again and take stock.

PR #734, opened 2016-05-26
Filter site bundle from repository.Bundles output

### LINKED ISSUE #733: Filter out "site" bundle from lists of bundles
It's really only a bundle in an implementation sense; exposing it to the user as one probably just raises questions and is unconstructive.

kevsmith:
Fixes #733

### DIFF
--- a/test/cog/repository/bundles_test.exs
+++ b/test/cog/repository/bundles_test.exs
@@ -43,12 +43,12 @@ defmodule Cog.Repository.BundlesTest do
     assert %BundleVersion{version: ^fixed_version, bundle: %Bundle{name: "site"}} = Bundles.site_bundle_version
   end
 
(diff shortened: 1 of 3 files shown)


test/cog/repository/bundles_test.exs
```diff
@@ -43,7 +43,7 @@ defmodule Cog.Repository.BundlesTest do
     assert %BundleVersion{version: ^fixed_version, bundle: %Bundle{name: "site"}} = Bundles.site_bundle_version
   end
 
-  Enum.each(["operable", "site", "user"], fn(protected_name) ->
+  Enum.each(["operable", "user"], fn(protected_name) ->
```
christophermaier: We should still test that somebody can't come in and create a bundle named "site".

PR #736, opened 2016-05-27
Formally define contract for parser metadata

christophermaier:
Previously, during parsing, we would return a `CommandVersion` model
corresponding to the specific command details for the appropriate
enabled bundle; this "metadata" was then utilized in several places in
the pipeline execution flow. In particular, this is the mechanism by
which we retrieved the rules to evaluate to determine if a user was
authorized to execute a given command.

With the introduction of bundle versions, the association of
user-entered rules with the "site" bundle, and the internal
enabling/disabling of rules, retrieval of rules for a given command
became more complicated, particularly since such requirements cannot
currently be expressed as an Ecto association. A minimal solution would
be to retrieve the appropriate rules "out of band" and then replace the
"wrong" `command_version.command.rules` data in the model with
them. While this works, it would mean that the same model field would
have two different meanings, depending on where it was used. That way
lies madness, though.

As an alternative, this commit concretely defines what we need from our
parser metadata. A new `Cog.Command.Pipeline.ParserMeta` struct is
introduced, with fields determined by tracing the flow of the parser
metadata through the executor. Instead of abusing a `CommandVersion`
struct, we now generate a `ParserMeta` struct in the `CommandResolver`.

A test was added to verify that the appropriate site rules and bundle
rules are retrieved, and that enabled status of rules is respected.

### DIFF
--- a/lib/cog/audit_message.ex
+++ b/lib/cog/audit_message.ex
@@ -17,8 +17,8 @@ defmodule Cog.AuditMessage do
   end
   def render({:denied, {_rule, current_invocation}}, request),
     do: "User #{request["sender"]["handle"]} denied access to '#{current_invocation}'"
-  def render({:no_relays, bundle}, _request),
-    do: ErrorResponse.render({:no_relays, bundle}) # Uses same user message for now
+  def render({:no_relays, bundle_name}, _request),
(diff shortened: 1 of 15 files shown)


kevsmith:
👍 

I like the way this codifies the interface between Piper and Cog, too.

PR #737, opened 2016-05-27
Ensure bundle installation is truly transactional

christophermaier:
The easiest way to ensure that we fail the installation transaction is
if everything that could fail raises an exception of some kind. Many
functions already did that but weren't named with the `!` mnemonic;
those have been fixed to make reasoning about the code a bit easier.

Additionally (and crucially) failure of rule ingestion outside of a
transaction of its own will now raise a new `Cog.RuleIngestionError` if
anything fails. This then bubbles up to trigger transaction rollbacks as
appropriate.

Tests were added.


kevsmith:
+1

Nice clean up with `!` semantics.

PR #740, opened 2016-05-27
Fix bundle command to work with bundle versions

vanstee:


lib/cog/commands/bundle.ex
```diff
@@ -69,70 +75,130 @@ defmodule Cog.Commands.Bundle do
...
+  def handle_message(%{args: ["status", bundle_name]} = req, state) do
+    case Bundles.with_status_by_name(bundle_name) do
+      nil ->
+        {:error, req.reply_to, error_message({:not_found, bundle_name}), state}
+      bundle ->
+        {:reply, req.reply_to, "bundle-status", bundle, state}
+    end
+  end
+
+  def handle_message(%{args: [action, "operable"|_]} = req, state)
```
christophermaier: This should be handled in the repository function call for you... I don't think you need to special-case it here.

(Incidentally, you shouldn't be able to enable or disable the `site` bundle either.)

lib/cog/commands/bundle.ex
```diff
@@ -69,70 +75,130 @@ defmodule Cog.Commands.Bundle do
...
+      {:error, error} ->
+        {:error, req.reply_to, error_message(error), state}
+    end
+  end
+
+  def handle_message(req, state) do
+    {:error, req.reply_to, error_message(:invalid_invocation), state}
+  end
+
+  defp enable_bundle(%BundleVersion{}=bundle_version) do
```
christophermaier: Since nothing is happening to the `bundle_version` you pass in (the same one is being passed back out), I think you can just call `Bundles.set_bundle_version_status/2` wherever you call this, but match on `:ok` instead.

lib/cog/commands/bundle.ex
```diff
@@ -69,70 +75,130 @@ defmodule Cog.Commands.Bundle do
...
+    end
+  end
+
+  def handle_message(%{args: [action, "operable"|_]} = req, state)
+      when action in ["enable", "disable"] do
+    {:error, req.reply_to, error_message({:embedded_bundle, action}), state}
+  end
+
+  def handle_message(%{args: ["enable", bundle_name|args]} = req, state) do
+    result = with {:ok, bundle} <- parse_and_find_version(bundle_name, args),
```
christophermaier: It's a little confusing to call it `bundle` when it's actually a `bundle_version`... applies to a few other places in the code, too.

lib/cog/commands/bundle.ex
```diff
@@ -69,70 +75,130 @@ defmodule Cog.Commands.Bundle do
...
+        {:error, {:invalid_version, version}}
+    end
+  end
+
+  def check_for_enabled_version(bundle_name) do
+    case Bundles.enabled_version_by_name(bundle_name) do
+      nil ->
+        :ok
+      bundle ->
+        {:error, {:already_enabled, bundle}}
```
christophermaier: Why is it an error if the bundle is already enabled?

lib/cog/commands/bundle.ex
```diff
@@ -69,70 +75,130 @@ defmodule Cog.Commands.Bundle do
...
+    end
+  end
+
+  def find_enabled_bundle(bundle_name) do
+    case Bundles.enabled_version_by_name(bundle_name) do
+      nil ->
+        {:error, {:already_disabled, bundle_name}}
```
christophermaier: What if the user typed the wrong name (i.e., the bundle doesn't exist)?

lib/cog/commands/bundle.ex
vanstee: You may be trying to enable a different version of a bundle that is already enabled. I could specifically handle the case where you try to enable the version of the bundle that is already enabled and return a successful response, but this seemed like a reasonable way to return a consistent response.

lib/cog/repository/bundles.ex
```diff
@@ -194,6 +195,69 @@ defmodule Cog.Repository.Bundles do
...
+            select: %{bundle_version: bv, enabled: not(is_nil(e.bundle_id))},
+            order_by: [desc: not(is_nil(e.version)), desc: bv.version],
+            limit: 1
+
+    case Repo.one(query) do
+      nil ->
+        nil
+     %{bundle_version: bundle_version, enabled: true} ->
+       bundle_version
+       |> Map.put(:status, :enabled)
```
christophermaier: The virtual field on the model says this should be a string instead of an atom

lib/cog/repository/bundles.ex
```diff
@@ -194,6 +195,69 @@ defmodule Cog.Repository.Bundles do
...
+       |> preload
+     %{bundle_version: bundle_version, enabled: false} ->
+       bundle_version
+       |> Map.put(:status, :disabled)
```
christophermaier: Ditto

lib/cog/repository/bundles.ex
vanstee: Good point, I'll swap these atoms with strings.

lib/cog/commands/bundle.ex
vanstee: Not sure how I missed that. I'll change the return values of this func to return a more descriptive error.

PR #741, opened 2016-05-31
Do not assign special bundles to relay groups

christophermaier:
We shouldn't be able to assign the `operable` or `site` bundles to a relay group. Assigning the latter should be harmless, though confusing; assigning the former would probably break everything, since it's an Elixir bundle and Relay doesn't know anything about those.

### DIFF
--- a/lib/cog/protected_bundle_error.ex
+++ b/lib/cog/protected_bundle_error.ex
@@ -1,8 +1,8 @@
 defmodule Cog.ProtectedBundleError do
-  defexception [:message, :reason]
+  defexception [:message, :bundle]
 
-  def exception(reason),
-    do: %__MODULE__{message: "Operation not permitted on a protected bundle: #{inspect reason}",
-                    reason: reason}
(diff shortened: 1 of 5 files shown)


lib/cog/protected_bundle_error.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Cog.ProtectedBundleError do
+  defexception [:message, :reason]
+
+  def exception(reason),
+    do: %__MODULE__{message: "Operation not permitted on a protected bundle: #{inspect reason}",
+                    reason: reason}
```
vanstee: Does this mean the the message here will end up being `Operation not permitted on a protected bundle: {:protected_bundle, "operable"}`. Seems like we'd really only need to send the bundle name arg as the `:protected_bundle` atom is sorta redundant.

christophermaier: Derp

vanstee:
LGTM

PR #742, opened 2016-05-31
Be more accepting in what we consider a "version"

christophermaier:
Previously, we strictly required a String in `"major.minor.patch"`
format for a version. Now, we'll accept any of the following:

* major version, as integer
* major.minor version as float
* "major" version as string
* "major.minor" version as string
* "major.minor.patch" version as string (previous behavior)

That is, `1`, `1.0`, `"1"`, `"1.0"`, and `"1.0.0"` are all equivalent
now.

### DIFF
--- a/lib/cog/models/types/version_triple.ex
+++ b/lib/cog/models/types/version_triple.ex
@@ -24,8 +24,15 @@ defmodule Cog.Models.Types.VersionTriple do
         # Major and minor version; e.g. "1.0" == "1.0.0"
         Version.parse(text <> ".0")
       true ->
-        # Treat it like semver
-        Version.parse(text)
+        # Treat it like semver, but only if it's just
+        # major.minor.patch... we don't handle prerelease or build
+        # metadata at the moment
+        case Version.parse(text) do
+          {:ok, %Version{pre: [], build: nil}=v} ->
+            {:ok, v}
+          _ ->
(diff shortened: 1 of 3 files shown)


imbriaco:
These semantics make sense, :+1:. 

The only thing I think this leaves as a potential gotcha is the fact that we don't allow pre-release patch versions (e.g. "1.0.1-foo1") in the database, but the `Version` module will happily accept them. Do we want to handle that case up front or are we happy with letting the database be the arbiter for those?

lib/cog/models/types/version_triple.ex
```diff
@@ -13,8 +13,21 @@ defmodule Cog.Models.Types.VersionTriple do
...
+    cond do
+      Regex.match?(~r/\A\d+\z/, text) ->
+        # Just a major version; e.g. "1" == "1.0.0"
+        Version.parse(text <> ".0.0")
+      Regex.match?(~r/\A\d+\.\d+\z/, text) ->
+        # Major and minor version; e.g. "1.0" == "1.0.0"
+        Version.parse(text <> ".0")
+      true ->
+        # Treat it like semver
+        Version.parse(text)
```
vanstee: Nice, I like how this still allows for pre and build segments.

imbriaco:
:shipit:

PR #747, opened 2016-05-31
Allow for enabling other versions of already enabled commands

vanstee:


mpeck:
Looks good to me :+1:

PR #748, opened 2016-05-31
Use bundle instead of namespace

vanstee:


christophermaier:
We should use `Cog.Repository.Permissions.create_permission/1` here instead.

PR #749, opened 2016-05-31
First stab at handling overly large chat responses

### LINKED ISSUE #739: Warn from executor when responding with too big of a message
We should log a warning to the console from the executor before sending a message which will be too long for the chat provider to handle.

Once this works you should see a warning in the console after running `!help` with the hipchat adapter.

christophermaier:
Chat adapters can have limits on the size of message they will
accept. While a complete solution will deal with this in an
adapter-specific manner (breaking long messages into multiple messages,
using an alternative format, etc.), we can at least log a message when
we see a message that is potentially "too big".

Here, we add a new option recognized by
`Carrier.Messaging.Connection.publish/3` that lets a caller specify a
size threshold in bytes. If the JSON-encoded message is larger than this
threshold, we emit a log warning.

Fixes #739

### DIFF
--- a/lib/carrier/messaging/connection.ex
+++ b/lib/carrier/messaging/connection.ex
@@ -96,7 +96,21 @@ defmodule Carrier.Messaging.Connection do
   """
   def publish(conn, message, kw_args) when is_map(message) do
     topic = Keyword.fetch!(kw_args, :routed_by)
-    :emqttc.publish(conn, topic, Poison.encode!(message))
+
+    encoded = Poison.encode!(message)
+    case Keyword.fetch(kw_args, :threshold) do
+      {:ok, threshold} ->
+        size = byte_size(encoded)
+        if size > threshold do
+          Logger.warn("Message potentially too long (#{size} bytes)")
+        else
+          :ok
+        end
+      :error ->
+        :ok
+    end
+
+    :emqttc.publish(conn, topic, encoded)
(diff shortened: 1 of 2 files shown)


lib/carrier/messaging/connection.ex
```diff
@@ -96,7 +96,21 @@ defmodule Carrier.Messaging.Connection do
...
   def publish(conn, message, kw_args) when is_map(message) do
     topic = Keyword.fetch!(kw_args, :routed_by)
-    :emqttc.publish(conn, topic, Poison.encode!(message))
+
+    encoded = Poison.encode!(message)
+    case Keyword.fetch(kw_args, :threshold) do
+      {:ok, threshold} ->
+        size = byte_size(encoded)
+        if size > threshold do
+          Logger.warn("Message potentially too long (#{size} bytes)")
```
mpeck: Should we also include the threshold in the response?

kevsmith: What would that allow us to do?

mpeck: Thought it might be useful information for the user to have.

kevsmith: It'll vary by chat provider and we don't really have a way to handle that cleanly yet.

kevsmith:
👍

PR #750, opened 2016-05-31
Retry failed HipChat room lookups

### LINKED ISSUE #59: Cache HipChat user and room information, allowing use of rooms with uppercase names
NOTE: this is based on Hedwig 0.3.0, but a new version is coming out soon with a markedly different codebase

Currently when we receive a message from the Hipchat adapter, it looks something like this:

```elixir
%Hedwig.Stanzas.Message{body: "@marvin help",
                        client: #PID<0.775.0>,
                        delayed?: false,
                        from: %Hedwig.JID{resource: "Christopher Maier",
                                          server: "conf.hipchat.com",
                                          user: "123456_bot-testing"},
                        html: nil,
                        id: "",
                        matches: nil,
                        payload: [%{name: "body",
                                    payload: ["@marvin help"]}],
                        to: %Hedwig.JID{resource: "<big long proxy string>",
                                        server: "chat.hipchat.com",
                                        user: "123456_1234567"},
                        type: "groupchat"}     
```

Currently, we've been slicing up the `from.user` string to extract a room name (above, this would be `"bot-testing"`). We then resolve the ID of this room by using the Hipchat V2 API call `/v2/room/bot-testing` (see https://www.hipchat.com/docs/apiv2/method/get_room)

However, if the room name begins with an uppercase letter (e.g. `"Operable"`), the sliced-out version of the name will be `"operable"` (note lower-case). Making an API call with this name will return a 404.

Instead what we'll want to do is to start caching both room and user information, just like we do in the Slack adapter. Among the data returned from the Hipchat API for a room is the `xmppp_jid`, which for our room above would be `"123456_bot-testing@conf.hipchat.com"` (that is,`"#{from.user}@#{from.server}"`, or `Hedwig.JID.bare(jid)` (see
https://github.com/hedwig-im/hedwig/blob/e7292ecab53ce119dd251f8378fdcf7facf037a3/lib/hedwig/jid.ex#L50-L62)

We can use this "bare JID" to query our room cache to retrieve a room record, and from there obtain the room ID. This will allow the bot to post back to rooms with upppercase names :)

kevsmith:
This is a incredibly hacky fix to address Cog's current difficulties
interacting with capitalized HipChat rooms. The core issue is the HipChat
send message API requires a room ID. Normally we obtain this ID via HipChat's
API to look up room IDs from room names.

The complicating bit is that the room name sent to us in the XMPP message is
downcased which causes the ID look up to fail. There appears to be no other
useful information embedded in the XMPP message which gives us enough information
to successfully look up the room.

I suspect the correct long term solution here is to see if we can move the bulk
of our interactions w/HipChat to purely XMPP. We could then avoid these types
of impedance mismatches. In the short term, this commit implements a complete hack
which hopefully enables the majority of users hitting this issue to work. To wit,
when the HipChat adapter fails to look up a room by name it retries ~~up to two more
times~~ once more using the capitalized room name before giving up. ~~The first retry capitalizes the room name. If that lookup fails the adapter uses the `Inflex` library to camel case the room name.~~ If neither look ups succeed then an error is returned.

Fixes #59

### DIFF
--- a/lib/cog/adapters/hipchat/api.ex
+++ b/lib/cog/adapters/hipchat/api.ex
@@ -91,11 +91,11 @@ defmodule Cog.Adapters.HipChat.API do
     result = case get(state.client, "/room/#{room_name}") do
       {:ok, room} ->
         {:ok, normalize_room(room)}
-      _error ->
+      error ->
         case get(state.client, "/room/#{String.capitalize(room_name)}") do
           {:ok, room} ->
             {:ok, normalize_room(room)}
-          error ->
+          _error ->
             error
         end
     end
--- a/lib/cog/adapters/hipchat/api.ex
+++ b/lib/cog/adapters/hipchat/api.ex
@@ -88,12 +88,17 @@ defmodule Cog.Adapters.HipChat.API do
   end
 
   def handle_call({:lookup_room, name: room_name}, _from, state) do
-    uri = "/room/#{room_name}"
-
-    result = with {:ok, room} <- get(state.client, uri) do
-      {:ok, normalize_room(room)}
+    result = case get(state.client, "/room/#{room_name}") do
+      {:ok, room} ->
+        {:ok, normalize_room(room)}
+      _error ->
+        case get(state.client, "/room/#{String.capitalize(room_name)}") do
+          {:ok, room} ->
+            {:ok, normalize_room(room)}
+          _error ->
+            get(state.client, "/room/#{Inflex.camlize(room_name)}")
+        end
     end
-
     {:reply, result, state}
   end
 
--- a/lib/cog/adapters/hipchat/api.ex
+++ b/lib/cog/adapters/hipchat/api.ex
@@ -95,8 +95,8 @@ defmodule Cog.Adapters.HipChat.API do
         case get(state.client, "/room/#{String.capitalize(room_name)}") do
           {:ok, room} ->
             {:ok, normalize_room(room)}
-          _error ->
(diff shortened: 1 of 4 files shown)


lib/cog/adapters/hipchat/api.ex
```diff
@@ -88,12 +88,17 @@ defmodule Cog.Adapters.HipChat.API do
...
-      {:ok, normalize_room(room)}
+    result = case get(state.client, "/room/#{room_name}") do
+      {:ok, room} ->
+        {:ok, normalize_room(room)}
+      _error ->
+        case get(state.client, "/room/#{String.capitalize(room_name)}") do
+          {:ok, room} ->
+            {:ok, normalize_room(room)}
+          _error ->
+            get(state.client, "/room/#{Inflex.camlize(room_name)}")
```
mpeck: Is it `Inflex.camelize`?

mpeck: Also, will the camelize bit actually catch anything? It looks like camelize just camel cases words with '_' or '-' delimiters. If that be the case it would only catch things like 'Room_Name'.

kevsmith: Good points. I'll nuke using Inflex.

vanstee:
Looks good as a quick-fix for the 0.7.0 release. :+1:

It does look like the correctly capitalized room name is hidden in some presence messages (found this in the logs after joining the room `%{name: "name", payload: ["CaPs"]}`). Still seems like we would need to do a decent amount of work to use this room name though.

PR #754, opened 2016-06-02
Allow Triggers to receive x-www-form-urlencoded payloads as well as JSON

imbriaco:
This PR will allow triggers to accept and decode POST bodies that are encoded as `x-www-form-urlencoded` with the key/value pairs mapped to `cog_env.body` in the same was as JSON payloads. This allows triggers to serve as an endpoint for webhooks like Slack's slash commands that do not encode the body as JSON.

### Example:

Simply sends the text after the slash command to @imbriaco via DM:

```
 $ cogctl triggers create --name cog_register --pipeline 'echo $body.text > chat://@imbriaco' --as-user admin --enabled
ID              f9d362a5-910a-4941-bafb-3a277015bf6b
Name            cog_register
Pipeline        echo $body.text > chat://@imbriaco
Enabled         true
As User         admin
Timeout (sec)   30
Description
Invocation URL  http://localhost:4001/v1/triggers/f9d362a5-910a-4941-bafb-3a277015bf6b
```

### Slack Setup:

slash commands operable slack 2016-06-02 11-22-04

### Result:

The following was generated in response to `/cog testing, 123` being sent via the Slack client.

slack 2016-06-02 11-23-14

### DIFF
--- a/test/controllers/v1/trigger_execution_controller_test.exs
+++ b/test/controllers/v1/trigger_execution_controller_test.exs
@@ -7,9 +7,14 @@ defmodule Cog.V1.TriggerExecutionControllerTest do
   @endpoint Cog.TriggerEndpoint
   @bad_uuid "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
 
-  setup do
+  setup context do
+    content_type = case context[:content_type] do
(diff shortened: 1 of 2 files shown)


christophermaier:
👍

PR #756, opened 2016-06-02
use alpine 3.4 and the default erlang/elixir

imbriaco:
- Updated image to use Alpine 3.4 
  - Note: Gliderlabs images not yet available so we built our own from their tooling
- Use default Elixir and Erlang packages from Alpine

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.3
+FROM operable/alpine:3.4
 
 # Setup Operable APK repository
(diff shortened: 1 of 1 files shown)


imbriaco:
To expand on the `operable/alpine:3.4` image, the following steps can be used to rebuild it using a fork of the official `docker-alpine` tooling from Gliderlabs:

```
git clone https://github.com/operable/docker-alpine
./build versions/operable-3.4/options
docker push operable/alpine
```

Issue #758, opened 2016-06-03
Executor crashes when failing to authorize against a rule mentioning multiple permissions


christophermaier:
When failing to authenticate against a rule that mentions multiple permissions, Cog will crash when rendering the error message. All proper authorization checks have taken place, but no output is returned to the user.

The offending code is at https://github.com/operable/cog/blob/a793edb656c6eedc28f60c560dead7e854b88e8c/lib/cog/error_response.ex#L15-L16

An example of a rule that will trigger this condition is:

    when command is twitter:tweet with option[as] == "support" must have twitter:tweet and site:support

If your user does not have both permissions, executing 

    twitter:tweet --as=support "blah blah blah"

 will crash the executor process.

The error rendering code was written assuming a rule would only mention a single permission. This specific failure is caused because the AST node in question is now a `Piper.Permissions.Ast.ConditionalExpr`, and it has no `:perms` field.

PR #764, opened 2016-06-09
Query rule for mentioned permissions

kevsmith:
This commit uses the newly [proposed](https://github.com/operable/piper/pull/20) `Permissions.Ast.Rule.permissions_used/1` function to obtain the list of permissions mentioned in a access rule to generate a better access denied error message. This also prevents Cog from crashing when rules mention more than one permission.

This isn't an ideal solution as we're not telling the user exactly which specific permissions they're lacking but it is a step in the right direction and vastly better than crashing.

50% fix for #758


vanstee:
LGTM :+1:

PR #769, opened 2016-06-14
Implement rule updating by disabling or deleting old rules

vanstee:


christophermaier:
👍

test/controllers/v1/rule_controller_test.exs
```diff
@@ -35,7 +35,8 @@ defmodule Cog.V1.RuleController.Test do
     id = body["id"]
 
     assert %{"id" => id,
-             "command" => "cog:s3",
+             "enabled" => true,
```
christophermaier: the enabled state shouldn't be  returned from the API; it's purely internal.

lib/cog/repository/rules.ex
```diff
@@ -50,7 +50,7 @@ defmodule Cog.Repository.Rules do
 
   def replace(bundle_version_id, id, rule_text) do
     rule_bundle_version = RuleBundleVersion
-    |> Repo.get_by(bundle_version_id: bundle_version_id, rule_id: id)
+    |> Repo.get_by!(bundle_version_id: bundle_version_id, rule_id: id)
```
christophermaier: We should probably pull this stuff into the transaction below.

vanstee: :+1:

lib/cog/repository/rules.ex
```diff
@@ -176,16 +176,7 @@ defmodule Cog.Repository.Rules do
...
-        # If we have disabled a bundle rule, but then add the same
-        # rule ourselves as a site rule, we need to ensure it's
-        # enabled here.
-        rule = if Cog.Repository.Bundles.is_site_version?(bundle_version) do
-          set_rule_status(rule, :enabled)
-        else
-          rule
-        end
-
+        set_rule_status(rule, :enabled)
```
christophermaier: No, we want to only do this if we're doing a site rule. Otherwise if you install a new version of a bundle that has a rule you've previously removed, this will resurrect it.

vanstee: This was backed out in the last commit. I didn't understand that until reading through the failing tests.

web/controllers/v1/rule_controller.ex
```diff
@@ -27,8 +27,15 @@ defmodule Cog.V1.RuleController do
     |> json(%{"errors" => "Unknown parameters #{inspect params}"})
   end
 
-  def create(conn, %{"rule" => rule_text}) do
-    case Rules.ingest(rule_text) do
+  def create(conn, %{"rule" => rule_text} = params) do
+    bundle_version = case params["bundle_version_id"] do
```
christophermaier: Users cannot create rules in bundles... they can only ever create "site" rules.

vanstee: This is also backed out in a later commit. Sorry for the mess.

web/controllers/v1/rule_controller.ex
```diff
@@ -42,6 +42,16 @@ defmodule Cog.V1.RuleController do
     end
   end
 
+  def show(conn, %{"id" => id}) do
+    case Rules.rule(id) do
```
christophermaier: This function implementation should be modified to only return a rule if it's currently enabled.

christophermaier: (where "this function" is `Rules.rule/1`)

vanstee: Ah it already does actually.

Issue #771, opened 2016-06-15
Add an `operable:trigger` command


christophermaier:
See https://github.com/operable/cog/issues/681

PR #775, opened 2016-06-16
Add `operable:trigger` command

christophermaier:
This brings the embedded bundle to parity with `cogctl` with respect
to trigger manipulation.

Fixes: #771

### DIFF
--- a/priv/repo/migrations/20160615225508_default_trigger_timeout.exs
+++ b/priv/repo/migrations/20160615225508_default_trigger_timeout.exs
@@ -1,0 +1,9 @@
+defmodule Cog.Repo.Migrations.DefaultTriggerTimeout do
+  use Ecto.Migration
+
+  def change do
+    alter table(:triggers) do
+      modify :timeout_sec, :integer, null: false, default: 30
+    end
+  end
+end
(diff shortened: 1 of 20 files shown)


priv/repo/migrations/20160615225508_default_trigger_timeout.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Cog.Repo.Migrations.DefaultTriggerTimeout do
+  use Ecto.Migration
+
+  def change do
+    alter table(:triggers) do
+      modify :timeout_sec, :integer, null: false, default: 30
```
kevsmith: How does this timeout differ from Cog's pipeline timeout?

christophermaier: Unrelated: this is how long a trigger will hang around waiting for the pipeline to complete before finalizing the HTTP response. This is already the default in the code, but we didn't actually have a non-null constraint here in the database; I added it here while I was in the neighborhood.

kevsmith: Aaah. Got it. Thanks for the clarification.

kevsmith:
👍

PR #778, opened 2016-06-17
add desc to relay groups

mpeck:


vanstee:
:+1:

If we have extra time it'd be nice to change the field to `description` to be consistent with other models.

mpeck:
It was already 'desc' in the db, so I just left it alone.

PR #781, opened 2016-06-17
Protect admin group memberships

vanstee:
We now protect the admin's membership to the cog-admin group via a db trigger. If deleted via the API we return a descriptive error message.

remove-admin

It turns out the admin user and cog-admin group are both already protected from name changes.


christophermaier:
👍

PR #785, opened 2016-06-20
Bring `operable:bundle` command to parity with cogctl

### LINKED ISSUE #776: Bring `operable:bundle` chat command to parity with `cogctl`
`cogctl` currently does:
     
     bundle
     bundle versions
     bundle info
     bundle install
     bundle uninstall
     bundle enable
     bundle disable
     
`operable:bundle` currently does:
     bundle status
     bundle enable
     bundle disable

We probably don't want to install bundles via chat (unless we download the config file remotely). That can come later.
     
`bundle status` can be folded into `bundle info`
     
`bundle` from cogctl should be `bundle list` in chat.

The rest of the functionality seems like a straightforward port.

christophermaier:
for the most part, anyway... we're not dealing with bundle install /
uninstall from chat right now.

`list`, `info`, and `versions` subcommands were added, so we can now
see what's actually installed. The `status` command was removed, since
`info` does its job now.

Existing subcommands were refactored to the subcommand-per-module
pattern.

Fixes #776

### DIFF
--- a/lib/cog/commands/bundle/info.ex
+++ b/lib/cog/commands/bundle/info.ex
@@ -1,0 +1,33 @@
+defmodule Cog.Commands.Bundle.Info do
+  alias Cog.Repository.Bundles
+  require Cog.Commands.Helpers, as: Helpers
+
+  Helpers.usage """
+  Show detailed information about a specific bundle.
+
+  USAGE
+    bundle info [FLAGS] <bundle>
+
+  ARGS
+    bundle   The name of a bundle
+
+  FLAGS
+    -h, --help  Display this usage info
+  """
+
+  def info(%{options: %{"help" => true}}, _args) do
+    show_usage
+  end
+  def info(_req, [bundle_name]) do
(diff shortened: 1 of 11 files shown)


lib/cog/commands/bundle/info.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  def info(%{options: %{"help" => true}}, _args) do
+    show_usage
+  end
+  def info(_req, [bundle_name]) do
+    case Bundles.bundle_by_name(bundle_name) do
+      nil ->
+        {:error, {:resource_not_found, "bundle", bundle_name}}
+      bundle ->
+        rendered = Cog.V1.BundlesView.render("show.json", %{bundle: bundle})
```
vanstee: This is super minor, but I think you can render the `"bundle.json"` template here instead to avoid the the extra `:bundle` key.

vanstee: Ah nevermind, I like this used along with the `"index.json"` below.

Issue #777, opened 2016-06-17
Cache the relay for a given bundle during pipeline execution


christophermaier:
Recent optimizations to Relay allow us to cache Docker containers in order to eliminate the startup overhead for long pipelines, or pipelines with many iterations. 

In order to maximize the effect of such optimizations, we should dispatch successive command invocations to a given bundle to the same Relay instance, ensuring we hit a "hot cache".

Currently, we select a Relay at random on each invocation, meaning we could theoretically incur container startup costs once per each Relay instance hosting a given bundle.

We can solve the problem by remembering which Relay is chosen at random for a given bundle in a pipeline, and then dispatching to that same Relay on successive invocations of commands in that bundle for the remainder of the pipeline (instead of choosing randomly each time).

PR #790, opened 2016-06-22
Randomly select Relay at beginning of stage only

kevsmith:
This commit modifies the Relay selection logic in Cog.Executor such that a Relay is randomly selected once at the start of each pipeline stage and is then reused for subsequent invocations for that stage. Doing so allows Cog to take maximum advantage of Relay's ability to reuse containers for the same pipeline stage and is expected to improve performance for Cog installations with multiple Relays.


mpeck:
:+1:

This looks reasonable to me, but I do have a question. If you are in the middle of a long running plan, and the bundle is unassigned from the relay, what happens? Would Cog just crash?

kevsmith:
The container wouldn't get torn down when the bundle is removed. Container destruction happens once it's idle. In the scenario you describe I'd expect the plan for the pipeline stage to finish without error (assuming the command invocations didn't throw any) but future invocations wouldn't select the same Relay(s).

christophermaier:
This isn't going to do what we want; an invocation can generate multiple plans, each of which is completely isolated from the others in its cohort.

We're going to need to remember the relay that's chosen at a higher level, probably in the executor state itself. I had a proposal for this in #777.

mpeck:
Ah, I see what your saying @christophermaier. Each time we `execute_plan` is called it gets a new plan. So with this change we would still be potentially selecting a new relay each time.

mpeck:
@kevsmith cool that makes sense. I asked more out of curiosity than anything, but is that something we should actually worry about? Should we check to see if relay is still valid on each execution and pick a new relay if it isn't?

PR #792, opened 2016-06-22
Cache and reuse relay assignments

### LINKED ISSUE #777: Cache the relay for a given bundle during pipeline execution

kevsmith:
This commit modifies Cog.Command.Pipeline.Executor to remember and reuse the Relay chosen to run an invocation for a bundle version. The memory persists between plans and stages allowing Cog to take maximum advantage of Relay's container reuse for a given pipeline.

The Executor also performs a safety check when it uses a cached Relay assignment. Before executing `current_plan`, the Executor double checks the selected Relay is still valid for the bundle version. If the Relay assignment is invalid (Relay is offline or the bundle version has been disabled/uninstalled) the Executor will attempt to select another Relay once before giving up and returning an error to the user.

Fixes #777

### DIFF
--- a/lib/cog/command/pipeline/executor.ex
+++ b/lib/cog/command/pipeline/executor.ex
@@ -10,6 +10,7 @@ defmodule Cog.Command.Pipeline.Executor do
   alias Cog.ErrorResponse
   alias Cog.Events.PipelineEvent
   alias Cog.Queries
+  alias Cog.Relay.Relays
   alias Cog.Repo
   alias Cog.ServiceEndpoint
   alias Cog.Template
@@ -56,6 +57,7 @@ defmodule Cog.Command.Pipeline.Executor do
     mq_conn: Carrier.Messaging.Connection.connection(),
     request: %Cog.Command.Request{}, # TODO: needs to be a type
     destinations: [Destination.t],
+    relays: Map.t,
     invocations: [%Piper.Command.Ast.Invocation{}], # TODO: needs to be a type
     current_plan: Cog.Command.Pipeline.Plan.t,
     plans: [Cog.Command.Pipeline.Plan.t],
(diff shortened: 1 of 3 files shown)


kevsmith:
I'm not sure I like the safety check as it stands. It will serialize pipeline execution to a degree since there's only one relay tracker process.

lib/cog/command/pipeline/executor.ex
```diff
@@ -653,6 +655,31 @@ defmodule Cog.Command.Pipeline.Executor do
...
+            {nil, relays}
+          # Store the selected relay in the relay cache
+          relay ->
+            {relay, Map.put(relays, bundle_name, relay)}
+        end
+      # Relay was previously assigned
+      relay ->
+        # Is the bundle still available on the relay? If not, remove the current assignment from the cache
+        # and select a new relay
+        if Relays.relay_available?(relay, bundle_name, bundle_version) == false do
```
christophermaier: Minor, but I'd probably flip this around to have the `true` arm  come first, simplifying the conditional.

christophermaier:
👍 , modulo a small style comment

christophermaier:
@kevsmith Not sure it serializes it any worse than it was before... we were still hitting the Tracker to get a relay at random.

christophermaier:
:shipit:

kevsmith:
True. And also, meh.

PR #798, opened 2016-06-23
Update the role command to use arguments, not options

### LINKED ISSUE #678: Redesign `role` command
`role` is similar to `permissions`. We'll need to swap long args for subcommands. Also I'm going to rename `role` to `roles` to be consistent with other command names.

```
USAGE

roles <subcommand>

roles list
roles create <role-name>
roles drop <role-name>
roles grant <role-name> <group-name ...>
roles revoke <role-name> <group-name ...>

EXAMPLES

roles list
> NAME  GROUPS
  ops   cog flywheel
  ux    flywheel

roles create ops
> Created role ops

roles drop ops
> Deleted role ops

roles grant ops cog
> Granted role ops to group cog

roles grant ops cog flywheel
> Granted role ops to groups cog flywheel

roles revoke ops cog
> Revoked role ops from group cog

roles revoke ops cog flywheel
> Revoked role ops from cog flywheel

christophermaier:
Fixes #678


kevsmith:
👍

PR #802, opened 2016-06-23
Improve command prefix handling

imbriaco:
* Allow command prefix to be configured via `$COG_COMMAND_PREFIX`
* `Regex.escape` the prefix when we strip it from spoken commands

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -53,7 +53,7 @@ config :httpotion, :default_timeout, 30000
 # ========================================================================
 # Commands, Bundles, and Services
 
-config :cog, :command_prefix, "!"
+config :cog, :command_prefix, System.get_env("COG_COMMAND_PREFIX") || "!"
 
 config :cog, Cog.Bundle.BundleSup,
   bundle_root: Path.join([File.cwd!, "bundles"])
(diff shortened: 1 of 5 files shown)


imbriaco:
To provide some more commentary on the Regex part. I tried setting the prefix to `^` and hilarity ensued:

slack 2016-06-23 17-46-06

vanstee:
:+1: LGTM

Might be nice to do all the escaping in one place, but this is fine for now.

vanstee:
Oh I see why we can't do that; we're doing a `String.starts_with?` compare first.

mpeck:
:+1:

vanstee:
Oh wait. Do we need to escape this as well? https://github.com/operable/cog/blob/master/lib/cog/command/pipeline/executor.ex#L650

mpeck:
Good catch @vanstee. I think you might be right.

imbriaco:
Yeah, I added an escape in the executor to catch that case too. Good find. Ultimately, I think we should do what the HipChat adapter does and compile a regex when the adapter initializes that we cache in the process. I will write an issue for that and we can come back to it, but this should help mitigate the immediate issue.

PR #804, opened 2016-06-24
allow self updates takes a list of actions

mpeck:
Since allow_self_updates was hardcoded to only work with :update and :show, chat_handle :upserts would fail if the user didn't have the `manage_users` permission. This fixes the problem by allowing a list of actions to be passed to `allow_self_updates`. We also now have a test to make sure that even users without `mange_users` permission can update their own chat_handle.

This partially resolves https://github.com/operable/flywheel/issues/362

### DIFF
--- a/test/controllers/v1/chat_handle_controller_test.exs
+++ b/test/controllers/v1/chat_handle_controller_test.exs
@@ -49,6 +49,12 @@ defmodule Cog.V1.ChatHandleControllerTest do
     end
   end
 
+  test "users can update their own chat handles even if they aren't authorized", params do
+    conn = api_request(params.unauthed, :post, "/v1/users/#{params.unauthed.id}/chat_handles", body: %{"chat_handle" => @valid_attrs})
+    id = json_response(conn, 201)["chat_handle"]["id"]
+    assert Repo.get_by(ChatHandle, id: id)
+  end
+
   test "creates and renders resource when data is valid", params do
     conn = api_request(params.authed, :post, "/v1/users/#{params.authed.id}/chat_handles", body: %{"chat_handle" => @valid_attrs})
     id = json_response(conn, 201)["chat_handle"]["id"]
--- a/web/plugs/authorization.ex
+++ b/web/plugs/authorization.ex
@@ -42,11 +42,13 @@ defmodule Cog.Plug.Authorization do
       raise ":permission key must be a string, but was '#{inspect permission}' instead"
     end
     {_,_} = Permission.split_name(permission) # error if can't be split
-    self_updates = Keyword.get(opts, :allow_self_updates, [])
-    unless is_list(self_updates) do
-      raise ":allow_self_updates must be a list, but was '#{inspect self_updates}' instead"
+    self_updates = Keyword.get(opts, :self_updates_on, [])
+    |> List.wrap
+
+    unless Enum.all?(self_updates, &is_atom/1) do
+      raise ":allow_self_updates must be a list of atoms, but was '#{inspect self_updates}' instead"
     end
-    opts
+    Keyword.put(opts, :self_updates_on, self_updates)
   end
 
   # Expects to be called after the `Cog.Plug.Authentication` plug
@@ -75,7 +77,7 @@ defmodule Cog.Plug.Authorization do
   # in the future. For example, this would allow users to update
   # related data with paths such as `/v1/users/:user_id/profiles/:id`.
   defp self_updating?(opts, conn) do
-    case Keyword.get(opts, :allow_self_updates, []) do
+    case Keyword.get(opts, :self_updates_on) do
       [] -> false
       actions ->
         conn.private.phoenix_action in actions and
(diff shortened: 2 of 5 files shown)


web/plugs/authorization.ex
```diff
@@ -42,9 +42,9 @@ defmodule Cog.Plug.Authorization do
       raise ":permission key must be a string, but was '#{inspect permission}' instead"
     end
     {_,_} = Permission.split_name(permission) # error if can't be split
-    self_updates = Keyword.get(opts, :allow_self_updates, false)
-    unless self_updates == true || self_updates == false do
-      raise ":allow_self_updates must be a boolean, but was '#{inspect self_updates}' instead"
+    self_updates = Keyword.get(opts, :allow_self_updates, [])
+    unless is_list(self_updates) do
```
christophermaier: Might want to expand the check to ensure that all members of the list are atoms.

web/plugs/authorization.ex
```diff
@@ -75,12 +75,12 @@ defmodule Cog.Plug.Authorization do
...
   defp self_updating?(opts, conn) do
-    if Keyword.get(opts, :allow_self_updates, false) == true do
-      conn.private.phoenix_action in [:update, :show] and
-      (conn.assigns.user.id == conn.params["id"] or
-       conn.params["id"] == "me")
-    else
-      false
+    case Keyword.get(opts, :allow_self_updates, []) do
+      [] -> false
+      actions ->
```
christophermaier: This all might read easier if we changed it from `:allow_self_updates` to something like `:self_updating_actions`. The former "sounds" Boolean, while the latter is more suggestive of a list.

web/plugs/authorization.ex
```diff
@@ -75,12 +75,12 @@ defmodule Cog.Plug.Authorization do
...
+        conn.private.phoenix_action in actions and
+        (conn.assigns.user.id == conn.params["id"] or
+         conn.params["id"] == "me")
```
christophermaier: I realize this didn't change in this PR, but where do we ever have an `id` of `me`?

web/plugs/authorization.ex
```diff
@@ -75,12 +75,12 @@ defmodule Cog.Plug.Authorization do
   # in the future. For example, this would allow users to update
   # related data with paths such as `/v1/users/:user_id/profiles/:id`.
   defp self_updating?(opts, conn) do
-    if Keyword.get(opts, :allow_self_updates, false) == true do
-      conn.private.phoenix_action in [:update, :show] and
-      (conn.assigns.user.id == conn.params["id"] or
-       conn.params["id"] == "me")
-    else
-      false
+    case Keyword.get(opts, :allow_self_updates, []) do
```
christophermaier: This can just be `Keyword.fetch!(opts, :allow_self_updates)` if you update the value of  `:allow_self_updates` in `opts` to `self_updates` in `init` before returning instead of just `opts`.

test/controllers/v1/chat_handle_controller_test.exs
```diff
@@ -49,6 +49,12 @@ defmodule Cog.V1.ChatHandleControllerTest do
     end
   end
 
+  test "users can update their own chat handles even if they aren't authorized", params do
+    conn = api_request(params.unauthed, :post, "/v1/users/#{params.unauthed.id}/chat_handles", body: %{"chat_handle" => @valid_attrs})
+    id = json_response(conn, 201)["chat_handle"]["id"]
+    assert Repo.get_by(ChatHandle, id: id)
+  end
+
```
christophermaier: yay tests!

web/plugs/authorization.ex
mpeck: Ha, I didn't even notice that. I guess thats what I get for copy and pasting :|

I actually have no idea where that would come in to play. I feel like it probably has something to do with Flywheel, but I'm a bit leery to remove it at this point.

web/plugs/authorization.ex
mpeck: How about, `self_updates_on`?

web/plugs/authorization.ex
```diff
@@ -42,11 +42,13 @@ defmodule Cog.Plug.Authorization do
...
     end
     {_,_} = Permission.split_name(permission) # error if can't be split
-    self_updates = Keyword.get(opts, :allow_self_updates, [])
-    unless is_list(self_updates) do
-      raise ":allow_self_updates must be a list, but was '#{inspect self_updates}' instead"
+    self_updates = Keyword.get(opts, :self_updates_on, [])
+    |> List.wrap
+
+    unless Enum.all?(self_updates, &is_atom/1) do
+      raise ":allow_self_updates must be a list of atoms, but was '#{inspect self_updates}' instead"
```
christophermaier: "atom or list of atoms", given that you're doing `List.wrap`

web/plugs/authorization.ex
```diff
@@ -75,7 +77,7 @@ defmodule Cog.Plug.Authorization do
   # in the future. For example, this would allow users to update
   # related data with paths such as `/v1/users/:user_id/profiles/:id`.
   defp self_updating?(opts, conn) do
-    case Keyword.get(opts, :allow_self_updates, []) do
+    case Keyword.get(opts, :self_updates_on) do
```
christophermaier: `Keyword.fetch!` would be better here

christophermaier:
Might be worth a test on the plug itself to verify the self-update functionality.

Other than that, 🚢

PR #805, opened 2016-06-24
Remove old commands

christophermaier:
Removes a handful of old commands. Reasons for some of the less-obvious ones are in the commit messages; if we want to keep any of them, the corresponding commits can easily be removed from this PR prior to merge.


vanstee:
LGTM. I think there might be some places that we use `operable:table` in the docs. Might want to suggest people install the `format` bundle during bootstrapping.

imbriaco:
:+1:

PR #806, opened 2016-06-24
Rename alias subcommands

christophermaier:
This will make them consistent with the rest of the subcommands in the
embedded bundle.

`alias new` -> `alias create`
`alias mv`  -> `alias move`
`alias rm`  -> `alias delete`
`alias ls`  -> `alias list`

Also, `alias list` is the default subcommand


lib/cog/commands/alias/move.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Cog.Commands.Alias.Mv do
+defmodule Cog.Commands.Alias.Move do
```
vanstee: This feels a little bit more like a rename, but I don't really feel that strongly about it.

vanstee:
LGTM

PR #810, opened 2016-06-24
Add a `relay-group list` subcommand

christophermaier:
Turns out the `relay-group info` command behaved like this when given 0
arguments; this basically pulls that out into a `list` command,
consistent with other commands.

This new `list` subcommand is not the default if a subcommand is not specified.

(While I was here, I fixed the `relay-group-info-verbose` template,
which emitted a `Bundles:` line for each bundle, prior to listing the
names of all bundles at the end)


lib/cog/commands/relay_group.ex
```diff
@@ -50,7 +53,7 @@ defmodule Cog.Commands.RelayGroup do
         if Helpers.flag?(req.options, "help") do
           show_usage
         else
-          show_usage(error(:required_subcommand))
+          RelayGroup.List.list(req, args)
```
kevsmith: Very minor style nit -- `RelayGroup.List.list` is repetitive. Maybe `RelayGroup.List.run`?

kevsmith:
👍

lib/cog/commands/relay_group.ex
christophermaier: I think I'm going to go back through all these subcommands and convert them all over to calling their main function `run` as a separate PR.

PR #811, opened 2016-06-24
Automatically bootstrap using environment variables, if present.

imbriaco:
Create an administrative user using values from the environment, if present. If any of the following variables are not set, the bootstrap will not be performed. Required variables:

```
COG_BOOTSTRAP_USERNAME
COG_BOOTSTRAP_PASSWORD
COG_BOOTSTRAP_EMAIL_ADDRESS
COG_BOOTSTRAP_FIRST_NAME
COG_BOOTSTRAP_LAST_NAME
```

Additionally, if the user bootstrap completes successfully and the `RELAY_ID` and `RELAY_COG_TOKEN` variables are set, a relay will be created using the variables and assigned to a default relay group, which will also be created.  If the relevant variables are not present, this step will be skipped.

If any creation attempt fails, the entire bootstrap process will be rolled back.

### DIFF
--- a/lib/cog/bootstrap.ex
+++ b/lib/cog/bootstrap.ex
@@ -3,16 +3,31 @@ defmodule Cog.Bootstrap do
   Support functions for bootstrapping a Cog system.
   """
 
+  require Logger
+
   alias Cog.Models.Group
   alias Cog.Models.Role
   alias Cog.Models.User
   alias Cog.Repo
+  alias Cog.Repository.Relays, as: RelaysRepo
+  alias Cog.Repository.RelayGroups, as: RelayGroupsRepo
 
   @default_admin_params %{
     "username"      => "admin",
     "first_name"    => "Cog",
     "last_name"     => "Administrator",
     "email_address" => "cog@localhost"
+  }
+
+  @default_relay_params %{
+    "name"        => "Default",
+    "description" => "Default relay",
+    "enabled"     => true
+  }
+
+  @default_relay_group_params %{
+    "name"        => "Default",
+    "description" => "Default relay group"
   }
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/cog/bootstrap.ex
```diff
@@ -54,6 +67,65 @@ defmodule Cog.Bootstrap do
...
+  def maybe_bootstrap do
+    unless is_bootstrapped? do
+      Repo.transaction fn ->
+        if bootstrap_from_env, do: relay_from_env(System.get_env)
+      end
+
+      true
+    end
+
+    false
```
christophermaier: Why this return value? And the `true` above in the `unless` block?

lib/cog/bootstrap.ex
```diff
@@ -54,6 +67,65 @@ defmodule Cog.Bootstrap do
...
+  end
+
+  def bootstrap_from_env do
```
christophermaier: Can be private

lib/cog/bootstrap.ex
```diff
@@ -54,6 +67,65 @@ defmodule Cog.Bootstrap do
...
+      key = "COG_BOOTSTRAP_" <> String.upcase(param)
+      { param, System.get_env(key) }
+    end |> Enum.into(%{})
+
+    unless Enum.any?(user_map, fn({_k, v}) -> v == nil end) do
+      bootstrap(user_map)
+    end
+  end
+
+  def relay_from_env(%{"RELAY_ID" => relay_id,
```
christophermaier: Can be private

lib/cog/bootstrap.ex
```diff
@@ -54,6 +67,65 @@ defmodule Cog.Bootstrap do
...
+  and RELAY_COG_TOKEN variables are set, a relay will be created using the
+  variables and assigned to a default relay group, which will also be created.
+  If the relevant variables are not present, this step will be skipped.
+
+  If any creation attempt fails, the entire bootstrap process will be rolled
+  back.
+  """
+  def maybe_bootstrap do
+    unless is_bootstrapped? do
+      Repo.transaction fn ->
```
christophermaier: Bootstrapping already takes place within a transaction; this one isn't needed.

lib/cog/bootstrap.ex
```diff
@@ -54,6 +67,65 @@ defmodule Cog.Bootstrap do
...
+  end
+
+  def bootstrap_from_env do
+    user_keys = ~w(username password email_address first_name last_name)
+    user_map = for param <- user_keys do
+      key = "COG_BOOTSTRAP_" <> String.upcase(param)
+      { param, System.get_env(key) }
+    end |> Enum.into(%{})
+
+    unless Enum.any?(user_map, fn({_k, v}) -> v == nil end) do
```
christophermaier: Could be useful to emit a warning log message if any of the variables are unset.

lib/cog/bootstrap.ex
christophermaier: Ah, I see... `relay_from_env` is pulled into the transaction here.

christophermaier:
Have you tested to see if the transactions behave as expected? If you're not explicitly checking return values and rolling back the transaction on your own, you need to be using `!` methods and throwing exceptions to abort the transaction. In the past, I wasn't doing that and would happily carry on with the transaction when I thought I wasn't.

lib/cog/bootstrap.ex
imbriaco: I ... don't remember. Looking at the code now it doesn't seem to be used. I think I just wanted to indicate whether a bootstrap actually  happened or not as a result of that call. I can easily take taht out if we want.

imbriaco:
@christophermaier I made the functions you mentioned private and left a comment above about he return value from the `maybe_bootstrap`. i don't have a strong opinion about whether or not to keep it.

As far as the transaction goes, yes, I tested it. If you look in `relay_from_env` I'm doing `:ok =

lib/cog/bootstrap.ex
vanstee: Yeah let's take that out. Feels weird that it would always return `false`.

lib/cog/bootstrap.ex
vanstee: This might be a tiny bit more readable `user_map |> Map.values |> Enum.any?(&is_nil/1)`

christophermaier:
Ah, yeah, I missed the `:ok = with ...` bit.

I'd rather remove those booleans... it's potentially confusing.

imbriaco:
Added logging for missing variables. I only emit this if at least one of the `COG_BOOTSTRAP_*` variables was seen so we're not noisy if the user didn't intend to bootstrap from the environment.

```
2016-06-24T18:22:18.0424  (Cog.Bootstrap:116) [info] Skipping bootstrap: Missing value for COG_BOOTSTRAP_LAST_NAME.
```

Also removed the booleans.

christophermaier:
👍

imbriaco:
Thanks, @christophermaier for the review. You made my code better. I'm going to go ahead and merge as soon as CI finishes since I've addressed all the feedback.

Here's where we landed:

### Victory!

Environment:
```
export COG_BOOTSTRAP_USERNAME=imbriaco
export COG_BOOTSTRAP_PASSWORD=sekrit
export COG_BOOTSTRAP_FIRST_NAME=Mark
export COG_BOOTSTRAP_LAST_NAME=Imbriaco
export COG_BOOTSTRAP_EMAIL_ADDRESS=mark@operable.io
export RELAY_ID=8b610fc6-f643-4e0b-bba0-bfc140266393
export RELAY_COG_TOKEN=supersecret
```

All variables present:

```
2016-06-24T19:46:39.0428  (Cog.Bootstrap:112) [info] Bootstrapping Cog from environment variables.
2016-06-24T19:46:40.0595  (Cog.Bootstrap:130) [info] Configuring default relay from environment variables.
```

Validation:
```
 $ cogctl users --rest-user=imbriaco --rest-password=sekrit
USERNAME  FULL NAME      EMAIL_ADDRESS
imbriaco  Mark Imbriaco  mark@operable.io
```

### Error Logging:

Cog variables present, but no relay variables:

```
2016-06-24T19:42:18.0428  (Cog.Bootstrap:112) [info] Bootstrapping Cog from environment variables.
```

Some, but not all, Cog variables present:

```
2016-06-24T19:49:46.0210  (Cog.Bootstrap:119) [info] Skipping bootstrap: Missing value for COG_BOOTSTRAP_FIRST_NAME.
2016-06-24T19:49:46.0210  (Cog.Bootstrap:119) [info] Skipping bootstrap: Missing value for COG_BOOTSTRAP_USERNAME.
```

PR #812, opened 2016-06-24
Adds support for centrally managing bundle dynamic configs

kevsmith:
The work is essentially feature complete and ready for feedback. Test coverage is lacking but I'll address that shortly.


lib/cog/models/bundle_dynamic_config.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Cog.Models.BundleDynamicConfig do
+  use Cog.Model, :no_primary_key
```
christophermaier: Why no primary key?

lib/cog/models/bundle_dynamic_config.ex
```diff
@@ -0,0 +1,37 @@
...
+
+  end
+
+  @required_fields ~w(bundle_id config)
+  @optional_fields ~w()
+
+  def changeset(model, params \\ :empty) do
+    model
+    |> cast(params, @required_fields, @optional_fields)
+    |> unique_constraint(:bundle_id, name: :bundle_dynamic_configs_bundle_id_index)
```
christophermaier: Sounds like this is the primary key?

christophermaier: If the bundle ID is the primary key, would this be better as a field on the bundle table?

lib/cog/models/bundle_dynamic_config.ex
```diff
@@ -0,0 +1,37 @@
...
+    |> calculate_hash
+  end
+
+  defp calculate_hash(changeset) do
+    case fetch_change(changeset, :config) do
+      :error ->
+        changeset
+      {:ok, config} ->
+        json = Poison.encode!(config)
+        hash = :crypto.hash(:sha256, json) |> Base.encode16 |> String.downcase
```
christophermaier: Is it possible to hash the data-structure itself or normalize it instead? The JSON string representation won't be deterministic, so the hash could change.

lib/cog/relay/info.ex
```diff
@@ -54,28 +58,75 @@ defmodule Cog.Relay.Info do
...
+  end
+  defp calculate_signature(configs) do
+    configs
+    |> Enum.sort(&(&1.bundle.name <= &2.bundle.name))
+    |> Enum.map(&(&1.hash))
+    |> Enum.join("\n")
+    |> hash_string
+  end
+
+  defp hash_string(text), do: :crypto.hash(:sha256, text) |> Base.encode16 |> String.downcase
```
christophermaier: would be good to extract the hashing logic and re-use it here and in the model.

lib/cog/models/bundle_dynamic_config.ex
christophermaier: Actually, never mind... keep 'em separate.

lib/cog/util/ecto.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Cog.Util.Ecto do
+
+  require Logger
+
+  @doc """
+  Sanely logs Ecto query output
+  """
+  def log(entry) do
+    IO.puts("here")
```
christophermaier: leftovers?

lib/cog/util/ecto.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Cog.Util.Ecto do
```
christophermaier: How / where does this get used?

lib/cog/models/bundle_dynamic_config.ex
christophermaier: Actually, just ignore all these comments

lib/cog/util/ecto.ex
kevsmith: Argh. Yeah. That's from a failed experiment to tame Ecto's verbose logging. Sadly, it only works with 2.0 :(

lib/cog/models/bundle_dynamic_config.ex
kevsmith: For sure. I'd noticed that Poison appears to have stable output so I used that for the first pass. The hash is used as an ETag and is never computed anywhere but here, fwiw.

christophermaier:
💯 ❤️  the stable hash stuff

PR #815, opened 2016-06-25
preload relays and bundle versions on relay group update

mpeck:


web/controllers/v1/relay_group_controller.ex
```diff
@@ -42,7 +42,7 @@ defmodule Cog.V1.RelayGroupController do
   end
 
   def update(conn, %{"id" => id, "relay_group" => group_params}) do
-    relay_group = Repo.get!(RelayGroup, id)
+    relay_group = Repo.one(Queries.RelayGroup.for_id(id))
```
christophermaier: It'd be good to use `Cog.Repository.RelayGroups.by_id/1` for this instead. 

(Bonus points for switching the whole function over to using repository calls, but any little bit helps).

christophermaier:
😍

kevsmith:
👍

PR #818, opened 2016-06-25
Unrecognized subcommands are failures

christophermaier:
Previously, a few commands would actually display an error message,
but not create a formal error, which would allow a pipeline to
continue.

Now, each command behaves similarly, and appropriately.


vanstee:
:+1:

PR #820, opened 2016-06-25
Squat on certain bundle names

### LINKED ISSUE #817: Squat on certain bundle names
We should prevent the installation of any bundles named `"cog"`, `"cog-*"`, or `"operable-*"` to prevent confusion between "official" bundles from Operable itself and 3rd party bundles.

Should we decide to create any bundles with those names in the future, we can revisit the blocking logic.

christophermaier:
To avoid confusion between "official" Operable-authored bundles and 3rd
party bundles, we'll block the installation of bundles named

* `cog`
* `cog-*`
* `operable-*`

The bundle names `site` and `user` continue to be reserved for internal
usage. The embedded bundle continues to be named "operable", and
continues to be uninstallable by outside sources.

We were using "cog" as a bundle name for testing data; obviously that's
out now, so we'll now call it "test-bundle" instead.

Fixes #817

### DIFF
--- a/lib/cog/repository/bundles.ex
+++ b/lib/cog/repository/bundles.ex
@@ -17,7 +17,8 @@ defmodule Cog.Repository.Bundles do
   @reserved_bundle_names [
     Cog.embedded_bundle,
     Cog.site_namespace,
-    "user" # a bundle named "user" would break alias resolution
+    "user", # a bundle named "user" would break alias resolution
+    "cog"   # we're going to squat on this for now to prevent potential confusion
   ]
 
   @permanent_site_bundle_version "0.0.0"
--- a/lib/cog/support/model_utilities.ex
+++ b/lib/cog/support/model_utilities.ex
@@ -192,7 +192,7 @@ defmodule Cog.Support.ModelUtilities do
   Create a command with the given name
   """
   def command(name) do
-    bundle_version = bundle_version("cog")
+    bundle_version = bundle_version("test-bundle")
     bundle = bundle_version.bundle
 
     command = %Command{}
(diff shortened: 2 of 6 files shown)


lib/cog/repository/bundles.ex
```diff
@@ -17,9 +17,14 @@ defmodule Cog.Repository.Bundles do
...
     Cog.embedded_bundle,
     Cog.site_namespace,
-    "user" # a bundle named "user" would break alias resolution
+    "user", # a bundle named "user" would break alias resolution
+    "cog"   # we're going to squat on this for now to prevent potential confusion
   ]
 
+  # We're also going to block the installation of any bundles that are
+  # prefixed with "cog" or the embedded bundle name
+  @reserved_bundle_name_regex Regex.compile!("^(cog|#{Cog.embedded_bundle})-.*$")
```
mpeck: I'm not sure how locked down we need to be, but couldn't you create a bundle called `cog_foo` or `cogfoo` with this regex?

mpeck:
Looks good, just the one question on how strict we want to be with the regex.

lib/cog/repository/bundles.ex
christophermaier: You can indeed (and there are tests proving that). This just reflects the discussions we were having in chat. We could easily add other delimiters besides `-`, and possibly just block any string that starts with "operable", since I can't think of much else that legitimately starts that way. I don't know about blocking everything that starts with "cog", though, since there are a number of legitimate words that begin that way.

@imbriaco @kevsmith, do you have any thoughts on this?

kevsmith: I think handling the names "cog" and "operable" are good enough for now. We can change this in a future release if needed.

christophermaier:
Slight change of plans: we're only going to squat on the "cog" name itself for now.

PR #828, opened 2016-06-29
Change pipeline execution to per-stage

kevsmith:
This commit does two things:

1. Changes pipeline execution to a per-stage model. This means Cog sends
an entire stage down to a target Relay and, in turn, receives an entire
stage of output when the Relay completes execution.

2. Adds snappy compression to all bus traffic. This is meant to address
the increased message sizes caused by executing an entire stage at a time.

### DIFF
--- a/lib/cog/adapter.ex
+++ b/lib/cog/adapter.ex
@@ -36,8 +36,9 @@ defmodule Cog.Adapter do
         {:reply, :ok, state}
       end
 
-      def handle_info({:publish, topic, message}, state) do
+      def handle_info({:publish, topic, compressed}, state) do
         if topic == reply_topic() do
+          {:ok, message} = :snappy.decompress(compressed)
           payload = Poison.decode!(message)
           send_message(payload["room"], payload["response"])
         end
--- a/lib/cog/command/request.ex
+++ b/lib/cog/command/request.ex
@@ -1,81 +1,21 @@
 defmodule Cog.Command.Request do
 
   use Cog.Marshalled
-  alias Spanner.Config
 
   require Logger
 
-  defmarshalled [:room, :requestor, :user, :command, :args, :options,
-                 :command_config, :reply_to, :cog_env, :invocation_id,
-                 :invocation_step, :service_token, :services_root]
+  # defmarshalled [:room, :requestor, :user, :command, :args, :options,
+  #                :command_config, :reply_to, :cog_env, :invocation_id,
+  #                :invocation_step, :service_token, :services_root]
+
+  defmarshalled [:options, :args, :cog_env, :invocation_step]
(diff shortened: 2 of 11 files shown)


lib/cog/adapter.ex
```diff
@@ -36,8 +36,9 @@ defmodule Cog.Adapter do
         {:reply, :ok, state}
       end
 
-      def handle_info({:publish, topic, message}, state) do
+      def handle_info({:publish, topic, compressed}, state) do
```
vanstee: Good call on the param name change here.

lib/cog/command/request.ex
```diff
@@ -1,81 +1,21 @@
...
-  alias Spanner.Config
 
   require Logger
 
-  defmarshalled [:room, :requestor, :user, :command, :args, :options,
-                 :command_config, :reply_to, :cog_env, :invocation_id,
-                 :invocation_step, :service_token, :services_root]
+  # defmarshalled [:room, :requestor, :user, :command, :args, :options,
+  #                :command_config, :reply_to, :cog_env, :invocation_id,
+  #                :invocation_step, :service_token, :services_root]
```
vanstee: Should we go ahead and nuke this?

lib/cog/command/request.ex
```diff
@@ -1,81 +1,21 @@
...
-            # If multiple configs are found we should warn the user
-            with {:ok, config} <- Config.Parser.read_from_file(config_path) do
-              {:multiple_configs, config, config_path}
-            end
-        end
-      false -> :dir_not_found
+      request.invocation_step == nil ->
+        {:error, {:empty_field, :invocation_step}}
+      request.cog_env == nil ->
+        {:error, {:empty_field, :cog_env}}
```
vanstee: Looks like we can now switch this from a `cond` to a `case`:

```elixir
case request do
  %__MODULE__{invocation_step: nil} ->
    {:error, {:empty_field, :invocation_step}}
  %__MODULE__{cog_env: nil} ->
    {:error, {:empty_field, :cog_env}}
  request ->
    request
end
```

Also it looks like there's no `true` clause here which seems weird. Are we always expecting an error?

mix.exs
```diff
@@ -78,6 +78,7 @@ defmodule Cog.Mixfile do
      {:fumanchu, github: "operable/fumanchu"},
      {:exirc, "~> 0.9.2"},
      {:exjsx, "~> 3.2", override: true},
+     {:snappy, github: "fdmanana/snappy-erlang-nif", ref: "76417d16a8d39e04d66114d8f8a3c75df19474f0"},
```
vanstee: Do we need to lock to a specific ref?

PR #830, opened 2016-06-29
Ecto 2.0

vanstee:
Warnings will be removed in a future PR


lib/cog/repository/bundles.ex
```diff
@@ -464,7 +464,7 @@ defmodule Cog.Repository.Bundles do
     from bv in BundleVersion,
     join: b in assoc(bv, :bundle),
     where: b.name == ^site,
-    where: bv.version == @permanent_site_bundle_version,
+    where: bv.version == ^@permanent_site_bundle_version,
```
christophermaier: Was this actually causing a problem?

vanstee: Yes, it failed to compile.

christophermaier: That's interesting... I'd thought the compiler would see that the same as putting a string literal in there.

In any event, good find 👍

christophermaier:
Good stuff! :shipit:

lib/cog/repository/bundles.ex
vanstee: Yep surprised me too.

PR #831, opened 2016-06-29
Fix compilation issues and warnings on Elixir 1.3

vanstee:


kevsmith:
I want to hold off on merging this and #832 until I branch for 0.9 this morning.

vanstee:
@kevsmith Let me try to rebase https://github.com/operable/cog/pull/832 onto master then so we can get that merged. It'd be nice to get rid of all the Ecto warnings before cutting a release.

kevsmith:
Ok. Lemme know when that's ready and I'll pull it into the 0.9 branch I just made.

vanstee:
Actually. This should still run just fine on 1.2, since I haven't bumped requirements or anything. Any reason not to include it?

kevsmith:
I saw some of the deps had changed and was concerned that might introduce last minute bugs. If we're confident that's not the case then I'm ok w/merging.

vanstee:
Good point. I'm not super confident about the dep updates. Most we're minor updates to fix 1.3 compilation issues, but I haven't looked into them too much.

kevsmith:
LGTM now that we've released 0.9.

PR #832, opened 2016-06-29
Fix remaining ecto warnings

vanstee:
Depends on https://github.com/operable/cog/pull/831 but I can rebase if we don't want to merge that.


vanstee:
Rebased onto master.

PR #840, opened 2016-07-06
Enable user self-registration

### LINKED ISSUE #277: Support automatic Cog account creation
Some installations do not want or need restrictions around who can interact with Cog. For these users, the manual step of creating a Cog account for new users is an unnecessary manual step.

To address this need while still continuing to support restricted accounts I propose the following:

1. Teach Cog to create a user account the first time it interacts with a user. I suspect we can get most, if not all, of the required data from the chat provider's APIs. If not, we could add a short wizard-like interaction where Cog asks the user for the missing info.
1. Add a config flag which toggles between restricted accounts and the behavior described above.

kevsmith:
This PR adds support for creating new users upon their first interaction with Cog. When an unknown user sends a request to Cog it will create a corresponding user and chat handle, send an informational message to the user informing them their account has been created, and then allowing their request to proceed.

Note that a user account is created but the user isn't placed into any groups nor granted any permissions. It remains up to the site Cog admin to grant appropriate permissions to the new account.

This behavior is controlled by a new `:cog` config entry named `:self_registration` which defaults to false (disabled). The feature can be enable either by setting the config value to true or setting the environment variable `$COG_ALLOW_SELF_REGISTRATION` to any non-nil value.

The current design handles two anticipated failure cases:

1. A Cog username corresponding to the user's chat handle is already in use. In this case Cog will attempt to generate a new username in the general form of `chat_handle_n` where `chat_handle` is the requesting user's chat handle and `n` is an incrementing integer starting at 1. Cog will make 20 tries at finding a free username before giving up. For example, if the username 'bob' is taken Cog will try `bob_1` thru `bob_20` before giving up.

1. A database error prevents the users chat handle from being assigned to their newly created Cog account.

When Cog hits one of these error conditions it sends an error message to the requesting user informing them of the error and directing them to contact a Cog admin to investigate the problem.

Fixes #277


lib/cog/command/pipeline/initializer.ex
```diff
@@ -30,11 +34,22 @@ defmodule Cog.Command.Pipeline.Initializer do
...
-      {true, payload, state} ->
-        {:ok, _} = ExecutorSup.run(payload)
-        {:noreply, state}
-      {false, _, state} ->
+    # Only self register when the feature is enabled via config
+    # and the incoming request is from a real user. We use
+    # the HTTP adapter as a proxy for the second condition as
+    # the only requests currently coming from that adapter are
+    # triggers.
+    self_register_flag = Application.get_env(:cog, :self_registration, false) and payload["adapter"] != "http"
```
christophermaier: We should use `Cog.Adapter.chat_adapter?/0` here instead of just relying on the adapter name; presumably self registration should only ever happen in the context of chat.

christophermaier: Also, doesn't the config value already default to `false`?

lib/cog/adapters/slack/api.ex
```diff
@@ -276,8 +276,9 @@ defmodule Cog.Adapters.Slack.API do
   defp cache_item(%{"is_group" => true, "id" => id, "name" => name}),
     do: %{id: id, name: name}
   # users have names, but not "is_user" fields :( they do have profiles, though
-  defp cache_item(%{"id" => id, "name" => name, "profile" => %{}}),
-    do: %{id: id, handle: name}
+  defp cache_item(%{"id" => id, "name" => name, "profile" => profile}),
+  do: %{id: id, handle: name, email: profile["email"], first_name: profile["first_name"],
+        last_name: profile["last_name"]}
```
christophermaier: This is fine for now since we're focusing on Slack, but this feature won't work in HipChat or IRC until those adapters are adjusted.

We should make issues tracking both of those so we don't lose track.

config/config.exs
```diff
@@ -64,6 +64,9 @@ config :cog, :command_rule_ttl, {10, :sec}
 config :cog, :template_cache_ttl, {60, :sec}
 config :cog, :user_perms_ttl, {10, :sec}
 
+# Enable/disable user self-registration
+config :cog, :self_registration, System.get_env("COG_ALLOW_SELF_REGISTRATION") != nil || false
```
christophermaier: We'll need to add an entry to the documentation site for this.

lib/cog/command/pipeline/initializer.ex
```diff
@@ -68,4 +83,120 @@ defmodule Cog.Command.Pipeline.Initializer do
...
+  defp alert_self_registration_success(user, request, state) do
+    {:ok, message} = self_registration_success_message(user, request)
+    publish_response(message, request["room"], request["adapter"], state)
+  end
+  defp alert_self_registration_failed(request, state) do
+    {:ok, message} = self_registration_failed_message(request)
+    publish_response(message, request["room"], request["adapter"], state)
+  end
+
+  defp publish_response(message, room, adapter, state) do
```
christophermaier: Some of the comments in here are a little out of place / misleading given the use of this particular function. We should try and pull out the original implementation to some other module and use that here instead of the copy / paste.

lib/cog/command/pipeline/initializer.ex
```diff
@@ -68,4 +83,120 @@ defmodule Cog.Command.Pipeline.Initializer do
...
+
+    context = %{
+      handle: handle,
+      first_name: request["sender"]["first_name"],
+      username: user.username,
+      mention_name: adapter.mention_name(handle),
+      display_name: adapter.display_name(),
+    }
+
+    Template.render("any", "self_registration_success", context)
```
christophermaier: Why not use the specific adapter name here instead of `"any"`? We'll fallback to `"any"` now anyway, but that leaves open the possibility of adapter-specific templates at a later date.

lib/cog/command/pipeline/initializer.ex
```diff
@@ -68,4 +83,120 @@ defmodule Cog.Command.Pipeline.Initializer do
...
+    adapter = String.to_existing_atom(request["module"])
+    handle = request["sender"]["handle"]
+
+    context = %{
+      handle: handle,
+      mention_name: adapter.mention_name(handle),
+      display_name: adapter.display_name(),
+    }
+
+    Template.render("any", "self_registration_failed", context)
```
christophermaier: Same comment re: `"any"`

lib/cog/command/pipeline/initializer.ex
```diff
@@ -68,4 +83,120 @@ defmodule Cog.Command.Pipeline.Initializer do
...
+
+  defp find_available_username(username) do
+    case Users.is_username_available?(username) do
+      true ->
+        {:ok, username}
+      false ->
+        find_available_username(username, 1)
+    end
+  end
+  defp find_available_username(_username, 21) do
```
christophermaier: Out of curiosity, why a threshold of 20?

kevsmith: Why not? I felt we had to limit the number of tries to prevent the appearance of lockups in truly degenerate cases. 20 was just a number I picked.

lib/cog/command/pipeline/initializer.ex
kevsmith: Because we do the same thing in the executor which was my reference pattern. imo, `Template.render/3` should have a function head that extracts the adapter from the request so we don't duplicate that code as well.

Strikes me as a thing we could tackle in a separate issue/PR.

lib/cog/command/pipeline/initializer.ex
kevsmith: Comments?

config/config.exs
kevsmith: Yep. Already have that as a to do.

lib/cog/command/pipeline/initializer.ex
kevsmith: 1. Good call on `Cog.Adapter.chat_adapter?`
1. It does but you never know how the config might get mangled. Adding `false` seemed like cheap conservative insurance.

lib/cog/command/pipeline/initializer.ex
christophermaier: The code comments in the body of this function that were copied from the original implementation. For instance, *this* particular function should never be called from the context of the `http` adapter, but it's written and commented as though it could.

lib/cog/command/pipeline/initializer.ex
christophermaier: Sounds good.

PR #842, opened 2016-07-07
Disable Ecto query logging

### LINKED ISSUE #841: Cog logging is too verbose
Cog generates a lot of log output, much of it not useful (Ecto I'm looking at you).

kevsmith:
Disable Ecto query logging "noise" since almost (all?) of the output isn't helpful.

Lines logged during startup:
Before: 1552
After: 115

Fixes #841


vanstee:
yay! 🎉

PR #843, opened 2016-07-07
Setup a docker-compose override file.

tecnobrat:
This moves the "secrets" into a  docker-compose override.  This allows the docker-compose file to be committed, and we have an example override.

Also adds the new bootstrap env vars so that you can launch cog without any configuration at all.


docker-compose.override.example.yml
```diff
@@ -0,0 +1,28 @@
...
+    - COG_BOOTSTRAP_FIRST_NAME=foo
+    - COG_BOOTSTRAP_LAST_NAME=bar
+    - COG_BOOTSTRAP_PASSWORD=foobar
+    - COG_BOOTSTRAP_USERNAME=foo
+    - COG_API_URL_HOST=192.168.99.100
+    - COG_API_URL_PORT=80
+    - COG_TRIGGER_URL_HOST=192.168.99.100
+    - COG_TRIGGER_URL_PORT=4001
+    - COG_SERVICE_URL_HOST=192.168.99.100
+    - COG_SERVICE_URL_PORT=4002
```
tecnobrat: Perhaps these ports can be in the base `docker-compose`

imbriaco: Yeah, I think they make sense in the base because they match the defaults from Cog.

imbriaco:
Thanks for the PR, I like the direction!  

One behavior that we have with the existing `docker-compose.example.yml` that I'd like to keep is that we only require users to set a couple of environment variables if they just want to kick the tires with a test environment. I think we can keep that behavior while still adding the override file which I quite like by making the defaults for `SLACK_API_TOKEN` and the `*_URL_HOST` variables use environment variable interpolation in the override file for their defaults.

I also think we should use default values for the `COG_BOOTSTRAP_*` variables that mirror what Cog does by default as much as possible. Perhaps:

```
COG_BOOTSTRAP_USERNAME=admin
COG_BOOTSTRAP_PASSWORD=changeme
COG_BOOTSTRAP_EMAIL_ADDRESS=cog@localhost
COG_BOOTSTRAP_FIRST_NAME=Cog
COG_BOOTSTRAP_LAST_NAME=Administrator
```

What do you think?

tecnobrat:
@imbriaco made those changes.

Also tweaked the hostnames to be `cog` and `postgres` instead of aliasing to `localhost`, which has the potential to cause some significant issues.

PR #844, opened 2016-07-08
Allow commands to abort pipeline execution

### LINKED ISSUE #691: Allow commands to return a simple error
Right now if you return an error from a command, the bot prints the following message:

```
An error has occurred.

At `2016-05-12T20:36:16Z`, @vanstee initiated the following pipeline, assigned the unique ID `ec536613be7840189c133a932f10eaeb`:

   ` help wut`

The pipeline failed planning the invocation:

   `operable:help wut`

The specific error was:

   Whoops! An error occurred. Command "wut" does not exist
```

That's great for when a command unexpectedly crashes, but if you mistyped something it would be nice for the bot to just say:

```
help: ERROR: Command "wut" does not exist
```

kevsmith:
Implements the Cog part of the design described in #691.

Fixes #691

### DIFF
--- a/lib/cog/command/pipeline/executor.ex
+++ b/lib/cog/command/pipeline/executor.ex
@@ -233,9 +233,9 @@ defmodule Cog.Command.Pipeline.Executor do
           "ok" ->
             collected_output = collect_output(resp, state.output)
             {:next_state, :execute_plan, %{state | output: collected_output}, 0}
-          "stop" ->
+          "abort" ->
             collected_output = collect_output(resp, state.output)
-            terminate_pipeline(%{state | output: collected_output})
+            abort_pipeline(%{state | output: collected_output})
           "error" ->
             fail_pipeline_with_error({:command_error, resp}, state)
         end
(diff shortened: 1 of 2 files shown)


christophermaier:
What's the intended distinction between "aborted" and "error"? The first is for intentional failures, but the latter is for unintended / unhandled situations (e.g., invoking a method on `nil`, etc.)?

kevsmith:
Yep. Your description is spot on. I'm going to add some verbiage to that effect in the docs for 0.10.

lib/cog/command/pipeline/executor.ex
```diff
@@ -231,20 +231,11 @@ defmodule Cog.Command.Pipeline.Executor do
...
-                                                            &store_with_template(&1, resp.template))
-                                   # body may be a map or a list
-                                   # of maps; if the latter, we
-                                   # want to accumulate it into
-                                   # one flat list
-                               end
+            collected_output = collect_output(resp, state.output)
             {:next_state, :execute_plan, %{state | output: collected_output}, 0}
+          "abort" ->
+            collected_output = collect_output(resp, state.output)
```
christophermaier: I believe this collected output can theoretically include successful responses as well as the failing one (e.g., the first few iterations were good, but the last one failed).

Is this intentional, or do we only want to template a response for just the failing case?

kevsmith: I tried to be conservative and err on the side of providing more information rather than not quite enough in the case of a command aborting. I'm on the fence on which is the correct behavior, tho.

PR #848, opened 2016-07-12
Upgrade to Elixir 1.3.1

kevsmith:


Dockerfile
```diff
@@ -10,13 +10,16 @@ RUN echo "@operable https://storage.googleapis.com/operable-apk/" > /etc/apk/rep
...
-# build stage layer.
+# Install runtime dependencies & nice-to-have packages
+RUN apk update -U && apk add bash ca-certificates curl git openssl postgresql-client
+
+# Install Erlang
 RUN apk update -U && \
-    apk add erlang erlang-crypto erlang-dev erlang-ssh erlang-ssl erlang-mnesia erlang-syntax-tools erlang-parsetools \
-            erlang-xmerl elixir \
-            bash ca-certificates curl git openssl postgresql-client
+    apk add `apk search erlang | grep -E "cos|eldap|gs|mibs|snmp|common-test|jinterface|megaco|diameter|odbc|observer|orber|test-server" -v | sed "s/-18.3.2-r0//g"`
```
vanstee: This freaks me out a little bit, but it's fine.

vanstee:
👍

Dockerfile
kevsmith: It's definitely ugly but short of `apk add`-ing all the Erlang things it's the only way I could think of to keep the image size down.

PR #856, opened 2016-07-14
Use first word to detect mention name

### LINKED ISSUE #855: Substring match on mention name yields false positives
Given two Cog instances, A and B, where B's name starts with A's name -- `cog_staging` and `cog`, for example -- both bots will respond when B is addressed via `@` style mention.

kevsmith:
Fixes #855


vanstee:
Does this still work with `@bender: help` style invocations? Matching the bot name + `\b` could work.

christophermaier:
👍

PR #857, opened 2016-07-15
Use cog link value rather than localhost for relay connection

vanstee:
Was reported in chat that this fixed the relay connection issue for some folks.


kevsmith:
Included in merged PR #843

PR #862, opened 2016-07-18
bundle.name -> bundle_name

### LINKED ISSUE #861: Relay managed config hangs when more than one bundle is configured.
Relays seems to hang when attempting to start if more than one bundle has dynamic configuration enabled. I don't have any additional debugging information right now, but intend to troubleshoot further if time permits this weekend.

kevsmith:
Typo caused Cog to crash when calculating dynamic config
signatures. Since it would recover quickly the crash would give the
appearance of Relay hanging.

Fixes #861


imbriaco:
Sneaky sneaky failure mode!

PR #866, opened 2016-07-19
update version requirement for Postgres

jeffweiss:
Prior to this commit the enumerated requirement was simply `Postgres`; however,
`priv/repo/migrations/20160524190528_parse_tree_is_json.exs` attempts to use
the `jsonb` type, which is not available until Postgres 9.4 and later.


vanstee:
Thanks @jeffweiss

PR #868, opened 2016-07-20
Only checksum embedded bundle when MIX_ENV=dev

christophermaier:
TL;DR - checksums can change with different MIX_ENV, a detail that is
irrelevant for our purposes. Read the code comments for details.

Also changes the stored checksum for the embedded bundle; the old value
was actually computed under MIX_ENV=test. This change returns it to the
MIX_ENV=dev checksum that it originally was (but the version will stay
the same, because that's how this stuff works....)


kevsmith:
LGTM modulo testing the 0.10 -> 0.11 upgrade path to ensure upgrades work smoothly.

PR #884, opened 2016-07-25
force content-type of application/json for amazon sns

imbriaco:
The Amazon Simple Notification service sets the `Content-Type` to `text/plain` for all `POST` requests, despite the fact that it sends a JSON payload. This PR forces the `Content-Type` to be `application/json` when we detect the SNS user agent for `POST` requests to the Trigger endpoint.


imbriaco:
I don't love special casing this, but I also don't love the alternative which is to assume that `text/plain` requests really include JSON payloads. This seems like the lesser of two evils right now.

imbriaco:
Reviewed by @christophermaier and given 👍  in Slack. Merging.

PR #886, opened 2016-07-27
Upgrade to emqttd 1.1.2

### LINKED ISSUE #865: Update emqttd to latest stable
`emqttd` currently causes compilation issues on R19. Let's see if we can upgrade without too much work.

kevsmith:
This PR improves Cog's message bus in two ways:

1. Upgrades emqttd to 1.1.2 (latest stable release) and current mater for emqttc.
1. Uses upstream deps directly meaning we no longer have to carry our own fork.

I've verified I can bring up Cog, a Relay, and execute a pipeline but would like some review and testing before merging.

Fixes #865


priv/config/common_mqtt.exs
```diff
@@ -6,15 +6,59 @@ config :emqttd, :access,
...
+
+# config :emqttd, :mqtt,
+#   packet: [max_clientid_len: 128, max_packet_size: 163840],
+#   client: [ingoing_rate_limit: :'64KB/s', idle_timeout: 1],
+#   session: [max_inflight: 100, unack_retry_interval: 60,
+#             await_rel_timeout: 15, max_awaiting_rel: 0,
+#             collect_interval: 0, expired_after: 4],
+#   queue: [max_length: 100, low_watermark: 0.2, high_watermark: 0.6,
+#           queue_qos0: true],
+#   modules: [presence: [qos: 0]]
```
vanstee: Are these still useful to keep around?

lib/cog/bus_driver.ex
```diff
@@ -47,34 +49,38 @@ defmodule Cog.BusDriver do
...
+                # SSL misconfigured (either cert_file is nil or key_file is nil)
+                true ->
+                if cert_file == nil do
+                  Logger.error("Message bus SSL configuration error. Path to certificate file is empty.")
+                  {:error, {:missing_config, :cert_file}}
+                else
+                  Logger.error("Message bus SSL configuration error. Path to key file is empty.")
+                  {:error, {:missing_config, :key_file}}
+                end
+          end
```
vanstee: Looks like a little bit of weird indentation above. Might be nice to combine the two conditionals above into a single `case` but not a big deal.

vanstee:
Left some minor comments but LGTM.

priv/config/common_mqtt.exs
kevsmith: Whoops. Those need to be deleted.

PR #887, opened 2016-07-28
Manually update embedded bundle version

kevsmith:
Reliably detecting changes to BEAM files turns out to be MUCH harder
than expected. Over the past few days I've seen the embedded bundle task
get false positives when it checks for changes which has resulted in
either unexpected embedded bundle version bumps or manual reverts.

This commit replaces the automated change detection machinery with an
explicit bundle version number which will need to be manually bumped as
needed. It's not ideal but Erlang and Elixir's surprising non-stable code
generation makes reliable change detection a real PITA.


christophermaier:
Makes me sad, but OK 👍

PR #888, opened 2016-07-28
Support layered dynamic bundle configuration

christophermaier:
Adds the ability to manage "layers" of dynamic bundle configuration. The
previous notion of dynamic configuration is referred to now as the
"base" layer, but now we may add "room" and "user" layers on top. Relay
can apply these in order to override the base layer depending on which
user issues a command, and from which room it is issued.

Layers can have names, indicating which user or room they apply to. The
base layer is always named "config" by convention (it could be named
anything, really).

There can only be one base layer per bundle, but multiple room and user
layers, provided they each have a different name.

A REST API is provided to manipulate each layer of a bundle
individually. Additionally, all layers for a bundle may be retrieved at
once.

Cog now returns all layers for all bundles that a Relay is currently
serving.

### DIFF
--- a/lib/cog/repository/bundles.ex
+++ b/lib/cog/repository/bundles.ex
@@ -14,6 +14,17 @@ defmodule Cog.Repository.Bundles do
   alias Ecto.Adapters.SQL
   import Cog.UUID, only: [uuid_to_bin: 1]
 
+  @bundle_preloads [:versions, :relay_groups]
+  @bundle_version_preloads [:bundle,
+                            :enabled_version_registration,
+                            permissions: [:bundle],
+                            commands: [:options, [command: [:bundle, rules: (from r in Rule, where: r.enabled)]]]]
+  @enabled_version_preloads [enabled_version: [:bundle,
+                                               :enabled_version_registration,
+                                               permissions: [:bundle],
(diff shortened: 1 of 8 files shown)


lib/cog/repository/bundles.ex
```diff
@@ -14,6 +14,17 @@ defmodule Cog.Repository.Bundles do
...
+  @bundle_preloads [:versions, :relay_groups]
+  @bundle_version_preloads [:bundle,
+                            :enabled_version_registration,
+                            permissions: [:bundle],
+                            commands: [:options, [command: [:bundle, rules: (from r in Rule, where: r.enabled)]]]]
+  @enabled_version_preloads [enabled_version: [:bundle,
+                                               :enabled_version_registration,
+                                               permissions: [:bundle],
+                                               commands: [command: :bundle]]]
+  @bundle_dynamic_config_preloads [:bundle]
```
kevsmith: Curious - why use an attribute if there's just one value?

christophermaier: Consistency with other preloads (and it's used in multiple places)

kevsmith:
LGTM 👍

PR #890, opened 2016-07-29
BE QUIET, WE DON'T CARE!

christophermaier:
Stops the annoying and useless "compilation taking longer than 10s"
messages when compiling.


mpeck:
:shipit:

PR #891, opened 2016-07-29
Excising dead code

kevsmith:
Removes `Carrier.CredentialManager` and friends. Uses a hardcoded Relay UUID for the embedded which should be ok since we don't write it to persistent storage.


christophermaier:
👍

vanstee:
I think we might be able to kill this as well https://github.com/operable/cog/blob/master/config/config.exs#L122

kevsmith:
@vanstee Good catch on the config entry. Added a commit to delete it.

PR #892, opened 2016-08-01
Username and/or email changes for flywheel

vanstee:
* Allow users to retrieve a token with email address instead of username
* Identify group members by username

### DIFF
--- a/priv/repo/migrations/20160803181844_replace_username_with_slack_handle.exs
+++ b/priv/repo/migrations/20160803181844_replace_username_with_slack_handle.exs
@@ -1,0 +1,35 @@
+defmodule Cog.Repo.Migrations.ReplaceUsernameWithSlackHandle do
+  use Ecto.Migration
+  alias Cog.Repo
+  alias Cog.Models.User
+
+  def change do
+    users = User
+    |> Repo.all
+    |> Repo.preload([chat_handles: :chat_provider])
+
+    Enum.each(users, fn user ->
+      username_is_email = user.username == user.email_address
+
+      slack_handle = Enum.find_value(user.chat_handles, fn chat_handle ->
+        case chat_handle.chat_provider.name do
+          "slack" ->
+            chat_handle.handle
+          _ ->
+            nil
+        end
+      end)
+
+      slack_handle = slack_handle || user.username
+
+      params = case username_is_email do
+        false ->
+          %{}
+        true ->
+          %{"username" => slack_handle}
+      end
+
+      Repo.update!(User.changeset(user, params))
+    end)
+  end
+end
--- a/priv/repo/migrations/20160803181844_replace_username_with_slack_handle.exs
+++ b/priv/repo/migrations/20160803181844_replace_username_with_slack_handle.exs
@@ -1,35 +1,15 @@
 defmodule Cog.Repo.Migrations.ReplaceUsernameWithSlackHandle do
   use Ecto.Migration
-  alias Cog.Repo
-  alias Cog.Models.User
 
   def change do
-    users = User
-    |> Repo.all
-    |> Repo.preload([chat_handles: :chat_provider])
-
-    Enum.each(users, fn user ->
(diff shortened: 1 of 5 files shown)


priv/repo/migrations/20160803181844_replace_username_with_slack_handle.exs
```diff
@@ -0,0 +1,35 @@
+defmodule Cog.Repo.Migrations.ReplaceUsernameWithSlackHandle do
+  use Ecto.Migration
+  alias Cog.Repo
+  alias Cog.Models.User
+
+  def change do
+    users = User
+    |> Repo.all
+    |> Repo.preload([chat_handles: :chat_provider])
```
christophermaier: In previous migrations when we've used models directly, it has bitten us badly, causing us to have to go back and rewrite old migrations. Using model code ends up coupling the database migration to the state of Cog code _when the migration was written_; migrations should only be coupled to the state of the database resulting from the previous migration. If we change the structure of the models in the future (e.g., perhaps we no longer have a `ChatProvider` model? Depending on how the new Adapter work goes, we might not) then this could end up blowing up in the middle of a migration, perhaps making it impossible to install a new bot.

This can be rewritten to use an `execute` call with an [anonymous PL/pgsql block](https://www.postgresql.org/docs/9.0/static/sql-do.html)... I had to do something similar [here](https://github.com/operable/cog/blob/af167bc9e842b4601a0902d96f0beeae0d4f809f/priv/repo/migrations/20160505200538_bundle_version_schema.exs#L181-L203)

You might also be able to use `Cog.Repo` calls referring to raw table names and columns, instead of using models, if that makes the code easier to write. I think that should provide sufficient insulation, too.

web/controllers/v1/token_controller.ex
```diff
@@ -6,23 +6,60 @@ defmodule Cog.V1.TokenController do
...
+            Passwords.matches?(nil, "")
+
+            conn
+            |> put_status(:forbidden)
+            |> json(%{errors: "Invalid credentials"})
+        end
+      {:error, :invalid_params} ->
+        conn
+        |> put_status(:unauthorized)
+        |> json(%{errors: "Must supply both username or email and password"})
```
christophermaier: It might be less ambiguous to say "both username/email and password"... anytime I see "or"s and "and"s without parentheses, my spidey sense gets triggered 😄

web/controllers/v1/token_controller.ex
```diff
@@ -6,23 +6,60 @@ defmodule Cog.V1.TokenController do
...
+      when not(is_nil(email)) and not(is_nil(password)) do
+    {:ok, params}
+  end
+
+  defp validate_params(_params) do
+    {:error, :invalid_params}
+  end
+
+  defp find_user(%{"username" => username}) do
     case Repo.get_by(User, username: username) do
```
christophermaier: Might be a good time to pull this functionality into the core in a Repository somewhere.

priv/repo/migrations/20160803181844_replace_username_with_slack_handle.exs
```diff
@@ -0,0 +1,35 @@
...
+      slack_handle = slack_handle || user.username
+
+      params = case username_is_email do
+        false ->
+          %{}
+        true ->
+          %{"username" => slack_handle}
+      end
+
+      Repo.update!(User.changeset(user, params))
```
christophermaier: We only need to do this in the case that `username_is_email == true`

priv/repo/migrations/20160803181844_replace_username_with_slack_handle.exs
vanstee: I'm going to make this change. But, this is really just a symptom of how we're misusing migrations to rebuild the db schema. Ideally, we'd just be storing a schema dump of the database and use migrations to alter that dump. I remember seeing a schema dump feature available in Ecto 2.0. We should research what'd take to start using it.

PR #896, opened 2016-08-02
Adds support for Cog sending emails

### LINKED ISSUE #895: Cog needs to be able to send emails
Cog needs to be able to send emails. We can use Bamboo (https://github.com/thoughtbot/bamboo) from Thoughtbot since we already have some familiarity with it. We can use the bamboo SMTP adapter (https://github.com/fewlinesco/bamboo_smtp) and use a few env vars for config.

Required: `COG_SMTP_SERVER`, `COG_SMTP_PORT`, `COG_SMTP_USERNAME`, `COG_SMTP_PASSWORD`

Optional: `COG_SMTP_SSL` default `false`, `COG_SMTP_RETRIES` default 1

mpeck:
resolves #895


vanstee:
LGTM

PR #898, opened 2016-08-03
Fix emqttd mgmt;Refine process tree structure

### LINKED ISSUE #897: Cog doesn't stop or restart emqttd cleanly
`Cog.BusDriver` is a GenServer process responsible for configuring, starting, and stopping `emqttd` as an included application. Logic for configuring and starting `emqttd` is implemented in `Cog.BusDriver.init/1`. Logic for stopping the message bus is implemented in `Cog.BusDriver.terminate/2.` We've recently found two issues which point to implementation bugs in `BusDriver`:

1. Relays don't always refresh their online status with Cog when the app restarts. This only happens sometimes and only when Cog's process tree is reinitialized (restarting the top level supervisor, for example). It never happens if the entire VM is restarted.

1. Cog's test suite has started experiencing random failures. These failures don't seem to correlate with any test or test module but appear to happen randomly during runs. Once the error occurs all subsequent tests fail.

kevsmith:
- `Cog.BusDriver` now traps exits if it successfully starts
  `emqttd`. This allows Cog to keep the message bus state in sync with
  Cog's overall state. If Cog is up, the message bus is up. If Cog is
  down, the message bus is down.

- Removed the `mnesia` data file deletion logic from
  `Cog.BusDriver. Wound up causing more problems than it solved. We can
  tell users to delete the data dir in the release notes.

- Moved all of Cog core, modulo `Cog.Repo` and `Cog.BusDriver`, to a
  separate supervisor named `Cog.CoreSup` with a restart strategy of
  one-for-one. The top level supervisor is now responsible for
  `Cog.Repo`, `Cog.BusDriver`, and `Cog.CoreSup` and uses the
  one-for-all restart strategy. This should stabilize Cog as all
  processes will be restarted if the database or message bus goes down.

Fixes #897

### DIFF
--- a/lib/cog.ex
+++ b/lib/cog.ex
@@ -5,19 +5,11 @@ defmodule Cog do
   import Supervisor.Spec, warn: false
 
   def start(_type, _args) do
-    adapter_supervisor = get_adapter_supervisor!()
-    children = [worker(Cog.Repo, []),
-                worker(Cog.BusDriver, [], shutdown: 10000),
-                worker(Cog.TokenReaper, []),
-                supervisor(Cog.Relay.RelaySup, []),
-                supervisor(Cog.Command.CommandSup, []),
-                supervisor(adapter_supervisor, []),
-                supervisor(Cog.Endpoint, []),
-                supervisor(Cog.TriggerEndpoint, []),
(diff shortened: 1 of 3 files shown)


christophermaier:
:shipit:

PR #901, opened 2016-08-04
Use Conduit-typed messages

christophermaier:
Depends on https://github.com/operable/conduit/pull/1; when that's merged, `mix.exs` in this PR will need to be updated accordingly.


lib/carrier/messaging/connection.ex
```diff
@@ -94,10 +94,14 @@ defmodule Carrier.Messaging.Connection do
...
-  def publish(conn, message, kw_args) when is_map(message) do
+
+  # Here, we assume we're being passed a Conduit-enabled struct
+  # (aside: any way to verify that statically?) We'll do the encoding
+  # to JSON internally
+  def publish(conn, %{__struct__: _}=message, kw_args) do
     topic = Keyword.fetch!(kw_args, :routed_by)
 
-    encoded = Poison.encode!(message)
+    encoded = message.__struct__.encode!(message)
```
kevsmith: Might be nice to pull this into Conduit, no? Then we could check to see if the struct is Conduit-based or not and return an appropriate error.

```elixir
Conduit.encode!(message)
```

lib/cog/adapter.ex
```diff
@@ -31,35 +31,31 @@ defmodule Cog.Adapter do
...
 
       def handle_call({:receive_message, sender, room, message, id, initial_context}, _from, state) do
-        message = payload(sender, room, message, id, initial_context)
+        message = %Cog.Messages.AdapterRequest{id: id,
+                                               sender: sender,
+                                               room: room,
+                                               text: message,
+                                               initial_context: initial_context,
+                                               adapter: name(),
+                                               module: to_string(__MODULE__),
```
kevsmith: I really really really want this to die die die :)

kevsmith:
I think some/a lot of this new code will change when we integrate adapters v2 but this is a great start.

👍  to merge

lib/carrier/messaging/connection.ex
christophermaier: Agreed; added #902 to capture this.

lib/cog/adapter.ex
christophermaier: indeed

PR #904, opened 2016-08-05
Sync cogctl and cog in the Docker image

christophermaier:
Previously, if you didn't edit the `scripts/docker-build` file
beforehand, you could build an image that had the "wrong" version of
`cogctl` included. This version would always be build from `master` of
`cogctl`, and not necessarily the version that corresponded with the
`cog` you were building.

Now, the script figures out what branch `cog` is checked out to, and if
`cogctl` has a branch of the same name, it will check `cogctl` out on
that branch. If it doesn't, then it will fall back to the `master`
branch in `cogctl`.

Previously, we had `.git` in our `.dockerignore` file, but that would
make it impossible to determine `cog`'s current branch. Now, we don't
ignore `.git`, but we do remove it from the image before we finish
building. We also do the same for `cogctl` (which we weren't before).

### DIFF
--- a/scripts/docker-build
+++ b/scripts/docker-build
@@ -9,14 +9,46 @@ mix hex.info
 
 # build cog
 cd /home/operable/cog
+
+# We do this so we know what branch to build for cogctl later on
+COG_BRANCH=`git symbolic-ref HEAD | sed 's!refs/heads/!!'`
+if [ -z "$COG_BRANCH" ]
+then
+    echo "Looks like cog isn't a git repository! Do you have '.git' in the '.dockerignore' file? If so, remove it and try again."
+    exit 1
+fi
+
+# Get rid of .git now that we know the branch
+rm -Rf .git
+
 mix clean
 mix deps.get
 mix deps.compile
 mix compile
 
 # build cogctl
+#
+# First take a look at what branch / tag we're on in Cog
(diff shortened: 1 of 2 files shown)


scripts/docker-build
```diff
@@ -9,14 +9,46 @@ mix hex.info
 
 # build cog
 cd /home/operable/cog
+
+# We do this so we know what branch to build for cogctl later on
+COG_BRANCH=`git symbolic-ref HEAD | sed 's!refs/heads/!!'`
```
kevsmith: Can also do this with a single git command: `git rev-parse --abbrev-ref HEAD`

scripts/docker-build
```diff
@@ -9,14 +9,46 @@ mix hex.info
...
+    exit 1
+fi
+
+# Get rid of .git now that we know the branch
+rm -Rf .git
+
 mix clean
 mix deps.get
 mix deps.compile
 mix compile
```
kevsmith: I _think_ `compile` will also compile deps.

kevsmith:
LGTM

scripts/docker-build
christophermaier: Ah, cool... I'll use that instead.

PS: Obligatory https://git-man-page-generator.lokaltog.net/

kevsmith:
👍

PR #906, opened 2016-08-09
password reset tweaks

mpeck:
Just a couple tweaks for password resets

### DIFF
--- a/test/views/email_view.exs
+++ b/test/views/email_view.exs
@@ -1,0 +1,12 @@
+defmodule Cog.EmailViewTest do
+  use Cog.ConnCase, async: true
+
+  test "reset_url generates the proper url" do
+    Application.put_env(:cog, :password_reset_base_url, "http://example.com:4000/reset-password")
+
+    url = Cog.EmailView.reset_url("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee")
+
+    assert url == "http://example.com:4000/reset-password/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+  end
+
+end
--- a/test/views/email_view.exs
+++ b/test/views/email_view.exs
@@ -9,4 +9,13 @@ defmodule Cog.EmailViewTest do
     assert url == "http://example.com:4000/reset-password/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
   end
 
+  test "reset_url is not shown if password_reset_base_url is not set" do
+    Application.delete_env(:cog, :password_reset_base_url)
+
+    assigns = [token: "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"]
+    email = Phoenix.View.render_to_string(Cog.EmailView, "reset_password.text", assigns)
+
+    refute email =~ "reset your password using the webui"
+  end
+
 end
--- a/web/controllers/v1/password_reset_controller.ex
+++ b/web/controllers/v1/password_reset_controller.ex
@@ -9,6 +9,9 @@ defmodule Cog.V1.PasswordResetController do
          {:ok, _} <- Users.request_password_reset(user) do
       send_resp(conn, :no_content, "")
     else
+      {:error, :not_found} ->
+        Logger.warn("Unknown email address sent for password reset")
+        send_resp(conn, :no_content, "")
       {:error, error} ->
         Logger.warn("Failed to generate password reset: #{inspect error}")
         send_resp(conn, :internal_server_error, "")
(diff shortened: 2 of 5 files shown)


test/views/email_view.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Cog.EmailViewTest do
+  use Cog.ConnCase, async: true
+
+  test "reset_url generates the proper url" do
+    Application.put_env(:cog, :password_reset_base_url, "http://example.com:4000/reset-password")
```
christophermaier: Why is this URL set in application configuration?

web/controllers/v1/password_reset_controller.ex
```diff
@@ -4,16 +4,12 @@ defmodule Cog.V1.PasswordResetController do
...
-          _ ->
-            send_resp(conn, :internal_server_error, "")
-        end
-      _ ->
-        send_resp(conn, :ok, "")
+    with {:ok, user} <- Users.by_email(email_address),
+         {:ok, _} <- Users.request_password_reset(user) do
+      send_resp(conn, :no_content, "")
+    else
+      {:error, _error} ->
```
christophermaier: Might be worth logging this at least.

test/views/email_view.exs
mpeck: Where else would we set it?

christophermaier: Can't it be generated from `Cog.Router.Helpers`?

mpeck: Ah, I think I see what your saying now. `password_reset_base_url` isn't a url for Cog. It's a url for flywheel. Technically it could be a url for any webui for Cog but...

Let me explain how the password reset bits work. All of the view elements live in flywheel, Cog just functions as the api for resets. Flywheel will `post` to `/v1/password_reset` with an email_address in the body. If the email is a valid user email cog will add a record to the password_resets table in the db. It will then send an email with a url generated by appending the password reset id to `password_reset_base_url`. `password_reset_base_url` is a view in flywheel. So when the user clicks the link they are actually taken to flywheel where they can enter their new password. Then flywheel makes the request back to cog to actually do the reset.

Does that make sense? Forgive me if I sound elementary. For some reason this was hard to explain.

web/controllers/v1/password_reset_controller.ex
```diff
@@ -9,6 +9,9 @@ defmodule Cog.V1.PasswordResetController do
          {:ok, _} <- Users.request_password_reset(user) do
       send_resp(conn, :no_content, "")
     else
+      {:error, :not_found} ->
+        Logger.warn("Unknown email address sent for password reset")
+        send_resp(conn, :no_content, "")
```
christophermaier: wouldn't this be an error?

web/views/email_view.ex
```diff
@@ -2,23 +2,17 @@ defmodule Cog.EmailView do
...
-        by appending a token to the base url.
-
-        Please set the env var 'COG_PASSWORD_RESET_BASE_URL'.
-        """)
-
-        %URI{host: "localhost"}
-
-      base_url ->
-        URI.parse(base_url)
+  def webui? do
```
christophermaier: This could be 

```elixir
def webui?,
  do: not(is_nil(Application.get_env(:cog, :password_reset_base_url))
```

web/views/email_view.ex
```diff
@@ -2,23 +2,17 @@ defmodule Cog.EmailView do
...
+    if Application.get_env(:cog, :password_reset_base_url, false) == false do
+      false
+    else
+      true
     end
+  end
+
+  def reset_url(token) do
+    base_url = Application.get_env(:cog, :password_reset_base_url, nil)
```
christophermaier: `nil` is the default here.

web/controllers/v1/password_reset_controller.ex
mpeck: It would. I'm sending the same response as success in this case though so nefarious people don't just spam the endpoint to try and guess users emails. Maybe it doesn't matter though?

mpeck: oh and btw, you're commenting on an outdated diff.

web/views/email_view.ex
christophermaier: However, `URI.parse(nil)` blows up.

mpeck: I know, I just like to be explicit.

web/views/email_view.ex
christophermaier: I'd also add some documentation to this function... it's not immediately clear why it's named `webuil?`

PR #912, opened 2016-08-15
Replace TemplateCache w/genericized cache impl

kevsmith:
Cog.Util.Cache & co. unifies Cog's in-memory caches into a single implementation.


lib/cog/util/cache_impl.ex
```diff
@@ -0,0 +1,106 @@
...
+    GenServer.call(cacheref, {:delete, key}, :infinity)
+  end
+
+  def close(cacheref) do
+    :erlang.exit(cacheref, :shutdown)
+  end
+
+  def init([args]) do
+    name = Keyword.get(args, :name)
+    ttl = Keyword.get(args, :ttl)
```
christophermaier: Should these be `Keyword.fetch!/2` instead?

lib/cog/util/cache_impl.ex
```diff
@@ -0,0 +1,106 @@
...
+  def store(cacheref, key, value) do
+    GenServer.call(cacheref, {:store, key, value}, :infinity)
+  end
+
+  def delete(cacheref, key) do
+    GenServer.call(cacheref, {:delete, key}, :infinity)
+  end
+
+  def close(cacheref) do
+    :erlang.exit(cacheref, :shutdown)
```
christophermaier: Why not `GenServer.stop/3`?

lib/cog/util/cache_impl.ex
kevsmith: Ah, yeah. Will fix.

lib/cog/util/cache_impl.ex
kevsmith: Because I'm an old fart who forgets such niceties exist? ;-)

PR #913, opened 2016-08-16
New Providers

christophermaier:
Introduces a new "provider" interface, and reworks Cog to use it.

Key things to note: HipChat and IRC support are _removed_ with this PR. HipChat support will be returning shortly, though.


lib/cog/chat/adapter.ex
```diff
@@ -25,21 +29,46 @@ defmodule Cog.Chat.Adapter do
   end
 
   def lookup_user(provider, handle) when is_binary(handle) do
-    GenMqtt.call(@adapter_topic, "lookup_user", %{provider: provider, handle: handle}, :infinity)
+    cache = get_cache
+    case cache[{provider, handle}] do
```
christophermaier: Should we have a more specific cache key, like `{provider, :user, handle}`?

Issue #916, opened 2016-08-16
Piper dies on non-scalar variable bindings


kevsmith:
The pipeline planner dies due to an error raised by Piper when it encounters a non-scalar variable binding.

PR #917, opened 2016-08-16
Prevent non-scalar variable bindings

kevsmith:
Other half of fix for #916


lib/cog/command/pipeline/planner.ex
```diff
@@ -32,16 +33,21 @@ defmodule Cog.Command.Pipeline.Planner do
...
+        do: %Plan{parser_meta: invocation.meta,
+                  options: options,
+                  args: args,
+                  cog_env: cog_env,
+                  invocation_id: invocation.id,
+                  invocation_text: to_string(bound),
+                  invocation_step: step}
+    rescue
+      e in BadValueError ->
+        {:error, BadValueError.message(e)}
```
christophermaier: Does this error "flow" out of the system and back to the user in the right way?

christophermaier: Yup, it does.

PR #922, opened 2016-08-17
Fix access errors

christophermaier:
There were a handful of places where we were still treating chat rooms and users as plain maps, rather than Conduit-enabled structs, which would cause failures when we tried interacting with them in terms of the `Access` protocol, which structs do not implement by default.

This should flush out all those last remaining calls, as well as finishes the transition to completely validated messages, by ensuring that `Cog.Chat.User` and `Cog.Chat.Room` are used everywhere.

Additionally, tests were added around the autoregistration functionality, which is where this problem surfaced.

### DIFF
--- a/lib/cog/support/model_utilities.ex
+++ b/lib/cog/support/model_utilities.ex
@@ -79,7 +79,7 @@ defmodule Cog.Support.ModelUtilities do
     |> User.changeset(%{username: username,
                         first_name: Access.get(options, :first_name, String.capitalize(username)),
                         last_name: Access.get(options, :last_name, "Mc#{String.capitalize(username)}"),
-                        email_address: "#{username}@operable.io",
+                        email_address: Access.get(options, :email_address, "#{username}@operable.io"),
                         password: username})
     |> Repo.insert!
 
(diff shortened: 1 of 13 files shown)


lib/cog/support/model_utilities.ex
```diff
@@ -79,7 +79,7 @@ defmodule Cog.Support.ModelUtilities do
     |> User.changeset(%{username: username,
                         first_name: Access.get(options, :first_name, String.capitalize(username)),
                         last_name: Access.get(options, :last_name, "Mc#{String.capitalize(username)}"),
-                        email_address: "#{username}@operable.io",
+                        email_address: Access.get(options, :email_address, "#{username}@operable.io"),
```
mpeck: Why explicitly use access instead of the shorthand?

mpeck:
:+1:

lib/cog/support/model_utilities.ex
christophermaier: I was just going along with the rest of that code. Technically, doing it via `Access` is more generic, but for this usecase it probably doesn't really matter all that much.

Issue #919, opened 2016-08-16
Improve handling of list option types


imbriaco:
In order to set the value for options of type string, we have to pass the value as a comma separated string. For example, a command named `foo` with a list option named `things` would be called like this:

```
foo --things="one,two,three"
```

The `things` option that is passed to the command would then be set to `[one two three]`. 

This is not ideal either from the command user or developer point of view. 

## User

From the point of view of the user, it would be convenient if options of type `list` allowed the option to be specified more than once in the command invocation, and accumulated all the values. e.g.

```
foo --things one --things two --things three
```

The current comma separated list is also acceptable since that form remains useful in some cases, particularly when building a pipeline.

## Developer

From the developer point of view, the value of an option of type `list` is marshalled into an unfamiliar format. I can see a couple of ways we could handle this:

1. Marshall options with type list as a JSON array, or
2. Use individual environment variables for the number of values for the option and the values, e.g.: `COG_OPT_THING_COUNT` and `COG_OPT_THING_0`, `COG_OPT_THINGS_1`, and so on.

It's not clear to me which of these is the "better" choice right now.

imbriaco:
After thinking about it more, I lean toward the `COG_OPT_THING_#` style for passing lists of options for consistency with how we handle lists of args as well as simplicity of use.

PR #923, opened 2016-08-17
List-valued Options

christophermaier:
Allows list-valued options to be specified one-value-at-a-time:

    do:something --foo=one --foo=two --foo=three

In service of #919


imbriaco:
LGTM!

PR #931, opened 2016-08-24
Use "latest" Docker tag for Cog and Relay images

christophermaier:


mpeck:
Silly question but will latest pull the most recent uploaded image, ie if we are testing for release will it pull dev images, or do you specify which version is the latest?

kevsmith:
It will only pull the images we've specifically tagged as latest. This happens at the very end of the release process.

PR #933, opened 2016-08-24
Organize provider code by implementation

christophermaier:
Instead of tossing all the provider implementation code in a single
directory, consolidate by implementation; all Slack-related code goes in
`lib/cog/chat/slack`, HTTP-related code goes in `lib/cog/chat/http`,
etc.

This will make it easier to manage as we add more implementations, and
more features per implementation.


kevsmith:
👍 if tests are green

PR #941, opened 2016-09-09
Help docs improvements

vanstee:
* Fixes missing command option description
* Condenses options to a single segment of the synopsis once there are more than 5


mpeck:
:+1:

PR #946, opened 2016-09-13
Render Slack Attachments

kevsmith:
This PR adds attachment rendering support to `Cog.Chat.Slack.TemplateProcessor` and tweaks to `Slack.Connector`, `Slack.Provider`, and a fork of `Elixir-Slack` to send the rendered attachments to Slack.

Also added a utility module, `Cog.Util.Colors`, which translates CSS color names to their corresponding hex codes. This should make working with colors more natural for command authors.

Hopefully the `Elixir-Slack` fork is short lived and my small change is upstreamed.


lib/cog/chat/slack/template_processor.ex
```diff
@@ -1,8 +1,16 @@
...
-  def render(directives),
-    do: Enum.map_join(directives, &process_directive/1)
+  alias Cog.Util.Colors
+
+  def render(directives) do
+    {text, attachments} = directives
+                          |> Enum.map(&process_directive/1)
+                          |> List.flatten
+                          |> consolidate_outputs({"", []})
+  {text, Enum.map(attachments, fn(attachment) -> updated = Map.delete(attachment, "name")
```
christophermaier: Perhaps some comments describing why we do this bit of munging, with pointers to the relevant Slack API docs?

lib/cog/util/colors.ex
```diff
@@ -1,5 +1,8 @@
 defmodule Cog.Util.Colors do
-  hex_codes = File.read!("priv/css-color-names.json") |> Poison.decode!
+
+  @external_resource "priv/css-color-names.json"
```
christophermaier: Should we use `:code.priv_dir(:cog)` here instead?

imbriaco:
Looking good!

I think we want to set the `fallback` attribute to the same content as the `text` attribute when we send it to Slack's API. That content is displayed on devices that aren't the Slack web client like mobile and IRC.

christophermaier:
👍

PR #951, opened 2016-09-14
Attachments!

kevsmith:
This PR uses the new template engine and Slack renderer to display error messages and long command results as color-coded Slack attachments.


lib/cog/chat/slack/template_processor.ex
```diff
@@ -6,13 +6,20 @@ defmodule Cog.Chat.Slack.TemplateProcessor do
...
-  def render(directives) do
-    {text, attachments} = directives
-                          |> Enum.map(&process_directive/1) # Convert all Greenbar directives into their Slack forms
-                          |> List.flatten                   # Flatten nested lists into a single list
-                          |> consolidate_outputs({"", []})  # Separate message text and attachments into separate lists
-    attachments = Enum.map(attachments, &finalize_attachment/1) # Final attachment post-processing for Slack
-    {text, attachments}
+  def render(directives, convert_long_text \\ true)
+
+  def render(directives, true) do
```
christophermaier: Can we make the second argument a keyword list instead? That way we can call it like `render(directives, convert_long_text: true)`... makes it easier to read and extensible in the future.

Bare boolean arguments always trigger my spidey-sense.

lib/cog/chat/slack/template_processor.ex
```diff
@@ -4,14 +4,22 @@ defmodule Cog.Chat.Slack.TemplateProcessor do
...
+    {text, attachments} = render_directives(directives)
+    # Message is too long convert it to an attachment
+    if String.length(text) > @too_long_for_message do
+      {"", [%{"text" => text, "fallback" => text, "color" => Colors.name_to_hex("operableblue"),
+              "title" => "Pipeline Result", "mrkdwn_in" => ["text", "title"]}]}
```
christophermaier: This still swallows any attachments that may have been present... and still probably behaves weirdly when the children of an attachment render as a too-long string.

Are we just going to roll with this for now, since we have the ability to sidestep it for the error attachments and see how it shakes out with wider usage?

kevsmith: Dammit. I thought I had that fixed.......

PR #962, opened 2016-09-19
Improve dynamic config err msg

### LINKED ISSUE #867: Return a better error message when dynamic config file cannot be parsed as yaml
Currently cogctl responds with `cogctl: ERROR: "Internal server error"` if the yaml file could not be parsed as yaml (in this case an ENV var looking string was uploaded). We should make sure we just respond with `cogctl: ERROR: "Config file could not be parsed as yaml"`.

kevsmith:
Catch Ecto exception and use Cog.ChangesetView to render it.

Fixes #867

### DIFF
--- a/web/controllers/v1/bundle_dynamic_config_controller.ex
+++ b/web/controllers/v1/bundle_dynamic_config_controller.ex
@@ -76,27 +76,34 @@ defmodule Cog.V1.BundleDynamicConfigController do
       nil ->
         not_found(conn, "Bundle #{bundle_id} not found")
       bundle ->
-        case Bundles.create_dynamic_config_for_bundle(bundle, params) do
-          {:ok, dyn_config} ->
-            location = case Map.get(params, "layer") do
-                         "base" ->
-                           Helpers.bundle_dynamic_config_path(conn, :show_layer,
-                                                              dyn_config.bundle_id,
-                                                              dyn_config.layer, %{})
-                         _ ->
-                           Helpers.bundle_dynamic_config_path(conn, :show_layer,
-                                                              dyn_config.bundle_id,
-                                                              dyn_config.layer,
-                                                              dyn_config.name, %{})
-                       end
-            conn
-            |> put_status(:created)
-            |> put_resp_header("location", location)
-            |> render("show.json", dynamic_config: dyn_config)
-          {:error, changeset} ->
+        try do
+          case Bundles.create_dynamic_config_for_bundle!(bundle, params) do
+            {:ok, dyn_config} ->
+              location = case Map.get(params, "layer") do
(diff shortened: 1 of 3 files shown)


web/controllers/v1/bundle_dynamic_config_controller.ex
```diff
@@ -76,27 +76,34 @@ defmodule Cog.V1.BundleDynamicConfigController do
...
-                                                              dyn_config.layer,
-                                                              dyn_config.name, %{})
-                       end
-            conn
-            |> put_status(:created)
-            |> put_resp_header("location", location)
-            |> render("show.json", dynamic_config: dyn_config)
-          {:error, changeset} ->
+        try do
+          case Bundles.create_dynamic_config_for_bundle!(bundle, params) do
```
vanstee: Not a huge deal but it is a little weird to suffix with `!` when the function itself doesn't raise.

web/controllers/v1/bundle_dynamic_config_controller.ex
```diff
@@ -76,27 +76,34 @@ defmodule Cog.V1.BundleDynamicConfigController do
...
+              conn
+              |> put_status(:created)
+              |> put_resp_header("location", location)
+              |> render("show.json", dynamic_config: dyn_config)
+            {:error, changeset} ->
+              conn
+              |> put_status(:unprocessable_entity)
+              |> render(Cog.ChangesetView, "error.json", changeset: changeset)
+          end
+        rescue
```
christophermaier: This is good, but I'm curious if Phoenix's built-in exception handling will handle this. If it can't out of the box, then I suspect there's just a small bit of configuration we'd need to set in order for it to work.

That being said, I'm certain this is not the only place in the API where we're doing this. It may be worth leaving this as is, and then following up to make a more principled pass through the rest of the code to fix things up.

web/controllers/v1/bundle_dynamic_config_controller.ex
christophermaier: It does, though, since the implementation itself calls a `!` function.

vanstee: D'oh I'm dumb. Nevermind.

kevsmith: It does raise. If it didn't I wouldn't have had to add a `try/rescue` block to the call site.

web/controllers/v1/bundle_dynamic_config_controller.ex
kevsmith: It probably can but not without rejiggering I wanted to avoid this close to a release.

Long way of saying +1

PR #971, opened 2016-09-20
Add global switch to toggle access rule enforcment

### LINKED ISSUE #977: Feature: Global option to disable permission enforcement
Teams having prior experience with other bots like Hubot and Lita need a way to simplify Cog's learning curve especially around permissions and rules. An easy way to do this would be to provide a global toggle switch which controls whether or not Cog enforces permissions and access rules.

kevsmith:
_Caveat: I'm conflicted on whether or not this is a good idea. Feedback encouraged!_

This PR introduces a new config entry, `:access_rules` which controls whether or not Cog enforces access rules associated with command invocations. Setting this entry to `:enforcing` causes Cog to enforce access rules just as it currently does. `:enforcing` is also the default.

Setting the entry to `:unenforcing` disables access rule enforcement **globally** for the entire Cog installation. This is intended to simplify Cog adoption for teams which have previous experience with simpler bots such as Hubot and Lita. Later, they can set `:access_rules` to `:enforcing` and take advantage of Cog's access controls.

Disabling access rule enforcement causes Cog to log an explicit warning at startup. Values other than `:enforcing` and `:unenforcing` are ignored by Cog and logged to the console w/a warning message. Cog defaults to `:enforcing` when unknown config values are encountered.

Fixes #977

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,6 +2,12 @@ use Mix.Config
 import Cog.Config.Helpers
 
 # ========================================================================
+# Set this to :unenforcing to globally disable all access rules.
+# NOTE: This is a global setting.
+# ========================================================================
+config :cog, :access_rules, :unenforcing
+
+# ========================================================================
 # Embedded Command Bundle Version (for built-in commands)
 # NOTE: Do not change this value unless you know what you're doing.
 # ========================================================================
--- a/config/config.exs
+++ b/config/config.exs
@@ -5,7 +5,12 @@ import Cog.Config.Helpers
 # Set this to :unenforcing to globally disable all access rules.
 # NOTE: This is a global setting.
 # ========================================================================
-config :cog, :access_rules, :enforcing
+
+if System.get_env("DISABLE_RULE_ENFORCEMENT") do
+  config :cog, :access_rules, :unenforcing
+else
+  config :cog, :access_rules, :enforcing
+end
 
 # ========================================================================
 # Embedded Command Bundle Version (for built-in commands)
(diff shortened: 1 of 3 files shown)


vanstee:
Sounds good to me.

One alternative idea I've been thinking about would be to add an option to automatically add all permissions from installed bundles to the `cog-admin` group (or another specified group). That way, as an admin, you still can still easy test out new bundles, but still use permissions as normal.

christophermaier:
Can we also have an environment variable to toggle this? Given that we're mainly distributing via containers these days, it seems more complicated to manually tweak the `config.exs` file if you want this feature. Seems like a good way to ease adoption.

christophermaier:
Tangentially, I wonder if a later refinement of this should be disabling rules on a per-bundle basis. If the desired usecase is to help people migrating from other bots without an authorization system, they're going to try Cog, install a bunch of bundles, want to work with the authorization system, and then they're back at square one, because they have to then fix permissions for _everything_ all at once. 

I'm just spitballing, but it might help to bridge that gap by saying, in effect, "OK, we're going to tighten down permissions on bundle X now.... next week bundle Y..., etc."

christophermaier:
Overall, though, this seems like a potentially useful feature. I'm 👍  with an environment variable toggle.

imbriaco:
I'm :+1: to merge this. Additional notes below:

* **Configuration:** I added a `$DISABLE_RULE_ENFORCEMENT` environment variable to toggle this. If that variable is set at all, we configure `:access_rules` to be `:unenforcing`. Otherwise, it is `:enforcing`. 
* **Documentation:** I think we should give this coverage in generic documentation re: rule enforcement and available config knobs, but we should not go out of our way to hedge our other statements with "unless you have rule enforcement disabled" language. 
* **Upgrade Path:** If we get requests for help from users for help migrating from `:unenforcing` to `:enforcing` I think we can help with through an expedient shell script, but I'm not worried enough about it to solve it in advance or to build any core support into Cog for that workflow.

PR #978, opened 2016-09-22
get proper slack ids when redirecting

### LINKED ISSUE #976: Fix how redirects are specified

mpeck:
We weren't always getting the proper slack id when doing redirects. I think either the slack api or the lib that we are using changed. With this change we do a lookup when classifying ids in the connector, making sure to grab the correct identifier.

resolves #976

### DIFF
--- a/lib/cog/chat/slack/connector.ex
+++ b/lib/cog/chat/slack/connector.ex
@@ -300,6 +300,9 @@ defmodule Cog.Chat.Slack.Connector do
   defp classify_id(id, slack),
     do: id |> String.replace(~r/(^<|>$)/, "") |> do_classify_id(slack)
 
+  # We could get an actual id, U1234567, or the user, @username, or the
+  # channel, #channelname, reference. So we check for everything and
+  # return a normalized id.
   defp do_classify_id(<<"#C", id::binary>>, slack),
     do: normalize_channel_id(:by_id, "C#{id}", slack)
   defp do_classify_id(<<"C", id::binary>>, slack),
@@ -336,7 +339,11 @@ defmodule Cog.Chat.Slack.Connector do
     end
   end
   defp normalize_channel_id(:by_name, name, slack) do
-    channels = Map.values(slack.channels)
+    # Private groups and channels are both noted with the '#' but they are
+    # tracked separately. So we have to merge the two lists before searching.
+    channels = slack.channels
+    |> Map.merge(slack.groups)
+    |> Map.values
     case Enum.find(channels, &(&1.name == name)) do
       nil -> classify_error(name)
       channel -> {:channel, channel.id}
--- a/lib/cog/chat/slack/connector.ex
+++ b/lib/cog/chat/slack/connector.ex
@@ -300,12 +300,18 @@ defmodule Cog.Chat.Slack.Connector do
   defp classify_id(id, slack),
     do: id |> String.replace(~r/(^<|>$)/, "") |> do_classify_id(slack)
 
-  # TODO: use Slack.Lookups to verify the "C" and "U" cases here, to
-  # guard against bare user / room names that start with those letters
-  defp do_classify_id(<<"#C", id::binary>>, _), do: {:channel, "C#{id}"}
-  defp do_classify_id(<<"C", id::binary>>, _), do: {:channel, "C#{id}"}
-  defp do_classify_id(<<"@U", id::binary>>, _), do: {:user, "U#{id}"}
-  defp do_classify_id(<<"U", id::binary>>, _), do: {:user, "U#{id}"}
+  defp do_classify_id(<<"#C", id::binary>>, slack),
+    do: normalize_channel_id(:by_id, "C#{id}", slack)
+  defp do_classify_id(<<"C", id::binary>>, slack),
+    do: normalize_channel_id(:by_id, "C#{id}", slack)
+  defp do_classify_id(<<"#", name::binary>>, slack),
+    do: normalize_channel_id(:by_name, name, slack)
+  defp do_classify_id(<<"@U", id::binary>>, slack),
+    do: normalize_user_id(:by_id, "U#{id}", slack)
+  defp do_classify_id(<<"U", id::binary>>, slack),
+    do: normalize_user_id(:by_id, "U#{id}", slack)
+  defp do_classify_id(<<"@", name::binary>>, slack),
+    do: normalize_user_id(:by_name, name, slack)
   defp do_classify_id(other, slack) do
 
     # Try it as a channel name as last resort; this is just for tests
@@ -321,6 +327,40 @@ defmodule Cog.Chat.Slack.Connector do
         Logger.warn("Could not classify Slack identifier '#{other}'")
         :error
     end
+  end
+
+  defp normalize_channel_id(:by_id, id = "C" <> _, slack) do
+    case slack.channels[id] do
+      nil -> classify_error(id)
+      _channel -> {:channel, id}
+    end
+  end
+  defp normalize_channel_id(:by_name, name, slack) do
+    channels = Map.values(slack.channels)
+    case Enum.find(channels, &(&1.name == name)) do
+      nil -> classify_error(name)
+      channel -> {:channel, channel.id}
+    end
+  end
+
+  defp normalize_user_id(:by_id, id = "U" <> _, slack) do
+    case slack.users[id] do
+      nil -> classify_error(id)
+      _user -> {:user, id}
+    end
+  end
+  defp normalize_user_id(:by_name, name, slack) do
+    users = Map.values(slack.users)
+    case Enum.find(users, &(&1.name == name)) do
+      nil -> classify_error(name)
+      user ->
+        {:user, user.id}
+    end
+  end
+
+  defp classify_error(id) do
+    Logger.warn("Could not classify Slack identifier '#{id}'")
+    :error
   end
 
   defp lookup_channel_id_by_name("#" <> channel_name, slack) do
--- a/test/integration/slack_test.exs
+++ b/test/integration/slack_test.exs
@@ -65,9 +65,19 @@ defmodule Integration.SlackTest do
   test "redirecting from a private channel", %{user: user} do
     user |> with_permission("operable:echo")
     private_channel = "group_ci_bot_testing"
+    marker = send_message("Redirect test starts here")
 
-    message = send_message("@#{@bot}: operable:echo blah > #ci_bot_testing", private_channel)
-    assert_response "blah", [after: message]
+    send_message("@#{@bot}: operable:echo blah > #ci_bot_testing", private_channel)
+    assert_response "blah", [after: marker]
+  end
+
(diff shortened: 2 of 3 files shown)


lib/cog/chat/slack/connector.ex
```diff
@@ -323,6 +332,44 @@ defmodule Cog.Chat.Slack.Connector do
     end
   end
 
+  defp normalize_channel_id(:by_id, id = "C" <> _, slack) do
```
christophermaier: When you get here, all the IDs should start with `"C"`, anyway, correct?

lib/cog/chat/slack/connector.ex
```diff
@@ -300,12 +300,21 @@ defmodule Cog.Chat.Slack.Connector do
...
-  defp do_classify_id(<<"#C", id::binary>>, _), do: {:channel, "C#{id}"}
-  defp do_classify_id(<<"C", id::binary>>, _), do: {:channel, "C#{id}"}
-  defp do_classify_id(<<"@U", id::binary>>, _), do: {:user, "U#{id}"}
-  defp do_classify_id(<<"U", id::binary>>, _), do: {:user, "U#{id}"}
+  # We could get an actual id, U1234567, or the user, @username, or the
+  # channel, #channelname, reference. So we check for everything and
+  # return a normalized id.
+  defp do_classify_id(<<"#C", id::binary>>, slack),
+    do: normalize_channel_id(:by_id, "C#{id}", slack)
+  defp do_classify_id(<<"C", id::binary>>, slack),
```
christophermaier: I think group IDs start with `"G"`... how are those being handled?

lib/cog/chat/slack/connector.ex
```diff
@@ -323,6 +332,44 @@ defmodule Cog.Chat.Slack.Connector do
...
+  defp normalize_channel_id(:by_id, id = "C" <> _, slack) do
+    case slack.channels[id] do
+      nil -> classify_error(id)
+      _channel -> {:channel, id}
+    end
+  end
+  defp normalize_channel_id(:by_name, name, slack) do
+    # Private groups and channels are both noted with the '#' but they are
+    # tracked separately. So we have to merge the two lists before searching.
+    channels = slack.channels
```
christophermaier: Might be more descriptive and less confusing to call this variable `channels_and_groups` instead of just `channels`

lib/cog/chat/slack/connector.ex
```diff
@@ -323,6 +332,44 @@ defmodule Cog.Chat.Slack.Connector do
...
+    |> Map.merge(slack.groups)
+    |> Map.values
+    case Enum.find(channels, &(&1.name == name)) do
+      nil -> classify_error(name)
+      channel -> {:channel, channel.id}
+    end
+  end
+
+  defp normalize_user_id(:by_id, id = "U" <> _, slack) do
```
christophermaier: Similarly, wouldn't all the IDs start with `"U"` by the time we get here?

lib/cog/chat/slack/connector.ex
```diff
@@ -300,12 +300,21 @@ defmodule Cog.Chat.Slack.Connector do
   defp classify_id(id, slack),
     do: id |> String.replace(~r/(^<|>$)/, "") |> do_classify_id(slack)
 
-  # TODO: use Slack.Lookups to verify the "C" and "U" cases here, to
```
christophermaier: Is there a reason not to use `Slack.Lookups` for these functions?

lib/cog/chat/slack/connector.ex
christophermaier: Same logic suggests renaming the `normalize_channel_id` function, too.

test/integration/slack_test.exs
```diff
@@ -61,4 +61,58 @@ defmodule Integration.SlackTest do
     message = send_message("@#{@bot}: operable:echo blah", private_channel)
     assert_response "blah", [after: message], private_channel
   end
+
+  test "redirecting from a private channel", %{user: user} do
+    user |> with_permission("operable:echo")
+    private_channel = "group_ci_bot_testing"
+    marker = send_message("Redirect test starts here")
```
christophermaier: What room is this going to? From reading the test, I'd expect it to be going to `#ci_bot_testing`. If that's the default for `send_message`, then we should probably make that explicit so this test is more easily understood.

christophermaier: Reading through other tests in this module, it may be useful to introduce keyword arguments for the originating room... that way you could have code like `send_message("blah blah blah", from: "my_channel")`

That could be a follow-on, change, though, as it would end up affecting other tests, too.

test/integration/slack_test.exs
```diff
@@ -61,4 +61,58 @@ defmodule Integration.SlackTest do
...
+    assert_response "blah", [after: marker]
+  end
+
+  test "redirecting to a private channel", %{user: user} do
+    user |> with_permission("operable:echo")
+    private_channel = "group_ci_bot_testing"
+    marker = send_message("Redirect test starts here", private_channel)
+
+    send_message("@#{@bot}: operable:echo blah > ##{private_channel}")
+    assert_response "blah", [after: marker], private_channel
```
christophermaier: Why is  `private_channel` a plain argument, and not one of the keyword arguments like `after`?

test/integration/slack_test.exs
```diff
@@ -61,4 +61,58 @@ defmodule Integration.SlackTest do
...
+    assert_response "blah", [after: message]
+  end
+
+  test "redirecting to 'me'", %{user: user} do
+    user |> with_permission("operable:echo")
+    marker = send_message("echo here", "@#{@user}")
+
+    send_message("@#{@bot}: operable:echo blah > me")
+    # Since Cog responds when direct messaging it we have to assert
+    # that both our marker text and message test exist.
```
christophermaier: I thought the use of a marker was to get over the fact that different channels have their own clocks? If you're getting a reply from the same channel you're sending from, the result of `send_message("@#{@bot}: operable:echo blah > me")` should serve as your marker.

test/integration/slack_test.exs
```diff
@@ -61,4 +61,58 @@ defmodule Integration.SlackTest do
...
+    assert_response "here\nblah", [after: marker, count: 2], "@#{@user}"
+  end
+
+  test "redirecting to another channel", %{user: user} do
+    user |> with_permission("operable:echo")
+    redirect_channel = "ci_bot_redirect_tests"
+    marker = send_message("echo here", redirect_channel)
+
+    send_message("@#{@bot}: operable:echo blah > ##{redirect_channel}")
+    assert_response "blah", [after: marker], redirect_channel
```
christophermaier: Why are you not asserting a response of `"here\nblah"` for this test, but you are for the `"redirecting to a specific user"`. They seem to be structurally the same.

test/support/adapters/slack/helpers.ex
```diff
@@ -59,6 +59,8 @@ defmodule Cog.Adapters.Slack.Helpers do
         "https://slack.com/api/channels.history"
       "G" <> _ ->
         "https://slack.com/api/groups.history"
+      "D" <> _ ->
```
christophermaier: Does this need to be accounted for when we're classifying IDs in the connector? Where does this code come into play?

lib/cog/chat/slack/connector.ex
mpeck: I didn't know about the "G". I just fixed the functionality that was already here, and we weren't dealing with that before. I'll confirm that everything works as expected, but I believe we are already doing a test that check the private group behavior.

lib/cog/chat/slack/connector.ex
mpeck: Not if it isn't an id. The `"C"` verifies that what was passed was an id. If it isn't an id, then we assume it's the name of a channel and do a lookup.

kevsmith:
I'm not entirely clear on the problem this PR solves. iirc, we were previously receiving redirects that looked like:

`foo --bar > C123456789`
`foo --bar > U012345678`

Do redirects now look like:

`foo --bar > #team-room-a`
`foo --bar > @SomeUser` ?

lib/cog/chat/slack/connector.ex
christophermaier: But you'd be calling the function with `:by_name` as the first argument, then, right?

lib/cog/chat/slack/connector.ex
mpeck: Good point, originally I wasn't doing the group lookup bit, I was just looking at the channel. I'll rename the var. Honestly I'm not super happy with the name of the function either, but I wasn't sure what else to call it. We actually verify that we have an id, if we don't we assume it's the name of a channel or user and then do a lookup.

lib/cog/chat/slack/connector.ex
mpeck: Same as above, we could get an id or a user name here.

test/integration/slack_test.exs
mpeck: That's how the function was written. I was trying to refactor as little as possible.

test/integration/slack_test.exs
mpeck: We aren't getting a reply from the same channel. The original message is sent from the default room `#ci_bot_testing` the reply comes in as a direct message. Probably a goo reason to be more explicit like you brought up earlier.

test/integration/slack_test.exs
mpeck: When messaging Cog directly, Cog always responds. When we drop the marker `echo here` Cog responds with `here`. The marker notes the starting point to start reading messages from Cog. So when we assert said messages it will include the original response from the marker message.

Cog doesn't respond in public or private rooms unless you use the `@bot` syntax. So in this case we don't have to worry about an extra marker message.

mpeck:
Sorry @kevsmith, I forgot to link to the original issue, #976.

I think the problem may have came up with the adapter rework. Previously we would get the id, `C123456789` or `U012345678` when we got to this point. But now we could get either an id or a channel or user name.

There may be an issue deeper in the adapter, but after talking to @christophermaier I thought it would be better to fix here, where it seemed the lowest impact.

lib/cog/chat/slack/connector.ex
mpeck: Oh yeah, good point. I just added the `"C"` bit as an extra check. I can drop it if you think it's confusing.

test/integration/slack_test.exs
mpeck: Yeah, I was trying to refactor as little as possible to keep the PR manageable. But I agree, the tests could use a bit of work.

lib/cog/chat/slack/connector.ex
mpeck: Yeah, `Slack.Lookups` are super naive. They don't take into account rooms or users not existing, they just crash. I'm basically doing what the functions in `Slack.Lookups` do, I'm just verifying the data. Alternatively I could just rescue, but that seemed a bit dirty.

PR #987, opened 2016-09-26
Don't replace ids with names in redirects

### LINKED ISSUE #976: Fix how redirects are specified

mpeck:
When a valid room name or user name is sent in a message slack replaces it with the channel or user id respectively. We take those ids, do a lookup, and then replacing the ids with the original room or user name at the very beginning of execution.

In general that's probably a good idea. I would think that command authors would expect to receive the name of the room or user in a command's text instead of an id. But in the case of redirects, it was causing some unforeseen side effects. When verifying destinations we never know if we will receive an id or a name. This is mostly due to the special keywords 'me' and 'here' which get replaced with their proper ids earlier during destination validation.

This PR skips lookups for redirects at the start of execution. Which should ensure that the pipeline destination module gets ids and can deal with lookup during redirect validation.

Note: There are a couple issues that still need to be dealt with regarding redirect validation. Namely passing a string like `@USER` will cause Cog to crash and passing a tring like `<@KJHKJ>` will cause Cog to return `...`. But I think those are issues that deserve a follow up PR.

resolves #976

### DIFF
--- a/lib/cog/chat/slack/formatter.ex
+++ b/lib/cog/chat/slack/formatter.ex
@@ -1,16 +1,21 @@
 defmodule Cog.Chat.Slack.Formatter do
 
+  @redirect_symbol "&gt;"
+
   @channel ~r/<\#(C.*)(\|(.*))?>/U
   @user    ~r/<\@(U.*)(\|(.*))?>/U
   @command ~r/<\!(.*)(\|(.*))?>/U
   @link    ~r/<(.*)(\|(.*))?>/U
 
-  def unescape(message, slack) do
+  def unescape(original_message, slack) do
+    {message, redirects} = split_redirects(original_message)
+
     message
     |> unescape_channels(slack)
     |> unescape_users(slack)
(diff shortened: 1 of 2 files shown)


lib/cog/chat/slack/formatter.ex
```diff
@@ -70,4 +75,19 @@ defmodule Cog.Chat.Slack.Formatter do
         replace(left <> replacement <> right, regex, replace_fun)
     end
   end
+
+  defp split_redirects(message) do
+    case String.split(message, @redirect_symbol, parts: 2) do
```
christophermaier: This seems rather fragile; consider

```
echo "I really like to draw ASCII arrows: >>----->" > #general
```

christophermaier:
Incidentally, I think we will need to have names going into the destination validation logic, due to our optional `chat://<DESTINATION>` syntax (used with triggers, mostly). Pretty sure `chat://#general` is going to not get turned into an ID by Slack.

PR #988, opened 2016-09-26
Bundle install via registry

vanstee:
This adds support for installing a bundle via the registry.

Example:

```
@vanstee: !bundle install heroku
@cog: Bundle "heroku" version "0.0.4" installed.

@vanstee: !bundle install pagerduty 0.0.1
@cog: Bundle "pagerduty" version "0.0.1" installed.
```

I'll follow this up with a `bundle uninstall` subcommand as well.

### DIFF
--- a/lib/cog/bundle_registry.ex
+++ b/lib/cog/bundle_registry.ex
@@ -9,7 +9,14 @@ defmodule Cog.BundleRegistry do
       %HTTPotion.Response{status_code: 200, body: body} ->
         {:ok, Poison.decode!(body)}
       %HTTPotion.Response{status_code: 404} ->
-        {:error, "Bundle not found"}
+        case version do
+          "latest" ->
+            {:error, {:not_found, bundle}}
+          version ->
+            {:error, {:not_found, bundle, version}}
+        end
+      _ ->
+        {:error, :registry_error}
     end
   end
 end
--- a/lib/cog/bundle_registry.ex
+++ b/lib/cog/bundle_registry.ex
@@ -1,0 +1,15 @@
+defmodule Cog.BundleRegistry do
+  @registry_url "https://bundles.operable.io"
+
+  def get_config(bundle, version) do
+    headers = ["Accepts": "application/json"]
+    response = HTTPotion.get(@registry_url <> "/api/bundles/#{bundle}/#{version}", headers: headers)
+
+    case response do
+      %HTTPotion.Response{status_code: 200, body: body} ->
+        {:ok, Poison.decode!(body)}
+      %HTTPotion.Response{status_code: 404} ->
+        {:error, "Bundle not found"}
+    end
+  end
+end
--- a/lib/cog/commands/bundle.ex
+++ b/lib/cog/commands/bundle.ex
@@ -20,7 +20,7 @@ defmodule Cog.Commands.Bundle do
     "enable <bundle> [<version>]" => "Enable a specific version of an installed bundle",
     "disable <bundle>" => "Disable an installed bundle",
     "versions <bundle>" => "List all installed versions for a given bundle",
-    "install <bundle>[:<version>]" => "Install latest or specified version of bundle from registry"
+    "install <bundle> [<version>]" => "Install latest or specified version of bundle from registry"
   }
 
   permission "manage_commands"
--- a/lib/cog/commands/bundle.ex
+++ b/lib/cog/commands/bundle.ex
@@ -3,7 +3,7 @@ defmodule Cog.Commands.Bundle do
     bundle: Cog.Util.Misc.embedded_bundle
 
   alias Cog.Models.{Bundle, BundleVersion}
-  alias Cog.Commands.Bundle.{List, Versions, Info, Enable, Disable}
+  alias Cog.Commands.Bundle.{List, Versions, Info, Enable, Disable, Install}
 
   require Cog.Commands.Helpers, as: Helpers
   require Logger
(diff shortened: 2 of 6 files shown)


lib/cog/bundle_registry.ex
```diff
@@ -0,0 +1,15 @@
...
+
+  def get_config(bundle, version) do
+    headers = ["Accepts": "application/json"]
+    response = HTTPotion.get(@registry_url <> "/api/bundles/#{bundle}/#{version}", headers: headers)
+
+    case response do
+      %HTTPotion.Response{status_code: 200, body: body} ->
+        {:ok, Poison.decode!(body)}
+      %HTTPotion.Response{status_code: 404} ->
+        {:error, "Bundle not found"}
```
christophermaier: We should probably have a catch-all case for other errors, too (e.g., registry is unavailable, registry crashes, etc.) just so we don't crash Cog.

lib/cog/commands/bundle.ex
```diff
@@ -19,14 +19,10 @@ defmodule Cog.Commands.Bundle do
     "info <bundle>" => "Detailed information on an installed bundle",
     "enable <bundle> [<version>]" => "Enable a specific version of an installed bundle",
     "disable <bundle>" => "Disable an installed bundle",
-    "versions <bundle>" => "List all installed versions for a given bundle"
+    "versions <bundle>" => "List all installed versions for a given bundle",
+    "install <bundle>[:<version>]" => "Install latest or specified version of bundle from registry"
```
christophermaier: I thought we went with version being an optional standalone argument, instead of tacked on with a colon?

lib/cog/commands/bundle.ex
```diff
@@ -19,14 +19,10 @@ defmodule Cog.Commands.Bundle do
...
   }
 
-  @notes """
-  Installation and uninstallation of bundles cannot currently be done via
-  chat; please use `cogctl` for this functionality.
-  """
-
```
christophermaier: Yay! 🎊

test/cog/chat/slack/templates/embedded/bundle_install_test.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Cog.Chat.Slack.Templates.Embedded.BundleInstallTest do
+  use Cog.TemplateCase
+
+  test "bundle-info template" do
+    data = %{"results" => [%{"name" => "heroku",
+                             "version" => "0.0.4"}]}
+
+    expected = """
+    Bundle "heroku" version "0.0.4" installed.
+    """ |> String.strip
```
christophermaier: For short-and-sweet templates like this, a single line is fine, instead of the whole triple-quote-and-strip dance.

test/integration/commands/bundle_test.exs
```diff
@@ -61,4 +61,8 @@ defmodule Integration.Commands.BundleTest do
     assert_error_message_contains(response, "Unknown subcommand 'not-a-subcommand'")
   end
 
+  test "installing a bundle", %{user: user} do
+    [payload] = send_message(user, "@bot: operable:bundle install heroku 0.0.4")
+    assert %{name: "heroku", version: "0.0.4"} = payload
```
christophermaier: Musing out loud here... it may be better for composability to have bundle-related commands return the same shape of data.

lib/cog/commands/bundle.ex
```diff
@@ -19,14 +19,10 @@ defmodule Cog.Commands.Bundle do
...
   permission "manage_commands"
 
   rule "when command is #{Cog.Util.Misc.embedded_bundle}:bundle must have #{Cog.Util.Misc.embedded_bundle}:manage_commands"
```
christophermaier: Somewhat tangential to this PR, but we should probably consider adding some `allow` rules for the `list`, `info`, and `versions` subcommands.

(Locking down installation of bundles, of course, is definitely the way to go, though 😄 )

priv/templates/embedded/bundle-install.greenbar
```diff
@@ -0,0 +1 @@
+Bundle "~$results[0].name~" version "~$results[0].version~" installed.
```
christophermaier: We should wrap this in an `~each var=$results~` tag to handle cases where people install a bunch of bundles at once:
```
for bundle in heroku pingdom twitter | bundle install $bundle
```
Otherwise, we'll just show the installation of the first bundle.

(Alternatively, you could do a table for the more-than-one bundle scenario; that might be overkill in this case, though.)

vanstee: Good point. Will do.

test/integration/commands/bundle_test.exs
vanstee: I think they do. This just uses `Bundle`'s implementation of `Poison.encode` IIRC. Should I be rendering a view instead?

lib/cog/commands/bundle.ex
vanstee: Oops. Updated the usage but missed this. Thanks.

lib/cog/bundle_registry.ex
vanstee: Good point.

test/integration/commands/bundle_test.exs
christophermaier: I think we should try to use the views when we can... that way we have one central place to do the JSON conversion, and the Cog commands look just like the API requests. Documentation for one can double as documentation for the other.

And if /when we ever get around to calling embedded commands via the API, it'll be purely a backend change, without any user-visible effects.

kevsmith:
Two questions:

1. Do we have an issue tracking this work so we don't miss it in the changelog?
2. What's the plan for installing bundles via `cogctl`?

vanstee:
Answers:

1. Nope but I'll make one.
2. There's an open PR, but it needs some work before merging. I think @christophermaier is right that we need an API endpoint in Cog to install bundles from the registry.

PR #989, opened 2016-09-27
Lookup room refactor

### LINKED ISSUE #976: Fix how redirects are specified

mpeck:
This changes the `Cog.Chat.Adapter.lookup_room` function to take a keyword list as it's second argument containing either `[id: id]` or `[name: name]`. Making the call more explicit should help simplify the lookup room logic in the adapter.

resolves #976 

TODO:
 - [x] update tests
 - [x] add some tests for redirection

### DIFF
--- a/lib/cog/chat/adapter.ex
+++ b/lib/cog/chat/adapter.ex
@@ -57,11 +57,21 @@ defmodule Cog.Chat.Adapter do
     end
   end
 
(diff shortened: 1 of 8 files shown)


mpeck:
@kevsmith 10-4 I'm working through fixing the tests that I broke now. Then I'll add the redirect tests back in.

PR #991, opened 2016-09-27
Fix slack bullets

vanstee:
Slack uses `•` rather than `*` to denote a bullet.


christophermaier:
Good find; thanks for this!

PR #994, opened 2016-09-27
serialize maps and lists when passed to echo

### LINKED ISSUE #981: Echo command crashes with non-string args
`seed '{ "this" : { "that" : "and" }, "the" : "other" }' | echo $this` fails since we're expecting args to always be strings and not json. Let's correctly serialize json args for now. Also before making this as done double check that the commands on the [trigger docs page](http://docs.operable.io/docs/developing-a-trigger-1) still work as expected after the fix.

mpeck:
echo was causing Cog to crash whenever it received a value that didn't implement `to_string`, ie maps and lists. This change serializes those types with `Poison.encode!`.

resolves #981

### DIFF
--- a/lib/cog/commands/echo.ex
+++ b/lib/cog/commands/echo.ex
@@ -8,5 +8,13 @@ defmodule Cog.Commands.Echo do
   rule "when command is #{Cog.Util.Misc.embedded_bundle}:echo allow"
 
   def handle_message(req, state),
-    do: {:reply, req.reply_to, Enum.join(req.args, " "), state}
+    do: {:reply, req.reply_to, Enum.map_join(req.args, " ", &serialize/1), state}
+
+  # We should only have to worry about maps and lists here. Anything else
+  # we might get should implement 'to_string'.
+  defp serialize(val) when is_map(val) or is_list(val),
+    do: Poison.encode!(val)
+  defp serialize(val),
+    do: to_string(val)
+
 end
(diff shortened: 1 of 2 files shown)


lib/cog/commands/echo.ex
```diff
@@ -8,5 +8,15 @@ defmodule Cog.Commands.Echo do
...
+    do: {:reply, req.reply_to, Enum.map_join(req.args, " ", &serialize/1), state}
+
+  # We should only have to worry about maps and lists here. Anything else
+  # we might get should implement 'to_string'.
+  defp serialize(map) when is_map(map),
+    do: Poison.encode!(map)
+  defp serialize(list) when is_list(list),
+    do: Poison.encode!(list)
+  defp serialize(val),
+    do: "#{val}"
```
christophermaier: I'd just call `to_string(val)` directly, instead of doing an unnecessary string interpolation.

lib/cog/commands/echo.ex
```diff
@@ -8,5 +8,15 @@ defmodule Cog.Commands.Echo do
...
 
   def handle_message(req, state),
-    do: {:reply, req.reply_to, Enum.join(req.args, " "), state}
+    do: {:reply, req.reply_to, Enum.map_join(req.args, " ", &serialize/1), state}
+
+  # We should only have to worry about maps and lists here. Anything else
+  # we might get should implement 'to_string'.
+  defp serialize(map) when is_map(map),
+    do: Poison.encode!(map)
+  defp serialize(list) when is_list(list),
```
christophermaier: If you change the signature to 

```elixir
serialize(arg) when is_map(arg) or is_list(arg)
```
you can do it in one function head instead of two. It also makes the explanatory comments above unnecessary, since they are saying exactly what the code is saying.

christophermaier:
Also, tests for this new behavior would be good.

PR #999, opened 2016-09-28
Render help output with templates

vanstee:
screen shot 2016-09-28 at 5 42 56 pm
screen shot 2016-09-28 at 5 43 09 pm
screen shot 2016-09-28 at 5 43 21 pm


lib/cog/commands/help/command_formatter.ex
```diff
@@ -1,74 +1,34 @@
 defmodule Cog.Commands.Help.CommandFormatter do
```
christophermaier: Also, I notice there are a bunch of `render_X` functions here, but I was expecting to see maybe one public function `render` that takes a command and coordinates the whole conversion, with other functions being private.

lib/cog/commands/helpers.ex
```diff
@@ -19,13 +19,14 @@ defmodule Cog.Commands.Helpers do
         bundle_name = Cog.Util.Misc.embedded_bundle
         command_name = Cog.Command.GenCommand.Base.command_name(__MODULE__)
 
-        [command_version] = bundle_name <> ":" <> command_name
+        command_version = bundle_name <> ":" <> command_name
         |> Cog.Repository.Commands.with_status_by_any_name
-        |> Cog.Repo.preload([:bundle_version, :options])
+        |> Repo.preload([:command, :bundle_version, options: :option_type])
```
christophermaier: Same issue with preloads here, too.

lib/cog/models/bundle_version.ex
```diff
@@ -55,7 +55,7 @@ defimpl Poison.Encoder, for: Cog.Models.BundleVersion do
 
     map = %{}
     |> Map.merge(Map.take(bundle, [:name]))
-    |> Map.merge(Map.take(bundle_version, [:version, :status]))
+    |> Map.merge(Map.take(bundle_version, [:version, :status, :description]))
 
     map = Map.update!(map, :version, &to_string/1)
```
christophermaier: Why not just chain this along with the rest of the calls above? Same goes for the `Poison.Encoder.Map.encode` call below, too, I suppose.

christophermaier: Along the lines of an earlier comment, do we not have any views that render this information?

test/cog/chat/slack/templates/embedded/help_command_documentation_test.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Cog.Chat.Slack.Templates.Embedded.HelpCommandDocumentationTest do
+  use Cog.TemplateCase
+
+  test "help-command-documentation template" do
+    data = %{"results" => [%{"documentation" => "big ol' doc string"}]}
+    expected = """
+    ```big ol' doc string
```
christophermaier: Weren't we moving away from wrapping this kind of stuff in a code block?

test/cog/chat/slack/templates/embedded/help_command_test.exs
```diff
@@ -2,15 +2,24 @@ defmodule Cog.Chat.Slack.Templates.Embedded.HelpCommandTest do
...
+                             "bundle" => %{"author" => "vanstee"}}]}
     expected = """
-    ```big ol' doc string
+    *Name*
+
+    test - Do a test thing
+
+    *Synopsis*
+
+    ```test --do-a-thing
```
christophermaier: Code block?

web/views/command_version_view.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Cog.V1.CommandVersionView do
   use Cog.Web, :view
 
+  alias Cog.Commands.Help.CommandFormatter
```
christophermaier: This is a sign that the CommandFormatter needs to be in the views namespace.

test/cog/chat/slack/templates/embedded/help_command_test.exs
vanstee: I think a code block looks best here. Any thoughts?

lib/cog/commands/help/command_formatter.ex
```diff
@@ -1,74 +1,34 @@
...
-
-    synopsis = [name, options, arguments]
-    |> Enum.reject(&is_nil/1)
-    |> Enum.join(" ")
-
-    """
-    SYNOPSIS
-      #{synopsis}
-    """
+  def render_subcommands(%CommandVersion{subcommands: nil}),
```
vanstee: Now that I think about it, I think this only comes up during testing when the `subcommands` key isn't provided. I'll fix that.

lib/cog/commands/help/command_formatter.ex
vanstee: Yeah I guess this is leftover design from the string formatting. It's tough because we're really only using some of this stuff, for instance the synopsis and option string generation, in the help template. It doesn't seem generic enough for a view, but I can see the argument when looking at just the subcommands json rendering for instance. Let me think about this for a bit.

lib/cog/commands/help.ex
```diff
@@ -45,28 +43,35 @@ defmodule Cog.Commands.Help do
...
+            rendered = Cog.V1.BundleVersionView.render("bundle_version.json", %{bundle_version: bundle_version})
+            {:ok, {:bundle, rendered}}
         end
       [bundle_name, command_name] ->
         full_command_name = bundle_name <> ":" <> command_name
 
         command_version = full_command_name
         |> Commands.with_status_by_any_name
-        |> Repo.preload([:bundle_version, options: :option_type])
+        |> Repo.preload([:command, :bundle_version, options: :option_type])
```
vanstee: Hrm. Always doing a bunch of preloads sounds like a bad idea to me, since it makes the repo call less reusable in other cases where we don't want to preload. Would a chainable `Commands.preloads_for_help` function be a good compromise?

web/views/command_version_view.ex
vanstee: Do you mean something like `Cog.V1.CommandVersionView.CommandFormatter`? Feels weird to put that in a place that's so api route specific.

lib/cog/models/bundle_version.ex
vanstee: Ah I think that was committed by mistake. I'll back that change out. I think we can remove this implementation but I'm not entirely sure.

lib/cog/commands/help.ex
christophermaier: That's a nice idea, actually.

test/cog/chat/slack/templates/embedded/help_command_test.exs
christophermaier: Is the synopsis block intended for just code, then?

web/views/command_version_view.ex
christophermaier: Not sure exactly where it needs to ultimately reside, but it feels even weirder for that view to be reaching into the Cog command-specific code hierarchy for its implementation.

test/cog/chat/slack/templates/embedded/help_command_test.exs
vanstee: Yup, it's generated from available options + a arguments string.

vanstee:
@christophermaier Ended up creating a entirely new view to put all those format functions. Seems like the right place for it.

christophermaier:
@vanstee sweet!

PR #1006, opened 2016-09-29
update bundles in place when in dev mode

### LINKED ISSUE #969: Support updating config.yaml in place
Put bluntly, command developers need to be able to iterate quickly when building new bundles or they'll find better uses for their time.

A major obstacle in the Cog development experience is the requirement to bump the bundle version in config.yaml, upload the new version, and then assign it to a relay group. This adds several steps and many minutes of latency to what should be a fairly quick process.

Cog should support updating an existing config.yaml "in place" such that the changes take effect immediately (or as quick as possible). There are a couple of questions to answer:

* How will Relay detect changes? Currently Relay's change detection relies entirely on the bundle version. Relay should continue to use bundle versions but also supplement with another recency check. Perhaps Cog could include a checksum of `config.yaml` in addition to the markup when Relay requests updates?
* Should this ability always be available or should it require a `-dev` switch on Cog?

Supports roadmap issue #882

mpeck:
Was having a hard time explaining what I was thinking so I thought I would just do it. This adds an extra config that causes bundle installation to overwrite the existing version if it already exists.

resolves #969

### DIFF
--- a/lib/cog/repository/bundles.ex
+++ b/lib/cog/repository/bundles.ex
@@ -35,6 +35,9 @@ defmodule Cog.Repository.Bundles do
 
   @permanent_site_bundle_version "0.0.0"
 
+  @install_types [:normal, :force]
+  @default_install_type :normal
+
   ########################################################################
   # CRUD Operations
 
@@ -48,10 +51,11 @@ defmodule Cog.Repository.Bundles do
       #{inspect @reserved_bundle_names}
 
   """
-  def install(%{"name" => reserved}) when reserved in @reserved_bundle_names,
+  def install(install_type \\ @default_install_type, params)
+  def install(_, %{"name" => reserved}) when reserved in @reserved_bundle_names,
     do: {:error, {:reserved_bundle, reserved}}
-  def install(params),
-    do: __install(params)
+  def install(install_type, params) when install_type in @install_types,
+    do: __install(install_type, params)
 
   def install_from_registry(bundle, version) do
     case BundleRegistry.get_config(bundle, version) do
@@ -67,8 +71,12 @@ defmodule Cog.Repository.Bundles do
 
   # TODO: clean this up so we only need the config file part;
   # everything else is redundant
-  defp __install(%{"name" => _name, "version" => _version, "config_file" => _config}=params) do
-    case install_bundle(params) do
+  defp __install(install_type \\ @default_install_type, params)
+  defp __install(install_type, %{"name" => _name,
+                                 "version" => _version,
+                                 "config_file" => _config}=params)
+                             when install_type in @install_types do
+    case install_bundle(install_type, params) do
       {:ok, bundle_version} ->
         {:ok, preload(bundle_version)}
       {:error, _}=error ->
@@ -634,7 +642,7 @@ defmodule Cog.Repository.Bundles do
     Repo.preload(bundle, [:permissions, :commands])
   end
 
-  defp new_version!(bundle, params) do
+  defp new_version!(:normal, bundle, params) do
     params = Map.merge(params, Map.take(params["config_file"], ["description", "long_description", "author", "homepage"]))
 
     bundle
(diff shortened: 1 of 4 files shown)


lib/cog/repository/bundles.ex
```diff
@@ -631,6 +631,40 @@ defmodule Cog.Repository.Bundles do
...
+        # Create a new version if there isn't one
+        new_version!(bundle, params)
+      old_version ->
+        # If there is a version update it and then delete any commands and
+        # templates associated with it.
+        new_version = update_version!(old_version, params)
+        |> Repo.preload(:templates)
+
+        Enum.each(new_version.commands, &Repo.delete!/1)
+        Enum.each(new_version.templates, &Repo.delete!/1)
```
christophermaier: Why are you deleting the commands and templates of the thing you just installed?

test/controllers/v1/bundles_controller_test.exs
```diff
@@ -136,6 +136,28 @@ defmodule Cog.V1.BundlesControllerTest do
...
+
+    bundle_version = Cog.Repository.Bundles.version(bundle_version_id)
+
+    assert conn.status == 201
+    assert bundle_version.config_file == config
+    assert bundle_version.bundle.name == config["name"]
+
+    Application.put_env(:cog, :enforce_bundle_version, true)
+  end
+
```
christophermaier: I'm concerned that there are a lot of edge cases in overall behavior that we're not addressing with this test. I'm thinking in particular of the rules and permissions associated with the old and new bundles.

lib/cog/repository/bundles.ex
mpeck: The old commands and templates are removed here and they are reinstalled later in the transaction. We could go through and update all of the associations instead of just reinstalling, but that is a lot more logic and seemed more likely to introduce bugs. With this method bundle installation continues just as it did normally with the exception that we reference the old bundle version instead of a newly created one.

test/controllers/v1/bundles_controller_test.exs
mpeck: Yeah, I agree. I just did a quick smoke test to validate it was working. If you think my general approach is reasonable, I'll add some more tests to make sure everything is working as expected.

lib/cog/repository/bundles.ex
christophermaier: That's not clear from this... you're calling `delete` on the commands and templates from `new_version`.

christophermaier: Actually, the functionality should be more like "delete the existing bundle at version X, and then install this new "version" of version X". We already handle the associations properly that way.

However, for it to be truly transparent, you're also going to have to deal with enabled / disabled status.

mpeck: Hrm, that would be cleaner. I did it like this to preserve the id of the bundle version. That way we wouldn't have to go through and clean up references to it. But as I'm looking again you're right. The only one to worry about is the enabled version. Everything else references the bundle id.

Let me try that and see how if everything works like I'm thinking.

kevsmith:
Hmmm. I envisioned this feature working per-bundle not a global switch for the Cog install.

Something like:

`cogctl bundle install --force-overwrite path/to/config.yaml`

mpeck:
@kevsmith I can do that. In that case would it be a feature that's always available, or should we require a setting in Cog to enable it?

kevsmith:
Always available. Think of it like `git push --force`.

mpeck:
@kevsmith @christophermaier how does this look?

and as a sister PR, cogctl support https://github.com/operable/cogctl/pull/93

christophermaier:
Looks good. Squash the commits down and :shipit:

kevsmith:
@mpeck In the case of `install_type == :force` don't we need to wrap the delete and insert with a transaction?

mpeck:
`new_version!/3` is called inside a transaction so I didn't think it was necessary. That's the only place we call it.

kevsmith:
@mpeck Disregard. I missed the enclosing transaction on my first read thru.

kevsmith:
One other question -- what does the error look like if a forced install fails?

mpeck:
@kevsmith The installation path is basically identical as normal. The only difference being that we delete the old version before installing the new one. I tried to keep everything as similat as I could. So installation errors are the same regardless.

kevsmith:
Ok. The reason I ask is because we had issues in the past with errors inside a function running in a transaction (I think in rules ingestion) generating pretty cryptic error messages.

I agree w/ @christophermaier

PR #1008, opened 2016-10-03
use the slack id for the mention name instead of the user handle

### LINKED ISSUE #944: Cog should use user id references when responding to users
This may have been due to the recent adapter changes.

mpeck:
Returning the slack id of the user instead of the handle causes Slack to treat the output more predictably. Characters around the mention name don't cause Slack to not recognize the user's handle.

resolves #944 

TODO:
 - [x] Fix tests

### DIFF
--- a/lib/cog/chat/slack/provider.ex
+++ b/lib/cog/chat/slack/provider.ex
@@ -17,6 +17,17 @@ defmodule Cog.Chat.Slack.Provider do
     GenServer.start_link(__MODULE__, [config], name: __MODULE__)
   end
 
+  def mention_name(handle) do
+    # Lookup the user and return the id properly formatted instead of just the
+    # handle. If we just return the handle slack doesn't always recognize the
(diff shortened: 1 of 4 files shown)


lib/cog/chat/slack/provider.ex
```diff
@@ -17,6 +17,17 @@ defmodule Cog.Chat.Slack.Provider do
...
   end
 
+  def mention_name(handle) do
+    # Lookup the user and return the id properly formatted instead of just the
+    # handle. If we just return the handle slack doesn't always recognize the
+    # handle and alert the user. For example, when you return the mention name
+    # followed by a colon.
+    case lookup_user(handle) do
+      %Cog.Chat.User{id: id} -> "<@#{id}>"
+      _ -> "@" <> handle
```
vanstee: Not really a big deal, but we could call `super` here to use the default implementation.

PR #1009, opened 2016-10-04
Add PowerShell script to simplify example setup process on Windows

hershi:
The script would prompt the user for all required inputs, then it would
do all the steps mentioned in the installation guide automatically.


kevsmith:
@hershi Thanks!

PR #1010, opened 2016-10-05
HipChat Support

### LINKED ISSUE #968: HipChat support
We need to re-implement HipChat support in terms of the new chat provider API.

kevsmith:
This PR reinstates HipChat support on top of the new chat provider API.

- Full support for DMs, rooms, and private rooms
- Supports both `@` mentions and spoken commands
- Full support for Greenbar templates
- Replaced `AdapterCase` and provider-specific helper modules with `ProviderCase` and provider-specific stateful testing client to simplify integration testing. Tests will wait for an appropriate reply from Cog instead of polling chat history.

Fixes #968


kevsmith:
I tried reordering my commits to make it easier to review.  Clearly I broke something doing so....

PR #1011, opened 2016-10-05
Adds an optional config section to the help ouput for bundles

### LINKED ISSUE #948: Include config section in bundle docs for env vars
Man pages generally have a config section for env vars or general configuration. Since our bundles have similar needs, let's include a section like that for bundles when there's a clear need.

mpeck:
This adds an optional config section to the help output for bundles. The section contains notes and a list of env vars. Both notes and env vars are optional.

Note that I didn't add the config section info to the bundle version model. I'm just pulling straight from the config file in the template. Since the data was already available in the config file and it's not used anywhere else right now, I figured that was the easiest solution.

Example config.yaml excert:
```
...
config:
  notes:
    The cfn bundle makes use of CloudFormation stack templates and stack policies that are
    defined in JSON documents and stored in pre-defined S3 locations. These locations are
    defined with the 'CFN_TEMPLATE_URL' and 'CFN_POLICY_URL' configuration variables.
  env:
    - var: AWS_REGION
    - var: AWS_ACCESS_KEY_ID
    - var: AWS_SECRET_ACCESS_KEY
    - var: AWS_STS_ROLE_ARN
      description: STS role ARN that should be assumed. Defined as 'arn:aws:iam::<account_number>:role/<role_name>'.
    - var: CFN_TEMPLATE_URL
      description: S3 Location of your stack templates. Defined as 's3://<bucket>/<path>'.
    - var: CFN_POLICY_URL
      description: S3 Location of your stack policies. Defined as 's3://<bucket>/<path>'.
...
```

Help output:
screen shot 2016-10-05 at 12 46 15 pm

resolves #948 
depends on https://github.com/operable/spanner/pull/77

### DIFF
--- a/test/cog/chat/slack/templates/embedded/help_bundle_test.exs
+++ b/test/cog/chat/slack/templates/embedded/help_bundle_test.exs
@@ -8,6 +8,13 @@ defmodule Cog.Chat.Slack.Templates.Embedded.HelpBundleTest do
                              "commands" => [%{"name" => "test-command",
                                               "description" => "does just one thing"}],
                              "author" => "vanstee",
+                             "config_file" => %{
+                               "config" => %{
+                                 "notes" => "Some notes about config",
+                                 "env" => [%{"var" => "VAR1", "description" => "description1"},
+                                           %{"var" => "VAR2"}]
+                               }
+                             },
                              "homepage" => "test-bundle.com"}]}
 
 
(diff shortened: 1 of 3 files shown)


test/cog/chat/slack/templates/embedded/help_bundle_test.exs
```diff
@@ -8,6 +8,13 @@ defmodule Cog.Chat.Slack.Templates.Embedded.HelpBundleTest do
                              "commands" => [%{"name" => "test-command",
                                               "description" => "does just one thing"}],
                              "author" => "vanstee",
+                             "config_file" => %{
```
mpeck: I'm not sure I understand. Are you saying that we should add virtual fields for the config bits? I don't think that would change the test here, would it?

PR #1014, opened 2016-10-05
Removed edited command support

### LINKED ISSUE #1013: Remove re-execution of edited commands from Cog.
Currently (at least in Slack) if you edit a previous message that contained a command, Cog will attempt to re-execute the edited command. This has some unintended consequences:

1. It may not be obvious when an edit takes place for an invocation that has scrolled off the screen what the context was for the intended execution. 
2. When Slack "unfurls" content, that manifests to the API as an edited message. In this case, you can see Cog execute a command, Slack edit the message to unfurl content, then Cog execute it again. This can happen so quickly that it appears as if Cog executed the command twice.

Let's remove this behavior, particularly now that Slack has a preference to make up/down arrow cycle through history in the entry field rather than via message editing. This was a good idea at the time but it's much more preferable for it to be as obvious as possible about the source of any command that Cog executes.

kevsmith:
Fixes #1013

### DIFF
--- a/lib/cog/chat/slack/connector.ex
+++ b/lib/cog/chat/slack/connector.ex
@@ -247,23 +247,6 @@ defmodule Cog.Chat.Slack.Connector do
     end
   end
 
-  defp annotate(%{type: "message",
-                  subtype: "message_changed",
-                  channel: channel,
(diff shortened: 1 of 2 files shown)


imbriaco:
<3 :+1:

PR #1020, opened 2016-10-06
add env var for enabling a chat provider - WIP

mpeck:
@kevsmith I may be oversimplifying but could we do something like this make it easier to enable providers?

Adds an env var `COG_CHAT_PROVIDER` that can be set to "slack" or "hipchat" to enable either provider. Defaults to "slack".

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -10,6 +10,4 @@ config :comeonin,
   bcrypt_log_rounds: 14
 
 config :cog, Cog.Chat.Adapter,
-  providers: [slack: Cog.Chat.Slack.Provider],
-  chat: :slack,
   cache_ttl: {60, :sec}
--- a/config/config.exs
+++ b/config/config.exs
@@ -25,6 +25,10 @@ config :cog, :embedded_bundle_version, "0.13.0"
 config :cog, Cog.Chat.Http.Provider,
   foo: "blah"
 
+config :cog, Cog.Chat.Adapter,
+  providers: provider_list(),
+  chat: enabled_chat_provider()
+
 config :cog, :enable_spoken_commands, ensure_boolean(System.get_env("ENABLE_SPOKEN_COMMANDS")) || true
 
 config :cog, :message_bus,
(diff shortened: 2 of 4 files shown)


kevsmith:
I'm OK w/this approach but I think `enabled_providers` is unnecessary since we could get the same effect from building the `providers` list from env vars.

mpeck:
What would that var look like? `COG_CHAT_PROVIDERS=slack,http`?

kevsmith:
How about a function that inspects `COG_SLACK_ENABLED` and `COG_HIPCHAT_ENABLED` and builds the appropriate list?

mpeck:
That could work. We would have to add a var for every provider, but that's probably not a big deal. Would we worry about enforcing only having one enabled for now?

mpeck:
eh, now that I think about it, we would have to update the config for every provider anyway. So that's kind of a null point.

mpeck:
@kevsmith thought's on this? You can supply `COG_HIPCHAT_ENABLED` or `COG_SLACK_ENABLED`. I'm also using those two vars to set the `:chat` and `providers` fields.

If neither var is specified we default to the first one in the chat providers list and log a warning. I'm also enforcing the one provider at a time thing. If both vars are set we log a warning and start up only one.

config/prod.exs
```diff
@@ -10,6 +10,6 @@ config :comeonin,
   bcrypt_log_rounds: 14
 
 config :cog, Cog.Chat.Adapter,
-  providers: [slack: Cog.Chat.Slack.Provider],
-  chat: :slack,
-  cache_ttl: {60, :sec}
+  providers: provider_list(),
+  chat: enabled_chat_provider(),
+  cache_ttl: {60, sec}
```
christophermaier: This change is wrong; I fixed this `sec` -> `:sec` earlier this morning

kevsmith:
I like these changes and I think they can be simplified a bit further. Let's not make any provider the default and instead raise an error if none are enabled. We can handle setting a default provider higher up in the stack -- like in `docker-compose.yml` -- and avoid confusing users if they start up Cog w/no provider enabled.

christophermaier:
Since the code in `dev.exs` and `prod.exs` are the same, can that be pulled up into the main `config.exs`?

mpeck:
@christophermaier I tried that but the since we specify the cache_ttl bit in both dev and prod the config was overwritten.

I would think that it would merge instead of overwriting, but oh well.

mpeck:
Nevermind it was user error. configs do overlay.

PR #1026, opened 2016-10-07
don't raise in config when no provider is specified

mpeck:
We crash naturally in `Cog.Chat.Adapter` if there are no providers, so I just return nil if one isn't specified. I'm still raising in the config when a user specifies multiple chat adapters. I think that is still ok, and we don't have add checking logic to the adapter.

### DIFF
--- a/config/helpers.exs
+++ b/config/helpers.exs
@@ -27,16 +27,20 @@ defmodule Cog.Config.Helpers do
   end
 
   # Returns a list containing the enabled chat provider and all other providers.
-  def provider_list,
-    do: [Enum.find(@chat_providers, &enabled_provider?/1)] ++ @other_providers
+  # Or nil if no provider is specified.
+  def provider_list do
+    case Enum.find(@chat_providers, &enabled_provider?/1) do
+      nil -> nil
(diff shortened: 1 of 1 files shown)


config/helpers.exs
```diff
@@ -27,16 +27,20 @@ defmodule Cog.Config.Helpers do
   end
 
   # Returns a list containing the enabled chat provider and all other providers.
-  def provider_list,
-    do: [Enum.find(@chat_providers, &enabled_provider?/1)] ++ @other_providers
+  # Or nil if no provider is specified.
+  def provider_list do
+    case Enum.find(@chat_providers, &enabled_provider?/1) do
+      nil -> nil
```
vanstee: Would it make more sense for this to be `[]` or maybe `@other_providers` instead of `nil`?

mpeck: I set `nil` because that's what it would be if it wasn't set. And that's what the `Cog.Chat.Adapter` was expecting. We could change it to `[]` if that would be clearer, but it would also mean a change in the adapter.

vanstee: Ok cool that works.

PR #1031, opened 2016-10-10
changelog updates for 0.15.0

mpeck:


mpeck:
I wasn't sure about that one. It was technically behaving as expected. We just decided that we didn't want that behavior anymore. I can move it if you think it fits better as a bug fix though.

PR #1033, opened 2016-10-11
chat provider vars should just not be nil

mpeck:


config/helpers.exs
```diff
@@ -58,7 +58,10 @@ defmodule Cog.Config.Helpers do
   defp enabled_provider?({provider, _module}) do
     provider_var(provider)
     |> System.get_env
-    |> ensure_boolean
+    |> is_set?
   end
 
+  defp is_set?(nil), do: false
+  defp is_set?(_), do: true
```
christophermaier: Could just do

```elixir
defp is_set?(x): not(is_nil(x))
```

PR #1047, opened 2016-10-13
Kick off a cogctl build at the end of a cog build

### LINKED ISSUE #1035: Trigger builds of cogctl when cog builds
`cogctl`'s build process depends on `cog` (it requires a running Cog server). Ideally, anytime a new build of `cog` ran, a new `cogctl` run would also kick off, ensuring that any changes to `cog` that break `cogctl` would be quickly caught, instead of waiting for a new change to go through for `cogctl`.

Our Buildkite build for `cog` can kick off a [dependent build](https://github.com/buildkite/dependent-pipeline-example) using the Buildkite API. (In a perfect world, this would be driven from the `cogctl` side of things, but alas, it is not a perfect world.)

Whenever `cog` builds, we should kick off a new `cogctl` build. That build should use a branch of the same name as what's building in `cog`, if it exists, or fall back to `master` otherwise. The version of `cog` that the `cogctl` build uses should be whatever just built for `cog`. For now, we don't need to wait to see what the result of this job is in the `cog` job. This branch information can be passed in the API call.

We will need to set up a different bot token for the `cogctl` builds, though; otherwise, integration tests for `cog` can fail if the `cogctl` tests are running at the same time, because we'll end up with two Cog servers responding.

See https://buildkite.com/docs/rest-api/builds#create-a-build.  Also, @imbriaco did a little hacking for the branch detection:

```sh
#!/bin/sh

COG_BRANCH=$(git symbolic-ref --short HEAD)

if git ls-remote --exit-code --heads https://github.com/operable/cogctl refs/heads/${COG_BRANCH} > /dev/null 2>&1
then
  COGCTL_BRANCH=${COG_BRANCH}
else
  COGCTL_BRANCH=master
fi

echo "cog:    " $COG_BRANCH
echo "cogctl: " $COGCTL_BRANCH
```

christophermaier:
Fixes #1035

### DIFF
--- a/.buildkite/hooks/post-command
+++ b/.buildkite/hooks/post-command
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -eu
+
+DEPENDENT_PROJECT='cogctl'
+TRIGGER_PIPELINE_SLUG=${DEPENDENT_PROJECT}
+
+if git ls-remote --exit-code --heads https://github.com/operable/$DEPENDENT_PROJECT refs/heads/${BUILDKITE_BRANCH} > /dev/null 2>&1
+then
+    TRIGGER_BRANCH=${BUILDKITE_BRANCH}
+else
+    TRIGGER_BRANCH='master'
+fi
+
+TRIGGER_MESSAGE=":cogops: Dependent Build Triggered from ${BUILDKITE_BUILD_URL} :cogops:"
+
+curl \
+  "https://api.buildkite.com/v2/organizations/${BUILDKITE_ORGANIZATION_SLUG}/pipelines/${TRIGGER_PIPELINE_SLUG}/builds" \
+  --request POST \
+  --fail \
+  --header "Authorization: Bearer ${BUILDKITE_API_ACCESS_TOKEN}" \
+  --header "Content-Type: application/json" \
+  --data "{
+    \"commit\": \"HEAD\",
+    \"branch\": \"${TRIGGER_BRANCH}\",
+    \"message\": \"${TRIGGER_MESSAGE}\",
+    \"env\": {
+      \"COG_BRANCH\": \"${BUILDKITE_BRANCH}\"
+    }
+  }"


.buildkite/hooks/post-command
```diff
@@ -0,0 +1,29 @@
...
+TRIGGER_PIPELINE_SLUG=${DEPENDENT_PROJECT}
+
+if git ls-remote --exit-code --heads https://github.com/operable/$DEPENDENT_PROJECT refs/heads/${BUILDKITE_BRANCH} > /dev/null 2>&1
+then
+    TRIGGER_BRANCH=${BUILDKITE_BRANCH}
+else
+    TRIGGER_BRANCH='master'
+fi
+
+TRIGGER_MESSAGE=":cogops: Dependent Build Triggered from ${BUILDKITE_BUILD_URL} :cogops:"
```
vanstee: Nice emoji

.buildkite/hooks/post-command
```diff
@@ -0,0 +1,29 @@
...
+  --fail \
+  -H "Authorization: Bearer ${BUILDKITE_API_ACCESS_TOKEN}" \
+  -d "{
+    \"commit\": \"HEAD\",
+    \"branch\": \"${TRIGGER_BRANCH}\",
+    \"message\": \"${TRIGGER_MESSAGE}\",
+    \"env\": {
+      \"COG_BRANCH\": \"${BUILDKITE_BRANCH}\"
+    }
+  }"
```
vanstee: Should we set a content type header since we're sending json or does buildkite not care.

christophermaier: @vanstee I've added a json header, but absent any specific `Content-Type` header, `curl` will send `Content-Type application/x-www-form-urlencoded` when you use the `-d`/`--data` options, so I'm guessing that Buildkite accepts that as well. I think we do something similar for triggers, too.

mpeck:
@christophermaier instead of kicking off cogctl what if we just add it as an additional step in the Cog pipeline? Then we could see the results in the checks on the PR.

christophermaier:
@mpeck I am going to come back and make the kick-off a step in the pipeline (once we get that defined in code), but I'm not yet sure if we want a failure in the cogctl run to block the cog run. We'll be able to revisit that easily enough, though.

PR #1057, opened 2016-10-14
build image with erlang 19.1 and elixir 1.3.4

mpeck:
Probably want to hold off on this until we get the build stuff worked out. But this updates elixir and erlang in the docker image. Unfortunately there isn't a package for erlang 19.1 on alpine:3.4, which is what the image is based on, so we're building erlang from scratch. This also has the side effect of increasing the image size by about 100MB.


mpeck:
holding off until we get the kinks worked out of the new docker base images

PR #1058, opened 2016-10-14
return an error tuple instead of nil when a slack user isn't found

### LINKED ISSUE #1049: Chat handle creation erroneous error
When trying to create a chat handle that is not present in the slack organisation the following error is returned:

```
cogctl chat-handles create --user user --handle user --chat-provider slack              
** (FunctionClauseError) no function clause matching in Access.fetch/2
   (elixir) lib/access.ex:147: Access.fetch("Server internal error", "errors")
   (elixir) lib/access.ex:179: Access.get/3
   (cogctl) lib/cogctl/actions/chat_handles/create.ex:22: Cogctl.Actions.ChatHandles.Create.do_create/2
   (cogctl) lib/cogctl.ex:20: Cogctl.main/1
   (elixir) lib/kernel/cli.ex:76: anonymous fn/3 in Kernel.CLI.exec_fun/2 
```

A more descriptive error should be returned.

mpeck:
If a slack user wasn't found we returned `nil` instead of an `{:error, :not_found}` tuple. This caused Cog to crash when calling `User.from_map/1` in `Cog.Chat.Adapter`. HipChat already returns a tuple so everything is groovy there.

resolves #1049

### DIFF
--- a/lib/cog/chat/slack/connector.ex
+++ b/lib/cog/chat/slack/connector.ex
@@ -169,15 +169,19 @@ defmodule Cog.Chat.Slack.Connector do
   end
 
   defp lookup_user(value, users, [by: :id]) do
-    case Enum.reduce_while(users, nil, &(user_by_id(value, &1, &2))) do
-      nil -> {:error, :not_found}
-      user -> {:ok, user}
+    case Map.fetch(users, value) do
+      {:ok, user} ->
+        {:ok, make_user(user)}
+      :error ->
+        {:error, :not_found}
     end
   end
   defp lookup_user(value, users, [by: :handle]) do
-    case Enum.reduce_while(users, nil, &(user_by_name(value, &1, &2))) do
-      nil -> {:error, :not_found}
-      user -> {:ok, user}
+    case Map.values(users) |> Enum.find(&(&1.name == value)) do
+      nil ->
+        {:error, :not_found}
+      user ->
+        {:ok, make_user(user)}
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/cog/chat/slack/connector.ex
```diff
@@ -169,15 +169,19 @@ defmodule Cog.Chat.Slack.Connector do
...
+        {:ok, make_user(user)}
+      :error ->
+        {:error, :not_found}
     end
   end
   defp lookup_user(value, users, [by: :handle]) do
-    case Enum.reduce_while(users, nil, &(user_by_name(value, &1, &2))) do
-      nil -> {:error, :not_found}
-      user -> {:ok, user}
+    case Map.values(users) |> Enum.find(&(&1.name == value)) do
```
vanstee: :+1:

lib/cog/chat/slack/connector.ex
```diff
@@ -169,10 +169,16 @@ defmodule Cog.Chat.Slack.Connector do
   end
 
   defp lookup_user(value, users, [by: :id]) do
-    Enum.reduce_while(users, nil, &(user_by_id(value, &1, &2)))
+    case Enum.reduce_while(users, nil, &(user_by_id(value, &1, &2))) do
+      nil -> {:error, :not_found}
+      user -> {:ok, user}
+    end
```
mpeck: Good suggestion @vanstee. Thanky sir.

PR #1062, opened 2016-10-16
Dockerize CI Erlang 19.1 Elixir 1.3.4

mpeck:
Following @christophermaier's POC with dockerizing spanner, I did a quick POC to see how this would work with the erlang elixir upgrade.


mpeck:
holding off on upgrading erlang and elixir until we get the docker bits worked out

PR #1070, opened 2016-10-18
Make first reaping in the token reaper configurable

### LINKED ISSUE #1061: Intermittent test failures
Some of the integration tests are failing intermittently in CI. It looks like timeouts in some of the HipChat tests.

mpeck:
This makes the reaping of the token reaper configurable. Previously it was hard coded to 5000 milliseconds. The hard coded value caused a race condition during testing. Connections to the DB must be checked out or allowed before they can be used. Since reaping happens independently it would sometimes attempt to reap before the connection had been checked out. This caused the token reaper to crash randomly.

Tests are running more reliably in CI. There is a bit of "noise" in the integration tests. We get some red whenever the `Cog.Chat.Adapter` process is killed, but the tests still pass. Probably worth another pass to see if there is a better way to load the proper chat adapter, but I think this PR is good for now.

resolves #1061

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -16,7 +16,8 @@ config :cog, Cog.Chat.Test.Provider,
   verbose: true
 
 config :cog, Cog.Repo,
-  pool: Ecto.Adapters.SQL.Sandbox
+  pool: Ecto.Adapters.SQL.Sandbox,
+  ownership_timeout: 60000
 
 config :cog,
   :template_cache_ttl, {1, :sec}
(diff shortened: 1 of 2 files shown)


kevsmith:
I introduced the 5s delay attempting to avoid the annoying "someone didn't check out a DB connection correctly" errors Ecto throws when we restart Cog during test suite runs.

I think it's worth identifying the cause of those errors and fixing it and removing the initial delay entirely.

PR #1081, opened 2016-10-20
Include long_description in output for help command

vanstee:


kevsmith:
Is long_description optional? What happens when we render a nil field?

vanstee:
Yes; it's rendered as `null`.

PR #1085, opened 2016-10-20
Docker optimizations

mpeck:
Contains some optimizations for running the various tests in parallel and also adds a script to wait for postgres to start up before running tests.


scripts/wait-for-it.sh
```diff
@@ -48,9 +54,15 @@ wait_for_wrapper()
...
+        # Modified to work with Alpine Linux
+        # Original command
+        # timeout $TIMEOUT $0 --quiet --child --host=$HOST --port=$PORT --timeout=$TIMEOUT &
+        timeout -t $TIMEOUT -s KILL $0 --quiet --child --host=$HOST --port=$PORT --timeout=$TIMEOUT &
     else
-        timeout $TIMEOUT $0 --child --host=$HOST --port=$PORT --timeout=$TIMEOUT &
+        # Modified to work with Alpine Linux
+        # Original command
+        # timeout $TIMEOUT $0 --child --host=$HOST --port=$PORT --timeout=$TIMEOUT &
+        timeout -t $TIMEOUT -s KILL $0 --child --host=$HOST --port=$PORT --timeout=$TIMEOUT &
```
mpeck: I had to modify `wait-for-it` in order to get the timeout bits working. Since we use alpine linux for builds and for CI I just modified the original and made some notes.

PR #1089, opened 2016-10-21
Dedupe required option names before checking presence

### LINKED ISSUE #1088: Required options don't work w/short option names
screen shot 2016-10-20 at 4 56 20 pm

kevsmith:
This PR fixes Cog's evaluation of required options.

Prior to this fix the `required_set` of options Cog used to check a given invocation would contain duplicate entries for required options which also defined a short flag. This caused false negatives unless the user specified the option twice.

Fixes #1088

### DIFF
--- a/lib/cog/command/option_interpreter.ex
+++ b/lib/cog/command/option_interpreter.ex
@@ -189,17 +189,17 @@ defmodule Cog.Command.OptionInterpreter do
     do: {:error, error_msg(:type_error, value, type)}
 
   defp check_required_options(defs, opts) do
-    required_set = Enum.filter_map(defs,
-                                   fn({_,o_def}) -> o_def.required end,
-                                   fn({o_name,_}) -> o_name end) |> MapSet.new
+    required_set = defs
+                   |> Enum.filter_map(fn({_, optdef}) -> optdef.required end, fn({_, optdef}) -> optdef.name end)
+                   |> Enum.uniq
+                   |> MapSet.new
+    input_set = opts |> Map.keys |> MapSet.new
 
-    existing_set = Enum.map(opts, fn({o_name, _}) -> o_name end) |> MapSet.new
-
-    case MapSet.subset?(required_set, existing_set) do
+    case MapSet.subset?(required_set, input_set) do
       true ->
         :ok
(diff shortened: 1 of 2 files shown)


lib/cog/command/option_interpreter.ex
```diff
@@ -189,17 +189,17 @@ defmodule Cog.Command.OptionInterpreter do
     do: {:error, error_msg(:type_error, value, type)}
 
   defp check_required_options(defs, opts) do
-    required_set = Enum.filter_map(defs,
-                                   fn({_,o_def}) -> o_def.required end,
-                                   fn({o_name,_}) -> o_name end) |> MapSet.new
+    required_set = defs
+                   |> Enum.filter_map(fn({_, optdef}) -> optdef.required end, fn({_, optdef}) -> optdef.name end)
+                   |> Enum.uniq
```
christophermaier: Do you need to unique them before putting them into a set?

lib/cog/command/option_interpreter.ex
```diff
@@ -189,17 +189,17 @@ defmodule Cog.Command.OptionInterpreter do
...
+                   |> MapSet.new
```
vanstee: Might be nice to treat this as a keyword list.

```
defs
|> Keyword.values
|> Enum.filter_map(&(&1.required), &(&1.name))
|> MapSet.new
```

lib/cog/command/option_interpreter.ex
```diff
@@ -189,17 +189,17 @@ defmodule Cog.Command.OptionInterpreter do
...
 
-    existing_set = Enum.map(opts, fn({o_name, _}) -> o_name end) |> MapSet.new
-
-    case MapSet.subset?(required_set, existing_set) do
+    case MapSet.subset?(required_set, input_set) do
       true ->
         :ok
       false ->
-        missing = MapSet.difference(required_set, existing_set) |> MapSet.to_list
+        missing = MapSet.difference(required_set, input_set) |> MapSet.to_list
```
vanstee: I don't think you need to `to_list` this since `Enun.join` should work with a `MapSet`.

PR #1090, opened 2016-10-24
Use our Elixir base image

christophermaier:


mpeck:
I think you'll need to rebase to get the updates to the pipeline.yaml. But, looks good to go :+1:

PR #1098, opened 2016-10-26
Pick up elixir-slack fix for stale channel info

### LINKED ISSUE #1096: Triggers should be able to redirect to newly created rooms
There's a bug in how redirects are handled in trigger pipelines. If you startup the bot, create a new room and then create a trigger that redirects to that room, executing the trigger will fail saying the redirect is invalid. However, if you then restart the bot (and the channel exists when the bot starts) the trigger will execute successfully as we expect. There must be an issue where rooms are cached on startup and then looked up from that cache without being updated when a new channel is created. This may be a general issue with all redirects (not just triggers) but I haven't tested that yet.

kevsmith:
Fixes #1096

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -101,7 +101,7 @@ defmodule Cog.Mixfile do
       {:phoenix_html, "~> 2.6"},
       {:poison, "~> 2.0"},
       {:postgrex, "~> 0.11.2"},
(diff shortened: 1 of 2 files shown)


vanstee:
:+1:

PR #1099, opened 2016-10-26
Allow lists to bind to options of type list

### LINKED ISSUE #1097: Consider allowing arrays to be bound to list options.
Command options can be of type `list` which means that they can either be specified multiple times or can be passed a comma separated string in order to set more than one value for the option. 

In practice that means we write commands like the following, which are equivalent:

```
imbriaco:optdebug -l foo1 -l foo2 -l foo3 ...
imbriaco:optdebug -l foo1,foo2,foo3 ...
```

It would be nice if you could also directly bind an array, e.g.

```
seed '{ "foo": [ "foo1", "foo2", "foo3" ]}' | imbriaco:optdebug -l $foo env
```

Attempting this currently results in this:

slack - operable 2016-10-26 12-04-40

kevsmith:
What it says on the tin.

Fixes #1097

### DIFF
--- a/lib/cog/command/option_interpreter.ex
+++ b/lib/cog/command/option_interpreter.ex
@@ -185,6 +185,8 @@ defmodule Cog.Command.OptionInterpreter do
     |> Enum.map(&String.replace(&1, ~r/\\,/, ","))
     {:ok, splits}
   end
+  defp coerce_value("list", value) when is_list(value),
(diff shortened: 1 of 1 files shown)


imbriaco:
This changes everything. Looks great. Huge :+1: from me, pending green CI response.

slack - operable 2016-10-26 13-20-11

PR #1100, opened 2016-10-26
add simple healthcheck

### LINKED ISSUE #939: Check that $COG_HOST is accessible from the containers
When the `$COG_HOST` environment value is not set correctly, a lot of cryptic errors start to show up like cogcmd/format#11. It would be better to have a health check performed at startup to make sure that random commands are not going to fail later.

mpeck:
Adds a simple script to check that the api, trigger and service urls are available and runs it as a healthcheck for docker. Hopefully will help in diagnosing some issues folks might have with the docker-compose setup.

resolves #939


kevsmith:
I'm not sure this tests exactly what we want since we remap addresses and ports via docker-compose.

I think the gist of the ticket is that we'd like to verify Cog is reachable from outside. The simplest thing I can think of reuses your healthcheck script and runs it on the Relay node. We'd need to duplicate the relevant $COG_ environment variables in Relay's config so the appropriate values are available to the script. We'd also want to add warning comment around both sets of env vars to the effect of "these values must be kept in sync".

mpeck:
@kevsmith what are your thoughts about this? I added a `docker-compose.common.yml` file and then extended the cog and relay services with it. That should make it easier to keep things in sync.

I also updated the healthcheck script to fail when "localhost" or "127.0.0.1" is used.

Corresponding healthcheck in relay https://github.com/operable/go-relay/pull/56

imbriaco:
The `common` config seems reasonable. Have you tested that it works as expected and that Docker correctly reports health status from the healthcheck?

mpeck:
@imbriaco yup I've tested it and it seems to work fine. Any thoughts on the interval btw? It's set to 30 seconds which is the default.

mpeck:
```
➜  go-relay git:(peck/healthcheck) docker ps
CONTAINER ID        IMAGE                   COMMAND                  CREATED             STATUS                    PORTS                                                                             NAMES
3fdf3a98fa81        operable/relay:latest   "/usr/local/bin/relay"   53 seconds ago      Up 51 seconds (healthy)                                                                                     cog_relay_1
9415e4697578        operable/cog:latest     "/home/operable/cog/s"   53 seconds ago      Up 52 seconds (healthy)   0.0.0.0:4001-4002->4001-4002/tcp, 0.0.0.0:32771->1883/tcp, 0.0.0.0:80->4000/tcp   cog_cog_1
24e1f530dcba        postgres:9.5            "/docker-entrypoint.s"   54 seconds ago      Up 53 seconds             5432/tcp                                                                          cog_postgres_1
```

imbriaco:
The interval seems fine. LGTM!

PR #1101, opened 2016-10-26
Adds tee and cat commands and a supporting DataStore service

imbriaco:
* Creates a `DataStore` service based heavily on the `Memory` service, but which writes files to disk.
  * This service is not exposed via the web API and is only currently intended for use by Cog internals.
  * Data is written by default to a `service_data` directory under the top-level Cog `data` directory. This is configurable in `config.exs`.
  * File manipulation is handled by a new `NestedFile` module which creates a multi-level directory hierarchy in order to reduce number of files which will be written to a directory for performance reasons.
* Adds a `tee` command to save pipeline output using the DataStore service for later use
* Adds a `cat` command to retrieve saved pipeline output
  * Includes support for modifying the saved `tee` data with the current pipeline output in the following ways:
    * Append current output to saved data returning an array
    * Insert current output into saved map as a provided field name
    * Merge current output map into saved map

Note that this does include tests for the `cat` and `tee` commands which will pass but which are currently configured with `@moduletag :skip` to align with the other command tests.

### DIFF
--- a/lib/cog/command/service/data_store.ex
+++ b/lib/cog/command/service/data_store.ex
@@ -1,0 +1,81 @@
+defmodule Cog.Command.Service.DataStore do
+  @moduledoc """
+  Stores an arbitrary data structure for a given key. The only requirement is
+  that the data structure must be able to be encoded as JSON.
+
+  Keys may be fetched, replaced, and deleted.
+
+  The JSON data is stored on the filesystem of the Cog host. See the
+  Cog.NestedFile module for more details.
+  """
+
+  use GenServer
+
+  alias Cog.NestedFile
+
+  defstruct [:base_path]
+
+  @doc """
+  Starts the #{inspect __MODULE__} service. Accepts a path to use for the
+  base directory to store content under.
+  """
+  def start_link(base_path),
+    do: GenServer.start_link(__MODULE__, base_path, name: __MODULE__)
+
+  @doc """
+  Fetches the given key. Returns `{:ok, value}` if the key exists or `{:error,
+  :unknown_key}` if it doesn't exist.
+  """
+  def fetch(namespace, key),
+    do: GenServer.call(__MODULE__, {:fetch, namespace, key})
+
+  @doc """
+  Replaces or sets the given key with the value. Returns `{:ok, value}`.
+  """
+  def replace(namespace, key, value),
+    do: GenServer.call(__MODULE__, {:replace, namespace, key, value})
+
+  @doc """
+  Deletes the given key. Returns `{:ok, key}` when successfully
+  deleted or `{:error, :unknown_key}` if it doesn't exist.
+  """
+  def delete(namespace, key),
+    do: GenServer.call(__MODULE__, {:delete, namespace, key})
+
+  def init(nil),
+    do: {:stop, "Unable to start #{__MODULE__}: Data path not configured"}
+  def init(base_path) do
+    state = %__MODULE__{base_path: base_path}
+    {:ok, state}
+  end
+
+  def handle_call({:fetch, namespace, key}, _from, state) do
+    case NestedFile.fetch([state.base_path] ++ namespace, key, "json") do
+      {:ok, content} ->
+        data = Poison.decode!(content)
+        {:reply, {:ok, data}, state}
+      {:error, reason} ->
+        {:reply, {:error, reason}, state}
+    end
+  end
+
+  def handle_call({:replace, namespace, key, value}, _from, state) do
+    content = Poison.encode!(value)
+
+    case NestedFile.replace([state.base_path] ++ namespace, key, content, "json") do
+      {:ok, ^content} ->
+        {:reply, {:ok, value}, state}
+      {:error, reason} ->
+        {:reply, {:error, reason}, state}
+    end
+  end
+
+  def handle_call({:delete, namespace, key}, _from, state) do
+    case NestedFile.delete([state.base_path] ++ namespace, key, "json") do
+      :ok ->
+        {:reply, {:ok, key}, state}
+      {:error, reason} ->
+        {:reply, {:error, reason}, state}
+    end
+  end
+end
--- a/lib/cog/command/service_sup.ex
+++ b/lib/cog/command/service_sup.ex
@@ -1,6 +1,5 @@
 defmodule Cog.Command.Service.Supervisor do
   use Supervisor
-  require Logger
 
   alias Cog.Command.Service
 
@@ -12,10 +11,19 @@ defmodule Cog.Command.Service.Supervisor do
     token_monitor_table  = :ets.new(:token_monitor_table,  [:public])
     memory_table         = :ets.new(:memory_table,         [:public])
     memory_monitor_table = :ets.new(:memory_monitor_table, [:public])
+    data_path            = data_path
+
+      Application.get_env(:cog, Cog.Command.Service, [])[:data_path]
 
     children = [worker(Service.Tokens, [token_table,  token_monitor_table]),
+                worker(Service.DataStore, [data_path]),
                 worker(Service.Memory, [memory_table, memory_monitor_table])]
 
     supervise(children, strategy: :one_for_one)
+  end
+
+  def data_path do
+    Application.fetch_env!(:cog, Cog.Command.Service)
+    |> Keyword.fetch!(:data_path)
   end
 end
--- a/lib/cog/commands/cat.ex
+++ b/lib/cog/commands/cat.ex
@@ -1,0 +1,77 @@
+defmodule Cog.Commands.Cat do
+  use Cog.Command.GenCommand.Base,
+    bundle: Cog.Util.Misc.embedded_bundle
+
+  alias Cog.Command.Service.DataStore
+
+  # Note, we use the `tee` namespace because the namespace we read from
+  # must be the same one that data was written into.
+  @data_namespace [ "commands", "tee" ]
+
+  @description "Retrieve saved pipeline output"
+
+  @long_description """
+  The cat command retrieves pipeline output that was previously saved using the tee command.
+  """
+
+  @arguments "<name>"
+
+  rule "when command is #{Cog.Util.Misc.embedded_bundle}:cat allow"
+
+  option "merge", short: "m", type: "bool", required: false,
+         description: "Merge current pipeline map into saved pipeline map"
+
+  option "append", short: "a", type: "bool", required: false,
+         description: "Append current pipeline output to saved pipeline data, returning an array"
+
+  option "insert", short: "i", type: "string", required: false,
+         description: "Insert current pipeline output into saved pipeline map as the field specified for this option"
+
+  def handle_message(%{options: %{"merge" => true, "append" => true}} = req, state) do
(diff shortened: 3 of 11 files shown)


lib/cog/command/service/data_store.ex
```diff
@@ -0,0 +1,81 @@
...
+  Deletes the given key. Returns `{:ok, key}` when successfully
+  deleted or `{:error, :unknown_key}` if it doesn't exist.
+  """
+  def delete(namespace, key),
+    do: GenServer.call(__MODULE__, {:delete, namespace, key})
+
+  def init(nil),
+    do: raise "Unable to start #{__MODULE__}: Data path not configured"
+  def init(base_path) do
+    state = %__MODULE__{base_path: base_path}
```
christophermaier: Do we need to verify this path exists, is a directory, can be written to, etc? Is that done somewhere else?

lib/cog/commands/cat.ex
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def handle_message(%{args: [key], options: opts} = req, state) do
+    case DataStore.fetch(@data_namespace, key) do
+      {:ok, data} ->
+        cond do
+          opts["insert"] ->
+            handle_transform(:insert, req, data, state)
+          opts["append"] ->
+            {:reply, req.reply_to, List.wrap(data) ++ List.wrap(req.cog_env), state}
```
christophermaier: Isn't this actually "prepend"?

lib/cog/commands/cat.ex
```diff
@@ -0,0 +1,77 @@
...
+  defp handle_transform(action, req, data, state) do
+    case transform_data(action, req.cog_env, data, req.options) do
+      {:ok, result} ->
+        {:reply, req.reply_to, result, state}
+      {:error, reason} ->
+        {:error, req.reply_to, reason, state}
+    end
+  end
+
+  defp transform_data(action, [prev], curr, opts),
```
christophermaier: Might be nice to name this something like `transform_map_data`.

lib/cog/command/service/data_store.ex
imbriaco: The `NestedFile` module will try to create it if it does not exist the first time it is attempted to be written to and return an error if it is unable to. That felt like the right thing in this case since not everyone is likely to use the commands that make use of it and I wouldn't want to keep Cog from working. It also defaults to be nested under the default data directory just like logs and audit_logs so it feels pretty safe to me.

Happy to hear arguments to the contrary if you think we should do something else.

lib/cog/commands/cat.ex
imbriaco: It's appending the current pipeline output to the saved pipeline output. 

Example:

```
@ash seed '{ "foo": "value1" }' | tee foo
```
```
[
  {
    "foo": "value1"
  }
]
```
```
@ash seed '{ "foo": "value2" }' | cat -a foo
```
```
[
  {
    "foo": "value1"
  },
  {
    "foo": "value2"
  }
]
```

lib/cog/commands/cat.ex
imbriaco: Good idea. Done.

lib/cog/command/service/data_store.ex
```diff
@@ -0,0 +1,81 @@
...
+
+  @doc """
+  Deletes the given key. Returns `{:ok, key}` when successfully
+  deleted or `{:error, :unknown_key}` if it doesn't exist.
+  """
+  def delete(namespace, key),
+    do: GenServer.call(__MODULE__, {:delete, namespace, key})
+
+  def init(nil),
+    do: raise "Unable to start #{__MODULE__}: Data path not configured"
```
vanstee: I think the convention for this kind of thing is to return `{:stop, reason}`. That way `start_link` will return the correct `{:error, reason}` response.

lib/cog/commands/cat.ex
```diff
@@ -0,0 +1,77 @@
...
+  use Cog.Command.GenCommand.Base,
+    bundle: Cog.Util.Misc.embedded_bundle
+
+  alias Cog.Command.Service.DataStore
+
+  # Note, we use the `tee` namespace because the namespace we read from
+  # must be the same one that data was written into.
+  @data_namespace [ "commands", "tee" ]
+
+  @description "Save and pass through pipeline data"
```
vanstee: I think this might have been a copy-pasta'd and not updated, unless `cat` actually "saves" stuff.

test/cog/nested_file_test.exs
```diff
@@ -0,0 +1,41 @@
...
+
+  test "save and delete content to file under split path", %{base_paths: base_paths, content: content} do
+    path = Path.join(base_paths ++ ["te/st/fi/testfile01.data"])
+    key = "testfile01"
+
+    assert {:ok, ^content} = NestedFile.replace(base_paths, key, content)
+    assert File.exists?(path)
+
+    assert :ok = NestedFile.delete(base_paths, key)
+    assert !File.exists?(path)
```
vanstee: This is fine, but `refute` is handy for this kind of thing.

lib/cog/commands/tee.ex
```diff
@@ -0,0 +1,64 @@
...
+        {:reply, req.reply_to, nil, state}
+      "last"->
+        data =
+          MemoryClient.fetch(root, token, memory_key)
+          |> Enum.reject(fn(value) -> value == %{} end)
+          |> maybe_unwrap
+
+        MemoryClient.delete(root, token, memory_key)
+
+        case DataStore.replace(@data_namespace, key, data) do
```
vanstee: Might be worth calling out that existing keys are replaced in the desc.

test/commands/cat_tee_test.exs
```diff
@@ -0,0 +1,44 @@
...
+    response = send_message(user, "@bot: seed '{\"bar\":\"barval\"}' | cat -m test")
+    assert [%{foo: "fooval4", bar: "barval"}] = response
+  end
+
+  test "cat -a append input to saved content", %{user: user} do
+    send_message(user, "@bot: seed '{\"foo\":\"fooval5\"}' | tee test")
+    response = send_message(user, "@bot: seed '{\"foo\":\"fooval6\"}' | cat -a test")
+    assert [%{foo: "fooval5"},%{foo: "fooval6"}] = response
+  end
+end
```
imbriaco: This test seems to pass and doesn't seem to have anything provider specific (since there are no templates, etc) so it's not clear to me why it needs to be skipped. I was just following the pattern for the rest of the command tests. 

I think it's worth having even if it's skipped since it describes the expected behavior if nothing else.

lib/cog/commands/cat.ex
imbriaco: Fixed.

lib/cog/command/service/data_store.ex
imbriaco: Derp, yeah. Thanks.

test/cog/command/service/data_store_test.exs
```diff
@@ -0,0 +1,68 @@
...
+
+  test "fetch an unset key" do
+    assert {:error, "Object not found"} = DataStore.fetch(@namespace, "cheeseburgers")
+  end
+
+  test "replacing an unset key" do
+    foo1 = %{"foo1" => "fooval"}
+    DataStore.replace(@namespace, "foo1", foo1)
+    assert {:ok, ^foo1} = DataStore.fetch(@namespace, "foo1")
+    DataStore.delete(@namespace, "foo1")
```
christophermaier: Is this delete call (and others like it in these tests) really needed? If for some reason the assert failed, this line wouldn't be executed.

If it's important, an `on_exit` callback should be used. If it's not important, we can just chuck it 😄

test/commands/cat_tee_test.exs
christophermaier: If it can run, lets run it.

imbriaco:
Yep, anything that anyone saves with `tee` can be retrieved using `cat` by anyone else, unless they write rules to prevent it. I updated the `tee` docs with some discussion of that.

lib/cog/nested_file.ex
```diff
@@ -0,0 +1,86 @@
...
+  @moduledoc """
+  Stores an arbitrary String on the filesystem of the Cog host in a file
+  named by combining the provided key and extension. Files are written to
+  a directory hierarchy that is created by splitting the filename into
+  two character segments and joining up to three of those segments with a
+  provided list of base path segments.
+
+  Note: The key is sanitized to remove dangerous characters, but the
+  base paths and extension are not. Make sure not to use user entered
+  values for these unsafe arguments.
```
christophermaier: So is the `key` the only thing that's user-specified? The wording of the comment led me to believe otherwise.

lib/cog/command/service_sup.ex
```diff
@@ -12,8 +11,10 @@ defmodule Cog.Command.Service.Supervisor do
     token_monitor_table  = :ets.new(:token_monitor_table,  [:public])
     memory_table         = :ets.new(:memory_table,         [:public])
     memory_monitor_table = :ets.new(:memory_monitor_table, [:public])
+    data_path            = Application.get_env(:cog, Cog.Command.Service, [])[:data_path]
```
imbriaco: I updated it to use:

```
Application.fetch_env!(:cog, Cog.Command.Service)
|> Keyword.fetch!(:data_path)
```

Which should raise if either `Cog.Command.Service` or the `data_path` key under it are missing from the Cog application configuration.

lib/cog/nested_file.ex
```diff
@@ -0,0 +1,86 @@
...
+                  └── le
+                      └── myfilenamefoo.json
+
+  This directory structure is created in order to deal with the fact that
+  some filesystems demonstrate poor performance characteristics when working
+  with directories that contain a very large number of files.
+  """
+  require Logger
+
+  def fetch(base_paths, key, ext \\ "data") do
```
imbriaco: It was mostly there because I didn't know if we'd want to use `NestedFile` in other places in Cog since it's fairly generic.

test/cog/command/service/data_store_test.exs
imbriaco: It's mostly there to keep cruft from accumulating after tests since files get written to the filesystem and we don't have transactions to roll them back. I can use an `on_exit` instead.

lib/cog/nested_file.ex
imbriaco: Yeah, the key is the only thing that is user specified. That note is there for future-me who wants to reuse this for something else so that I don't make a mistake an let user-specified data into the namespace for instance.

lib/cog/nested_file.ex
kevsmith: What happens in the case of simultaneous writes to the same key?

imbriaco: Last one wins. They're serialized through the DataStore service's GenServer.

imbriaco:
I commented the command test back out. They pass locally if I run it in `dev` mode but test mode behaves strangely, probably related to the same issue with the rest of the command tests. When we fix those up, at least we have this to start from.

PR #1104, opened 2016-10-27
Add HTTP telemetry events to Cog startup

imbriaco:
Cog now sends an HTTP event to an Operable telemetry service every time it starts. The message that is sent contains the Cog version number and a unique ID for the Cog instance. Aside from this data, Operable also records the timestamp that the telemetry event was received.

This information is used by Operable in order to help understand the size of the Cog install base and the currently deployed versions. If you would like to opt out of having this telemetry data sent, you can set the `COG_TELEMETRY` environment variable to the value `false` before starting Cog.


imbriaco:
Did an in-person review with @kevsmith who gave a 👍 . Merging.

PR #1106, opened 2016-10-28
error messages

### LINKED ISSUE #1092: Improve "command not found in any bundle" error message
When installing and configuring a bundle, you may forget to enable the bundle or assign it to a relay-group. However, if you try to run a command that you haven't correctly configured, you get a confusing error message stating that the command could not be found. We need to improve these error messages so new users can figure out what went wrong without asking in chat.
- When a bundle is installed but disabled. We should explain that the bundle was installed and the command exists but it is currently not able to run because it has not been enabled, and then give them the chat command to enable it `@cog bundle enable pagerduty`
- When a bundle is installed and enabled but not assigned to a Relay group. We should explain that a bundle cannot run until it is assigned to a relay group that contains a relay, and then provide them the command to fix it `@cog relay-group member assign default pagerduty`
- When a bundle is installed, enabled, and assigned to a relay-group but still can't run. This typically happens when you've configured everything correctly but you haven't waited for the relay refresh interval to pickup the new bundle. We should explain that and tell them to wait.

mpeck:
This adds/updates a number of error messages to make it easier for users to troubleshoot issues.

resolves #1092 
- Executing a command in a bundle not assigned to a relay group
  - `Bundle 'test' is not assigned to a relay group. Assign the bundle to a relay group and try again.`
- Executing a command that only exists on a disabled relay
  - `No Cog Relays supporting the`test`bundle are currently available. There are one or more relays serving the bundle, but they are disabled. Enable an appropriate relay and try again.`
- Updates to the standard "No Cog Relays supporting..." error
  - `No Cog Relays supporting the`test`bundle are currently online. If you just assigned the bundle to a relay group you will need to wait for the relay refresh interval to pick up the new bundle.`

Provided by piper:
- Executing a command in a disabled bundle
  - `Bundle 'mybundle' is disabled. Please enable it and try running the command again.`
- Executing a command that does not exist in a bundle
  - `Bundle '#{bundle}' doesn't contain a command named '#{text}'.`

### DIFF
--- a/lib/cog/command/pipeline/executor.ex
+++ b/lib/cog/command/pipeline/executor.ex
@@ -228,9 +228,9 @@ defmodule Cog.Command.Pipeline.Executor do
     #   fail_pipeline_with_error({:disabled_bundle, bundle}, state)
     #
     case assign_relay(relays, bundle_name, version) do
-      {nil, _} ->
-        fail_pipeline_with_error({:no_relays, bundle_name}, state)
-      {relay, relays} ->
+      {:error, reason} ->
+        fail_pipeline_with_error({reason, bundle_name}, state)
+      {:ok, {relay, relays}} ->
         topic = "/bot/commands/#{relay}/#{bundle_name}/#{command_name}"
         reply_to_topic = "#{state.topic}/reply"
         req = request_for_plan(current_plan, request, user, reply_to_topic, state.service_token)
@@ -671,19 +671,23 @@ defmodule Cog.Command.Pipeline.Executor do
       # No assignment so let's pick one
       nil ->
         case Relays.pick_one(bundle_name, bundle_version) do
-          # No relays available
-          nil ->
-            {nil, relays}
           # Store the selected relay in the relay cache
-          relay ->
-            {relay, Map.put(relays, bundle_name, relay)}
+          {:ok, relay} ->
+            {:ok, {relay, Map.put(relays, bundle_name, relay)}}
+          error ->
+            # Query DB to clarify error before reporting to the user
+            if Cog.Repository.Bundles.assigned_to_group?(bundle_name) do
+              error
+            else
+              {:error, :no_relay_group}
+            end
         end
       # Relay was previously assigned
       relay ->
         # Is the bundle still available on the relay? If not, remove the current assignment from the cache
         # and select a new relay
         if Relays.relay_available?(relay, bundle_name, bundle_version) do
-          {relay, relays}
+          {:ok, {relay, relays}}
         else
           relays = Map.delete(relays, bundle_name)
           assign_relay(relays, bundle_name, bundle_version)
--- a/lib/cog/relay/tracker.ex
+++ b/lib/cog/relay/tracker.ex
@@ -120,15 +120,19 @@ defmodule Cog.Relay.Tracker do
   end
 
   @doc """
-  Return a list of relays serving the specified bundle version. If the bundle is
-  disabled, return an empty list.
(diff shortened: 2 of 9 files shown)


lib/cog/command/pipeline/executor.ex
```diff
@@ -667,27 +671,35 @@ defmodule Cog.Command.Pipeline.Executor do
...
-      relay ->
-        # Is the bundle still available on the relay? If not, remove the current assignment from the cache
-        # and select a new relay
-        if Relays.relay_available?(relay, bundle_name, bundle_version) do
-          {relay, relays}
-        else
-          relays = Map.delete(relays, bundle_name)
-          assign_relay(relays, bundle_name, bundle_version)
-        end
+    # If the bundle is assigned to a relay group carry on, otherwise bail
```
kevsmith: Flipping the logic will reduce the number of DB queries issued to execute a pipeline:

```
case Map.get(relays, bundle_name) do
  # No assignment so let's pick one
  nil ->
    case Relays.pick_one(bundle_name, bundle_version) do
      # No relays available
      {:error, :no_relays} ->
        {nil, relays}
        # relays exist, but are disabled
      {:error, :no_enabled_relays} ->
        # Query database to clarify error condition before reporting to user
        if Cog.Repository.Bundles.assigned_to_group?(bundle_name) do
          {:no_enabled_relays, relays}
        else
          {:no_relay_group, relays}
        end
      # Store the selected relay in the relay cache
      {:ok, relay} ->
        {relay, Map.put(relays, bundle_name, relay)}
    end
    # Relay was previously assigned
  relay ->
  # Is the bundle still available on the relay? If not, remove the current assignment from the cache
  # and select a new relay
    if Relays.relay_available?(relay, bundle_name, bundle_version) do
      {relay, relays}
    else
      relays = Map.delete(relays, bundle_name)
      assign_relay(relays, bundle_name, bundle_version)
    end
end
```

lib/cog/relay/tracker.ex
```diff
@@ -120,15 +120,15 @@ defmodule Cog.Relay.Tracker do
...
+  @spec relays(t, bundle_name, version) :: {[String.t], [String.t]}
   def relays(tracker, bundle_name, bundle_version) when is_binary(bundle_name) do
-    tracker.map
-    |> Map.get({bundle_name, bundle_version}, MapSet.new)
-    |> MapSet.difference(tracker.disabled)
-    |> MapSet.to_list
+    relays = Map.get(tracker.map, {bundle_name, bundle_version}, MapSet.new)
+    enabled_relays = MapSet.difference(relays, tracker.disabled)
+
+    {MapSet.to_list(relays), MapSet.to_list(enabled_relays)}
```
kevsmith: I think returning descriptive atoms instead of sets of relays would be more intention revealing.

`{[], []}` becomes `:no_relays`
`{[...], []}` becomes `:no_enabled_relays`
`{[...], [...]}` becomes `[...]`

where `[...`] indicates a non-empty list.

lib/cog/relay/tracker.ex
```diff
@@ -137,7 +141,7 @@ defmodule Cog.Relay.Tracker do
   """
   @spec is_bundle_available?(t, relay_id, bundle_name, version) :: boolean()
   def is_bundle_available?(tracker, relay, bundle_name, bundle_version) do
-    available_relays = relays(tracker, bundle_name, bundle_version)
+    {_all_relays, available_relays} = relays(tracker, bundle_name, bundle_version)
```
kevsmith: Does this work if `relays/3` returns one of the error tuples?

PR #1110, opened 2016-10-28
Upgrade to Elixir 1.3.4 / Erlang 19.1

christophermaier:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM operable/elixir:1.3.1-r4
+FROM operable/elixir:1.3.4-r0-alpha
```
kevsmith: alpha?

christophermaier: I'll update before making it final... that was just to facilitate testing.

PR #1115, opened 2016-11-01
help no longer requires the fully qualified command name

### LINKED ISSUE #1109: Allow the help command to work with non-qualified command names
We need to allow things like `help instance-create` when the `ec2` bundle is installed and contains an `instance-create` command.
- Allow help to be run with a command name only
- If more than 1 command name conflict across bundles, return an error with both as qualified command names (<bundle>:<command>)
- If no bundles or commands can be found explain that neither could be found

mpeck:
Adds support for passing just the command name without the bundle and improves error messaging a bit.

resolves #1109

### DIFF
--- a/lib/cog/commands/help.ex
+++ b/lib/cog/commands/help.ex
@@ -27,42 +27,28 @@ defmodule Cog.Commands.Help do
 
   rule "when command is #{Cog.Util.Misc.embedded_bundle}:help allow"
 
+  # When help is called with no arguments we return the list of installed bundles
   def handle_message(%{args: []} = req, state) do
     bundles = %{enabled: Repo.preload(Bundles.enabled, :bundle),
                 disabled: Repo.preload(Bundles.highest_disabled_versions, :bundle)}
     {:reply, req.reply_to, "help-bundles", bundles, state}
   end
-
-  def handle_message(%{args: [bundle_or_command]} = req, state) do
-    response = case String.split(bundle_or_command, ":") do
-      [bundle_name] ->
-        bundle_version = bundle_name
-        |> Bundles.with_status_by_name
-
-        case bundle_version do
-          nil ->
-            {:error, "Bundle #{inspect(bundle_name)} not found"}
-          %BundleVersion{config_file: config_file} ->
-            commands =
-              Enum.map(config_file["commands"], fn({name, map}) -> Map.put(map, "name", name) end)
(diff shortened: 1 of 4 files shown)


lib/cog/commands/help.ex
```diff
@@ -75,11 +59,81 @@ defmodule Cog.Commands.Help do
...
 
   def handle_message(req, state) do
     {:reply, req.reply_to, "usage", %{usage: @moduledoc}, state}
   end
+
+  # Note: We could do the command lookup first regardless of the presence of ':'
+  # This would eliminate the second case statement but would also mean that commands
+  # with the same name as bundles would take precedence.
+  defp lookup(name) do
+    # If the name contains a ':' then we just do a command lookup
```
kevsmith: We need to tweak this to accommodate git-style commands: `help ec2:instances-ls` vs `help ec2:instances ls`. Easiest way to do that is run the args thru another parse cycle to resolve the actual command name.

mpeck: oops, I didn't even think about the git style command bits. I'll fix that.

lib/cog/commands/help.ex
```diff
@@ -75,11 +61,84 @@ defmodule Cog.Commands.Help do
...
-        {:error, req.reply_to, error, state}
+        {:error, req.reply_to, error_msg(error), state}
+    end
+  end
+
+  # If multiple args are passed then we join with "-" before doing a lookup.
+  # This should account for users passing command names using the optional
+  # "git style" syntax.
+  defp lookup(args) when is_list(args) do
+    Enum.join(args, "-")
```
mpeck: @kevsmith thoughts on this? I think it's less complex and, we save both the call to the parser and the extra db query. But as a trade off users could technically do something like, `help test bundle:test command` and get the help for `test-bundle:test-command` whereas the command `test bundle:test command` would fail to parse normally.

kevsmith: We're trading consistency for implementation ease. Consistency is why I suggested reusing the parser but this is fine for now. If the bundle name becomes an issue then we can fix that.

Another thought is that you could add logic here to detect hyphenated bundle names and reject them.

mpeck:
One more try :)

@kevsmith any thoughts on this? I updated it to disallow bundle names with spaces. It's a bit funky because if the user doesn't use the fully qualified command name we don't know if they are referring to a bundle or a command. That adds some extra complexity to things. But for now, I think this is the simplest solution.

PR #1119, opened 2016-11-02
Make command tests more unit testy

mpeck:
This adds a new test case, `Cog.CommandCase` that basically just calls the `handle_message` function in the command module. Tests can now run async. 

I rewrote the `help` and `sort` tests to see how well it works with standard commands and commands that use the memory service.

depends on #1121 
part if #1122

### DIFF
--- a/test/commands/help_test.exs
+++ b/test/commands/help_test.exs
@@ -1,32 +1,25 @@
 defmodule Cog.Test.Commands.HelpTest do
-  use Cog.AdapterCase, adapter: "test"
+  use Cog.CommandCase, command_module: Cog.Commands.Help
 
-  @moduletag :skip
-
-  alias Cog.Support.ModelUtilities
-
-  setup do
-    user = user("vanstee", first_name: "Patrick", last_name: "Van Stee")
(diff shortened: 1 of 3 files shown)


christophermaier:
Why is the ExVCR removal in this PR? It looks like that's for testing the integration with warehouse for the bundle command, which is separate from the other stuff this PR is doing.

The "map a seed over a series of requests and extract the final one" pattern seems like it could be cleaned up a bit to be slightly more "unit test"-like. Perhaps some helpers that set up the test with certain expected data in the memory store, and then a single "last" command invocation to validate how it behaves? You could do similar tests to validate behavior on "first" invocations, as well as intermediate invocations.

mpeck:
@chris The ExVCR removal is part of the PR because it breaks the command tests. I can move it to it's own PR but this PR will still depend on it, that's why I just stuck it all in here. But I'm totally cool with moving it if it makes this less clear.

re: "map over a series of requests" I did that originally because I was trying to change the tests as little as possible, though mayhaps I have failed. I can revisit that bit. So your thinking that we basically test commands that use the memory service more on a per step basis then?

Issue #1122, opened 2016-11-03
update command tests


mpeck:
Most command tests are currently disabled due to some conflicts with how they interact with Cog. Even when the tests were working they tended to not be very reliable and were slow.

We need to update command tests to be actual unit tests. This should improve reliability and allow us to run tests async, which should improve speed.

PR #1124, opened 2016-11-04
Alias test rework

mpeck:
Reworks `alias` command tests. This also includes work to the `which` command tests since they are tied closely to aliases.

part of #1122

### DIFF
--- a/lib/cog/commands/alias/create.ex
+++ b/lib/cog/commands/alias/create.ex
@@ -14,7 +14,7 @@ defmodule Cog.Commands.Alias.Create do
     -h, --help  Display this usage info
 
   EXAMPLE
-    alias create my-awesome-alias "echo \"My awesome alias\""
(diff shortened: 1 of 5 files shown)


lib/cog/commands/alias/create.ex
```diff
@@ -14,7 +14,7 @@ defmodule Cog.Commands.Alias.Create do
     -h, --help  Display this usage info
 
   EXAMPLE
-    alias create my-awesome-alias "echo \"My awesome alias\""
+    alias create my-awesome-alias \"echo \"My awesome alias\"\"
```
vanstee: I don't think escaping those quotes are necessary here, although I think Elixir renders them just the same inside a triple quote block.

mpeck:
@kevsmith there is actually another set of tests that test alias execution, that's actually where I ran across the bug addressed in #1135. This is just to test the alias command itself. It actually bypasses all of the variable interpolation bits anyway.

PR #1128, opened 2016-11-04
Cat\Tee test updates

mpeck:
This also includes a couple tweaks to the command case to make it a bit less confusing to call multiple commands from a test.

part of #1122


test/commands/cat_tee_test.exs
```diff
@@ -1,44 +1,69 @@
...
+  use Cog.CommandCase, command_tag: "cat-tee"
 
-  @moduletag :skip
+  alias Cog.Commands.Tee
+  alias Cog.Commands.Cat
 
-  setup do
-    user = user("vanstee", first_name: "Patrick", last_name: "Van Stee")
-    |> with_chat_handle_for("test")
+  #setup do
```
kevsmith: Should be deleted if not used

test/support/command_case.ex
```diff
@@ -10,17 +10,28 @@ defmodule Cog.CommandCase do
     quote location: :keep do
       import unquote(__MODULE__)
 
-      @command_module Keyword.fetch!(unquote(opts), :command_module)
-      @command_name Module.split(@command_module) |> List.last |> String.downcase
+      @command_module Keyword.get(unquote(opts), :command_module)
+      @command_tag Keyword.get(unquote(opts), :command_tag)
 
-      @moduletag commands: @command_name
+      @command_name (if @command_module do
```
kevsmith: From reading the code it's not clear to me when or why each of these options should be used. An explanatory comment here would help greatly.

test/commands/cat_tee_test.exs
```diff
@@ -1,44 +1,69 @@
 defmodule Cog.Test.Commands.CatTeeTest do
-  use Cog.AdapterCase, adapter: "test"
+  use Cog.CommandCase, command_tag: "cat-tee"
```
kevsmith: What's the rationale for testing two commands in the same test module?

mpeck: The original test had them in the same module and I just copied that making the updates necessary to work in the new pattern. I can break them up though.

PR #1129, opened 2016-11-05
updates and re-enables echo and some of the smaller command tests

mpeck:
re-enables echo, min, max, unique and filter

part of #1122

### DIFF
--- a/test/commands/filter_test.exs
+++ b/test/commands/filter_test.exs
@@ -4,10 +4,11 @@ defmodule Cog.Test.Commands.FilterTest do
   test "returns an item that matches at a specific key" do
     data = %{"foo" => %{"bar" => "stuff", "baz" => "other stuff"}}
 
-    {:ok, match} = new_req(cog_env: data,
-                           options: %{"path" => "foo.bar",
-                                      "matches" => "stuff"})
-    |> send_req()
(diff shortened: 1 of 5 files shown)


test/commands/filter_test.exs
```diff
@@ -1,47 +1,40 @@
 defmodule Cog.Test.Commands.FilterTest do
-  use Cog.AdapterCase, adapter: "test"
+  use Cog.CommandCase, command_module: Cog.Commands.Filter
 
-  @moduletag :skip
+  test "can match on objects on a specific path" do
```
christophermaier: this test should probably be two separate tests

test/commands/filter_test.exs
```diff
@@ -1,47 +1,40 @@
...
-    --path="foo.bar"
-    """
-
-    response = send_message(user, command)
-
-    assert [%{foo: %{bar: "stuff",
-                     baz: "other stuff"}},
-            %{foo: %{bar: "me",
-                     baz: "now this stuff"}}] = response
+  test "filters a list of things based on a key" do
```
christophermaier: This isn't testing what the original test tested, and seems to be duplicating the `"can match on objects on a specific path"` test above.

PR #1131, opened 2016-11-05
sleep test updates

mpeck:
updates and additions to sleep tests

part of #1122

### DIFF
--- a/test/commands/sleep_test.exs
+++ b/test/commands/sleep_test.exs
@@ -1,17 +1,51 @@
 defmodule Cog.Test.Commands.SleepTest do
-  use Cog.AdapterCase, adapter: "test"
+  use Cog.CommandCase, command_module: Cog.Commands.Sleep
 
-  @moduletag :skip
+  test "basic sleeping" do
+    request = fn ->
+      new_req(cog_env: "foo", args: [3])
+      |> send_req()
+    end
 
-  setup do
-    user = user("jfrost", first_name: "Jim", last_name: "Frost")
-    |> with_chat_handle_for("test")
+    # We're running the command as a task so it doesn't block the test process
+    task = Task.async(request)
 
-    {:ok, %{user: user}}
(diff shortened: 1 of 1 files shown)


test/commands/sleep_test.exs
```diff
@@ -1,17 +1,56 @@
...
-  setup do
-    user = user("jfrost", first_name: "Jim", last_name: "Frost")
-    |> with_chat_handle_for("test")
+    # We're running the command as a task so it doesn't block the test process
+    Task.async(request)
+    # Then we can yield with a short timeout
+    |> Task.yield(1000)
+    # The timeout should expire before the command returns, resulting in nil
+    |> is_nil()
+    |> assert()
```
christophermaier: This pipeline-based assertion reads non-idiomatically.

test/commands/sleep_test.exs
```diff
@@ -1,17 +1,56 @@
...
+        |> send_req()
+      end
+
+      # We're running the command as a task so it doesn't block the test process
+      Task.async(request)
+      # The task should not sleep, but in case it does, we'll set the timeout
+      |> Task.yield(1000)
+      # The task should return before the timeout, so we shouldn't get nil this time
+      |> is_nil()
+      |> refute("Slept on the #{step} invocation step.")
```
christophermaier: This refute also is also structured confusingly.

test/commands/sleep_test.exs
christophermaier: It might be useful to send a request with a `cog_env` set, spin off a `Task`, assert that yielding to the task in a timeframe shorter than the sleep time results in `nil`, but then yield or await the task and verify that it returns whatever it was supposed to in `cog_env`. That way we can be sure that it was sleeping instead of otherwise failing silently.

Incidentally, that would also cover the "returns what it's passed" test down below.

mpeck: Yeah, you're right. I think this was one of those great ideas at midnight things :)

PR #1134, opened 2016-11-06
group test updates

mpeck:
part of #1122

### DIFF
--- a/test/commands/group_test.exs
+++ b/test/commands/group_test.exs
@@ -1,176 +1,245 @@
 defmodule Cog.Test.Commands.GroupTest do
-  use Cog.AdapterCase, adapter: "test"
+  use Cog.CommandCase, command_module: Cog.Commands.Group
 
-  @moduletag :skip
+  import Cog.Support.ModelUtilities, only: [user: 1,
+                                            group: 1,
+                                            role: 1,
+                                            add_to_group: 2]
+  describe "group subcommands" do
 
-  alias Cog.Models.Group
-  alias Cog.Repository.Groups
+    test "can't pass an unknown subcommand" do
+      {:error, error} =
+        new_req(args: ["not-a-subcommand"])
+        |> send_req()
 
-  setup do
-    user = user("belf", first_name: "Buddy", last_name: "Elf")
-    |> with_chat_handle_for("test")
-    |> with_permission("operable:manage_users")
-    |> with_permission("operable:manage_groups")
+      assert(error == "Unknown subcommand 'not-a-subcommand'")
+    end
 
-    {:ok, %{user: user}}
   end
 
-  test "adding and removing users to groups", %{user: user} do
-    response = send_message(user, "@bot: operable:group member add elves #{user.username}")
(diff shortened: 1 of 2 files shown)


test/commands/group_test.exs
```diff
@@ -1,131 +1,155 @@
...
 
-  test "listing group", %{user: user} do
+  test "listing group" do
     group("elves")
     cheer = group("cheer")
+    user = user("belf")
 
-    [response] = send_message(user, "@bot: operable:group member add cheer belf")
+    {:ok, response} = new_req(args: ["member", "add", "cheer", "belf"])
+    |> send_req()
```
christophermaier: Similarly, this is setup code. The assertions that follow are unnecessary.

mpeck:
I can rework this guy. I started doing that on some of the previous modules, but decided to just convert everything over as it is because it was taking a good bit of time. But it's probably not a bad idea to go ahead and make those improvements now.

test/commands/group_test.exs
```diff
@@ -1,176 +1,245 @@
...
 
-    [response] = send_message(user, "@bot: operable:group role add cheer admin")
-    assert response.name == "cheer"
+      {:ok, response} =
+        new_req(args: ["rename", "renamed_group", "has_been_renamed"])
+        |> send_req()
 
-    [response] = send_message(user, "@bot: operable:group info cheer")
-    assert hd(response.roles).name == "admin"
+      assert(%{name: "has_been_renamed"} = response)
```
kevsmith: I think you want to use `==` here, no?

test/commands/group_test.exs
```diff
@@ -1,176 +1,245 @@
...
-    [response] = send_message(user, "@bot: operable:group info cheer")
-    assert length(response.roles) == 0
-  end
+      {:ok, response} =
+        new_req(args: ["info", "info_group"])
+        |> send_req()
 
-  test "getting group info", %{user: user} do
-    group("cheer")
+      assert(%{name: "info_group"} = response)
```
kevsmith: Same here re: `==`

test/commands/group_test.exs
```diff
@@ -1,176 +1,245 @@
...
-    assert_error_message_contains(response, "name: has already been taken")
-  end
+      response = Enum.sort_by(response, &(&1.name))
 
-  test "errors using the group command", %{user: user} do
-    [response] = send_message(user, "@bot: operable:group create")
-    assert response.error == "Missing required argument: group_name"
+      assert([%{name: "group1"},
+              %{name: "group2"},
+              %{name: "group3"}] = response)
```
kevsmith: Same here re: `==`

test/commands/group_test.exs
mpeck: Nope `=` is correct. There is actually more to the record than just the name. There is also an id, which we don't care about, and `members` and `roles` keys that are empty.

Actually as I think about it, I should have included the members and roles to confirm that they are empty.

mpeck: The more I look at it, I think it would actually be clearer to `match?/2` instead.

PR #1140, opened 2016-11-11
update permission command tests

mpeck:
part of #1122


mpeck:
Dang it! I forgot that setup bit twice now. I'll drop it.

And yes I was planning on using `unwrap/1` and `unwrap_error/1` in the other tests. I start using them when that bit gets merged.

PR #1146, opened 2016-11-14
Expose default Cog ports in Docker image

christophermaier:


christophermaier:
Closing for now; we'll come back for it later.

PR #1163, opened 2016-11-17
Optionally add chat handle for bootstrap user

christophermaier:
When setting up a Cog system for the first time using `COG_BOOTSTRAP_*`
variables, a `COG_BOOTSTRAP_CHAT_HANDLE` variable may optionally be
set. The bootstrapping process will attempt to resolve this handle
according to the currently-active chat provider (controlled by
`COG_SLACK_ENABLED` and `COG_HIPCHAT_ENABLED`).

If the given handle is valid, it is associated with the bootstrapped
user. That user can immediately begin interacting with Cog via chat.

If the chat handle is not valid, a warning message is logged, but the
bootstrapped user is still created and the system still starts up as
usual. In this case, the user can use `cogctl` to set up a chat handle.

### DIFF
--- a/lib/cog/bootstrap.ex
+++ b/lib/cog/bootstrap.ex
@@ -167,7 +167,7 @@ defmodule Cog.Bootstrap do
         Logger.info("Associated bootstrap user with chat handle '#{handle}'")
         :ok
       {:error, reason} ->
-        Logger.error("Could not associate bootstrap user with chat handle '#{handle}': #{reason}")
+        Logger.error("Failed to associate chat handle '#{handle}' with bootstrap user; chat handle will need to be configured manually. Error: #{reason}")
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/cog/bootstrap.ex
```diff
@@ -141,6 +158,19 @@ defmodule Cog.Bootstrap do
...
+  defp maybe_bootstrap_chat_handle_from_env(_user, nil),
+    do: Logger.info("No chat handle specified for bootstrap user; skipping")
+  defp maybe_bootstrap_chat_handle_from_env(user, handle) do
+    {:ok, provider_name} = Cog.Util.Misc.chat_adapter_module()
+    case Cog.Repository.ChatHandles.set_handle(user, provider_name, handle) do
+      {:ok, _} ->
+        Logger.info("Associated bootstrap user with chat handle '#{handle}'")
+        :ok
+      {:error, reason} ->
+        Logger.error("Could not associate bootstrap user with chat handle '#{handle}': #{reason}")
```
kevsmith: I think we could stand to be more verbose here so the user knows exactly what to do.

```
Logger.error("""
Failed to create chat handle '#{handle}' for bootstrap user. Chat handle will need to be configured manually.
Error: #{reason}
""")
```

Or something like it.

Issue #1142, opened 2016-11-14
Make global error template user customizable


kevsmith:
## Overview
We've gotten several requests to make the error template user customizable. Emerging community consensus is the default error template is great for developers but contains too much information for non-technical users.

## Requirements

1. Cog admins must be able to revert back to the default template at any time.
1. Template customizations must persist across Cog restarts.
1. Documentation of the customization process, including a list of available variables, must be part of the doneness criteria.

## Constraints

1. A global customized error template is good enough for now.

mpeck:
# Proposal

I think we can solve the problem by providing a config option that specifies a path to the error template on the machine running Cog. During startup we check for the config option and replace the standard error template in the db if it is set. Vice versa if it isn't set we just set the error template to the standard one.

This shouldn't require too many changes. We just add the new config option and modify `Cog.Repository.Bundles.maybe_upgrade_embedded_bundle!/1`. It shouldn't require any changes to Cog's template rendering logic. The downside of course is that a changing the template would require a Cog restart and the ability to get the template file onto the machine running Cog. Though to be honest that doesn't seem like a terrible trade off to me.

PR #1165, opened 2016-11-17
Make common error template configurable

mpeck:
The common error template is now configurable via the env var `COG_CUSTOM_TEMPLATE_DIR`. The var should point to a directory that contains a file named `error.greenbar` that will replace the standard `error.greenbar`.

I went the directory route so we can make additional templates configurable more easily in the future. Common templates that are configurable are white listed. Currently the only configurable template is the standard error template.

Documentation to come.

part of #1142

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -74,6 +74,8 @@ config :cog, Cog.Command.Pipeline,
 config :cog, Cog.Command.Service,
   data_path: data_dir("service_data")
 
+config :cog, :custom_template_dir, System.get_env("COG_CUSTOM_TEMPLATE_DIR")
+
 # Set these to zero (0) to disable caching
 config :cog, :command_cache_ttl, {60, :sec}
 config :cog, :command_rule_ttl, {10, :sec}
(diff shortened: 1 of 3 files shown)


kevsmith:
What's the process for reverting from a custom to the stock error template?

mpeck:
@kevsmith unset `COG_CUSTOM_TEMPLATE_DIR` and restart Cog.

kevsmith:
Can we revert without requiring env var edits and bouncing Cog?

In my mind the logic for using a customized template looks like:

```
          Pipeline returns error
                      |
         Does custom template dir exist?
                      |
                      | (No)
                      +------ Use standard error template
                      |
                      | (Yes)
                      +------- Does the file error.greenbar exist?
                                            |
                                            | (No)
                                            +------- Use standard error template
                                            |
                                            | (Yes)
                                            +-------- Use custom template
```

In this schema all that's required to revert is to delete the template file or the custom template directory.

mpeck:
We could do that. I think we would have to change some of the logic in the template evaluator. This change just updates the record in the DB on start up. So all of the template rendering bits stay the same. I thought about doing something similar to what you're suggesting when I started, but figured this would be less invasive. But I can totally switch it up if you don't like the restart bit.

I'm assuming that with your method we would just read the template from the file system each time correct?

kevsmith:
Yep. We do a similar thing for dynamic config and the perf hit isn't really noticeable.

mpeck:
10-4 I'll have something for you soon.

PR #1174, opened 2016-11-22
Cleanups

christophermaier:
Clean up outdated bits of our makefile, as well as normalize / clarify some environment variables.


Makefile
```diff
@@ -1,83 +1,16 @@
...
-ci:
-	@echo "NO_CI file found. CI build targets skipped."
-	@exit 0
-endif
-
-setup:
-	mix deps.get
-	mix ecto.create
-	mix ecto.migrate
+DOCKER_IMAGE ?= operable/cog:0.5-dev
```
mpeck: Maybe something for later. But would it be better to default to something like `operable/cog:latest-dev` or something?

mix.exs
```diff
@@ -126,7 +126,8 @@ defmodule Cog.Mixfile do
   defp elixirc_paths(_),     do: ["lib", "web"]
 
   defp aliases do
-    ["ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
+    ["cog.setup": ["deps.get", "ecto.setup"],
+     "ecto.setup": ["ecto.create", "ecto.migrate"],
```
mpeck: This is awesome. I didn't even know this was here.

test/support/hipchat_client.ex
```diff
@@ -20,8 +20,8 @@ defmodule Cog.Test.Support.HipChatClient do
 
   def start_link() do
     api_token = get_env!("HIPCHAT_USER_API_TOKEN")
-    xmpp_jid = get_env!("HIPCHAT_USER_JID")
-    xmpp_password = get_env!("HIPCHAT_USER_PASSWORD")
+    xmpp_jid = get_env!("HIPCHAT_USER_JABBER_ID")
```
mpeck: Is `HIPCHAT_USER_JID` something we should deprecate, or was that never really exposed to the public?

Makefile
christophermaier: yeah, I'll fix that up while I'm here.

PR #1175, opened 2016-11-22
Remove healthcheck script

### LINKED ISSUE #1156: Support `COG_(TRIGGER|SERVICE)_URL_BASE` in docker healthchecks
When running on EC2 we only set `COG_TRIGGER_URL_BASE` and `COG_SERVICE_URL_BASE` so the current healthchecks fail. Let's support those env vars if they exist and fallback to the current `HOST` and `PORT` ones.

mpeck:
Removes the healtcheck script for now. This will be replaced in a future PR with a more explicit way for Cog admins to smoke test a Cog installation.

resolves #1156


kevsmith:
I think it's useful to take a step back and revisit the original problem we're aiming to solve and see if a better solution presents itself based on our recent experience.

We're currently trying to answer the question "can these components communicate?" Since the question is unbounded in time, which implies checking and re-checking to verify communication still works, we're using Docker's health check hook to find an answer. I'm increasingly of the mind this is overly broad and not that useful to Cog admins.

I suspect Cog admins are looking for an answer to the question "can these components communicate as they are configured right now?" which is much more specific and applicable to setting up a new Cog install, the source of the original user request.

The refined question clarifies the problem and suggests smoke testing from the point of view of various components, probably Relay to start, is a better path forward. Smoke testing tools would allow admins to sanity check their installs and troubleshoot configuration problems.

mpeck:
@kevsmith relay seems like a reasonable place to start. Are you thinking then that we would have a separate tool installed alongside relay for smoke testing, or that an admin could start relay up in a special "smoke test" or "dry run" mode?

kevsmith:
I'd say an easy place to start is a smoke test script they could `docker exec` on the Relay container.

PR #1181, opened 2016-11-28
Improved error template + HipChat attachment rendering

imbriaco:
This PR does a few things:

1. Updates the attachment rendering for HipChat to be more consistent with Slack.
2. It flips the logic that we used to select planning vs. execution errors. I believe we have had these inverted for some time and effectively every error message was being reported as "The pipeline failed planning the invocation" even when it was a command execution failure. The screenshots below illustrate the problem
3. Updated the greenbar template for errors to move the error message to the top so that it isn't hidden by a _more_ link by Slack. Also moved the display of the failed command or pipeline plan to fields within the attachment instead of rendering them to text in the body of the error, also to prevent truncation. Screenshots below.
4. Increase the wait time between Slack API requests from 750ms to 1100ms to more consistently avoid the Slack rate limit. Sadness intensifies.

**Before:**

before

**After:**

after

**HipChat:**

hipchat

### DIFF
--- a/priv/templates/common/error.greenbar
+++ b/priv/templates/common/error.greenbar
@@ -1,4 +1,13 @@
+#
+# CASE 1: COMMAND EXECUTION FAILURE
+#
+# Command execution returned an error either with or without an additional
+# planning error for failures further in the pipeline.
+#
 ~if cond=$execution_failure != ""~
+#
+# No additional pipeline stages after the failed command
+#
 ~if cond=$planning_failure == ""~
 ~attachment title="Command Execution Error" color="#ff3333" "Failed Executing"=$execution_failure Caller=$initiator Pipeline=$pipeline_text "Pipeline ID"=$id Started=$started~
 
@@ -7,6 +16,10 @@
 ```
 ~end~
 ~end~
+#
+# There are additional stages in the pipeline after the failed execution that
+# will not be executed due to the command execution failure.
+#
 ~if cond=$planning_failure != ""~
 ~attachment title="Command Execution Error" color="#ff3333" "Failed Executing"=$execution_failure "Failed Planning"=$planning_failure Caller=$initiator Pipeline=$pipeline_text "Pipeline ID"=$id Started=$started~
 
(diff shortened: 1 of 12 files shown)


priv/templates/common/error.greenbar
```diff
@@ -1,23 +1,38 @@
-~attachment title="Command Error" color="#ff3333" Caller=$initiator Pipeline=$pipeline_text "Pipeline ID"=$id Started=$started~
-~if cond=$planning_failure ~
-The pipeline failed planning the invocation:
-~br~
+~if cond=$execution_failure != ""~
```
mpeck: Whoa, that's not the easiest template to read. I wonder if it would behoove us to expand on the `if` tag to alleviate some of the duplication. Support for `and`/`or` and maybe `else` would make this a lot easier to read.

Not for this PR to be clear. Just musing about potential improvements.

test/cog/chat/slack/templates/common/error_test.exs
```diff
@@ -30,19 +24,15 @@ defmodule Cog.Chat.Slack.Templates.Common.ErrorTest do
...
              "execution_failure" => "I can't execute this!"}
     directives = directives_for_template(:common, "error", data)
     {"", [rendered]} = Cog.Chat.Slack.TemplateProcessor.render(directives)
-    assert """
-    The pipeline failed executing the command:
-
-    ```I can't execute this!```
 
-    The specific error was:
+    IO.inspect rendered
```
mpeck: Some errant debug code?

priv/templates/common/error.greenbar
imbriaco: Yeah, I agree that it's painful to read. You're exactly right that an `else` or `elseif` would have been super helpful. The other things that might be nice is a `~comment~` tag to leave some description of what the heck is going on when it gets tangled. I'll create issues for those.

test/cog/chat/slack/templates/common/error_test.exs
imbriaco: Yup, good catch. Fixed.

priv/templates/common/error.greenbar
imbriaco: Ooh, didn't know that! I'll go toss some comments in to make this template more clear and then merge it.

imbriaco:
Fixes #1182 and #1183.

PR #1185, opened 2016-11-28
Allow users to configure the previous command

vanstee:


config/config.exs
```diff
@@ -63,6 +63,7 @@ config :httpotion, :default_timeout, 30000
 # Commands, Bundles, and Services
 
 config :cog, :command_prefix, System.get_env("COG_COMMAND_PREFIX") || "!"
+config :cog, :previous_command, System.get_env("COG_PREVIOUS_COMMAND") || "!!"
```
christophermaier: Maybe call this `COG_PREVIOUS_COMMAND_PREFIX` instead?

lib/cog/command/pipeline/initializer.ex
```diff
@@ -17,19 +17,19 @@ defmodule Cog.Command.Pipeline.Initializer do
   alias Cog.Repository.ChatHandles
   alias Cog.Passwords
 
-  defstruct mq_conn: nil, history: %{}, history_token: ""
+  defstruct mq_conn: nil, history: %{}, previous_command: ""
```
christophermaier: This should be `previous_command_prefix` or `previous_command_token`, etc.  It might get a little confusing if people think this should literally hold the last command that was executed.

vanstee: Good point. `previous_command_token` it is.

config/config.exs
vanstee: stuck with the word token, changed to `COG_PREVIOUS_COMMAND_TOKEN`

PR #1193, opened 2016-11-30
support markdown links

### LINKED ISSUE #1076: Support parsing and outputting markdown links in greenbar templates
Most chat providers have support for links, so we should support handling markdown style links and formatting them correctly for Slack and Hipchat. Right now we're just outputing the text portion of the link.

mpeck:
This adds support for rendering the link directive to both hipchat and slack.

resolves #1076 

BLOCKED BY #1195

### DIFF
--- a/lib/cog/chat/hipchat/template_processor.ex
+++ b/lib/cog/chat/hipchat/template_processor.ex
@@ -41,7 +41,7 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
   # inform the user inline that there was a problem and log a warning.
   defp process_directive(%{"name" => "link", "text" => text, "url" => url}=directive) when url in [nil, ""] do
     Logger.warn("Invalid link; #{inspect directive}")
-    ~s[(invalid link! text:"#{text}" url: "#{url}")]
+    ~s[(invalid link! text:"#{text}" url: "#{inspect url}")]
   end
   defp process_directive(%{"name" => "link", "text" => text, "url" => url}),
     do: "<a href='#{url}'>#{text}</a>"
--- a/lib/cog/chat/hipchat/template_processor.ex
+++ b/lib/cog/chat/hipchat/template_processor.ex
@@ -33,6 +33,8 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
     do: "<pre>#{text}</pre>"
   defp process_directive(%{"name" => "paragraph", "children" => children}),
     do: Enum.map_join(children, &process_directive/1) <> "<br/><br/>"
+  # If you try to render a link with nil or blank text, nothing is displayed to the user.
+  # This way at least a link is rendered.
   defp process_directive(%{"name" => "link", "text" => text, "url" => url}) when text in [nil, ""],
     do: "<a href='#{url}'>#{url}</a>"
   defp process_directive(%{"name" => "link", "text" => text, "url" => url}),
--- a/lib/cog/chat/hipchat/template_processor.ex
+++ b/lib/cog/chat/hipchat/template_processor.ex
@@ -33,6 +33,8 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
     do: "<pre>#{text}</pre>"
   defp process_directive(%{"name" => "paragraph", "children" => children}),
     do: Enum.map_join(children, &process_directive/1) <> "<br/><br/>"
+  defp process_directive(%{"name" => "link", "text" => text, "url" => url}) when text in [nil, ""],
+    do: "<a href='#{url}'>#{url}</a>"
   defp process_directive(%{"name" => "link", "text" => text, "url" => url}),
     do: "<a href='#{url}'>#{text}</a>"
 
--- a/lib/cog/chat/hipchat/template_processor.ex
+++ b/lib/cog/chat/hipchat/template_processor.ex
@@ -33,8 +33,8 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
     do: "<pre>#{text}</pre>"
   defp process_directive(%{"name" => "paragraph", "children" => children}),
     do: Enum.map_join(children, &process_directive/1) <> "<br/><br/>"
-  # If you try to render a link with nil or blank text, nothing is displayed to the user.
-  # This way at least a link is rendered.
+  # If you try to render a link with nil or blank text, HipChat displays nothing to the user.
+  # This way at least a link is rendered. This basically mirrors the default Slack behavior.
   defp process_directive(%{"name" => "link", "text" => text, "url" => url}) when text in [nil, ""],
     do: "<a href='#{url}'>#{url}</a>"
   defp process_directive(%{"name" => "link", "text" => text, "url" => url}),
--- a/lib/cog/chat/hipchat/template_processor.ex
+++ b/lib/cog/chat/hipchat/template_processor.ex
@@ -31,9 +31,10 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
     do: "<code>#{text}</code>"
   defp process_directive(%{"name" => "fixed_width_block", "text" => text}),
     do: "<pre>#{text}</pre>"
-  defp process_directive(%{"name" => "paragraph", "children" => children}) do
-      Enum.map_join(children, &process_directive/1) <> "<br/><br/>"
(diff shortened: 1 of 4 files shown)


lib/cog/chat/hipchat/template_processor.ex
```diff
@@ -24,6 +24,8 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
     do: "<code>#{text}</code>"
   defp process_directive(%{"name" => "fixed_width_block", "text" => text}),
     do: "<pre>#{text}</pre>"
+  defp process_directive(%{"name" => "link", "text" => text, "url" => url}),
```
christophermaier: Do all `link` directives have both `text` and `url` fields?

mpeck: They did, there may be some changes with the new paragraph parsing bits, but I'll address that once those things are merged.

mpeck: Well I should say that the markdown style links did. There is another head that deals with links that have no text. I should probably group these together :|

mpeck: derp, I was wrong @christophermaier. That second head shouldn't ever get called. When links are specified using the markdown syntax they always have both text and url fields.

mpeck:
@kevsmith Tests added for the sad path.

One thing to note, I'm not really doing anything to validate the presence of the text or the url in a link specified by markdown. It seems like that should be something left up to the template author. The only exception is for hipchat templates. If the text is blank or nil I'm replacing it with the url. That mirrors Slack's default behavior and that's what I would expect.

lib/cog/chat/slack/template_processor.ex
```diff
@@ -90,8 +90,8 @@ defmodule Cog.Chat.Slack.TemplateProcessor do
   end
   defp process_directive(%{"name" => "list_item", "children" => children}, bullet: bullet),
     do: "   #{bullet} #{Enum.map_join(children, &process_directive/1)}\n"
-  defp process_directive(%{"name" => "link", "url" => url}, _) do
-    "#{url}"
+  defp process_directive(%{"name" => "link", "text" => text, "url" => url}, _) do
+    "<#{url}|#{text}>"
```
christophermaier: How can a link have empty or nil text for HipChat, but not for Slack? Shouldn't the processing be the same here as what you've done for HipChat?

test/cog/chat/hipchat/template_processor_test.exs
```diff
@@ -107,4 +107,37 @@ defmodule Cog.Chat.HipChat.TemplateProcessorTest do
...
+      %{"name" => "link", "text" => "a link", "url" => "http://www.example.com"}
+    ]
+
+    rendered = TemplateProcessor.render(directives)
+    expected = "<a href='http://www.example.com'>a link</a>"
+
+    assert expected == rendered
+  end
+
+  test "handles link directive with a nil url" do
```
christophermaier: How can we have a link with a `nil` URL?

mpeck: If the url portion of the link is set with a variable with a nil value.

For example:
`[my link](~$nil_var~)` would result in a directive like `%{"name" => "link", "text" => "my link", "url" => nil}`

lib/cog/chat/slack/template_processor.ex
mpeck: I could also be more explicit, that might make things clearer, but the behavior would remain the same.

test/cog/chat/hipchat/template_processor_test.exs
christophermaier: Pretty sure if that happens, it's unintentional, and the user should be made aware of it somehow.

lib/cog/chat/hipchat/template_processor.ex
```diff
@@ -33,6 +33,8 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
     do: "<pre>#{text}</pre>"
   defp process_directive(%{"name" => "paragraph", "children" => children}),
     do: Enum.map_join(children, &process_directive/1) <> "<br/><br/>"
+  # If you try to render a link with nil or blank text, nothing is displayed to the user.
+  # This way at least a link is rendered.
```
christophermaier: Clarifying that HipChat's default behavior is to render nothing with nil/blank text would be nice. At first, that wasn't quite clear, and this comment was a little confusing; for a second, I thought you were saying that it was Cog's default behavior.

lib/cog/chat/slack/template_processor.ex
```diff
@@ -90,6 +90,12 @@ defmodule Cog.Chat.Slack.TemplateProcessor do
   end
   defp process_directive(%{"name" => "list_item", "children" => children}, bullet: bullet),
     do: "   #{bullet} #{Enum.map_join(children, &process_directive/1)}\n"
+  # This is technially the default behavior in Slack. If you try to create a link with blank
+  # text only the link is displayed and is clickable, but it is included here for clarity
+  # and to conform more closely to the hipchat implementation.
+  defp process_directive(%{"name" => "link", "text" => text, "url" => url}, _) when text in [nil, ""] do
+    "<#{url}|#{url}>"
+  end
```
christophermaier: If this is Slack's default behavior, you don't need  to re-implement the default just so it matches the HipChat directive processor; a comment would suffice.

test/cog/chat/hipchat/template_processor_test.exs
mpeck: I would agree that it would probably be unintentional. What other way would you handle it? I'm basically mimicking Slack's behavior because that seemed like a reasonable default.

christophermaier: How do these behave in chat? They look like links, but when you click, nothing happens? That sounds maddening.

I was thinking perhaps something along the lines of how we handle unrecognized directives:
https://github.com/operable/cog/blob/402bfba5f427642640a09ecf4c48291ecaf98164/lib/cog/chat/slack/template_processor.ex#L103-L106

lib/cog/chat/hipchat/template_processor.ex
```diff
@@ -31,9 +31,20 @@ defmodule Cog.Chat.HipChat.TemplateProcessor do
...
+    do: Enum.map_join(children, &process_directive/1) <> "<br/><br/>"
+  # If you try to render a link with nil or blank text, HipChat displays nothing to the user.
+  # This way at least a link is rendered. This basically mirrors the default Slack behavior.
+  defp process_directive(%{"name" => "link", "text" => text, "url" => url}) when text in [nil, ""],
+    do: "<a href='#{url}'>#{url}</a>"
+  # Rendering a link in HipChat with a nil or blank url will obviously result in an invalid link. We
+  # inform the user inline that there was a problem and log a warning.
+  defp process_directive(%{"name" => "link", "text" => text, "url" => url}=directive) when url in [nil, ""] do
+    Logger.warn("Invalid link; #{inspect directive}")
+    ~s[(invalid link! text:"#{text}" url: "#{url}")]
```
christophermaier: You should do `#{inspect blah}` here so you can tell the difference between nil and `""`

lib/cog/chat/slack/template_processor.ex
```diff
@@ -90,8 +90,16 @@ defmodule Cog.Chat.Slack.TemplateProcessor do
   end
   defp process_directive(%{"name" => "list_item", "children" => children}, bullet: bullet),
     do: "   #{bullet} #{Enum.map_join(children, &process_directive/1)}\n"
-  defp process_directive(%{"name" => "link", "url" => url}, _) do
-    "#{url}"
+  # Rendering a link in Slack with a nil or blank url just results in '<|text>' which isn't super
+  # useful to a user. So instead we'll inform the user of the problem inline and log a warning.
+  defp process_directive(%{"name" => "link", "text" => text, "url" => url}=directive, _) when url in [nil, ""] do
+    Logger.warn("Invalid link; #{inspect directive}")
+    ~s[(invalid link! text:"#{text}" url: "#{url}")]
```
christophermaier: ditto

mpeck:
closed in favor of https://github.com/operable/greenbar/pull/33 since the template processing happens in greenbar now.

PR #1196, opened 2016-11-30
Avoid Slack API throttling in Integration tests

christophermaier:
See commit messages for details.


test/support/slack_client.ex
```diff
@@ -173,49 +130,39 @@ defmodule Cog.Test.Support.SlackClient do
...
   def handle_info({{ref, sender}=caller, {:chat_and_wait, room, message, reply_from, edited}}, state) do
-    :timer.sleep(@api_wait_interval)
+    api_wait()
     result = Slack.Web.Chat.post_message(room, %{as_user: true, text: message, token: state.token})
     case result["ok"] do
       false ->
         send(sender, {ref, {:error, result["error"]}})
         {:noreply, state}
       true ->
+        api_wait()
```
kevsmith: I wonder if it'd be cleaner to have a version of `api_wait` that took either a fun or an MFA and waited before executing the function and returned the results like so:

```
result = api_wait({Slack.Web.Chat, :update, [result["channel"], edited, result["ts"], 
                                             %{token: state.token, as_user: true, parse: :full}])
```

test/support/slack_client.ex
```diff
@@ -87,15 +55,20 @@ end
...
   # To avoid Slack throttling
-  @api_wait_interval 1100
+  @api_wait_interval 1_000
+  defp api_wait(),
+    do: :timer.sleep(@api_wait_interval)
 
-  @default_timeout 5000
+  # Be a little more liberal with the timeouts to account for
+  # potentially several Slack API calls.
+  @default_timeout 10_000
```
christophermaier: I can try futzing around to find a multiple that works, yeah.

test/support/slack_client.ex
christophermaier: Maybe? I'm not sure if it's worth it at this stage, though. As it is, it's already a bandaid on the Slack library anyway. 

Is there additional functionality you're envisioning for an MFA-accepting or HOF implementation of `api_wait` that we don't currently have and would benefit from?

kevsmith: I can think of things we could add like attaching retry or failure behavior declarations to the calls but nothing we need right now. And without looking at code I have no idea how much effort either of those would require.

kevsmith:
👍 to merge

PR #1197, opened 2016-12-01
Sane defaults cleanup

vanstee:
* Remove unnecessary env var now that `RELAY_MANAGED_DYNAMIC_CONFIG` defaults to true
* Downcase user provided relay ids

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,7 +68,6 @@ relay:
     - RELAY_DOCKER_CLEAN_INTERVAL=1m
     - RELAY_COG_HOST=cog
     - RELAY_DYNAMIC_CONFIG_ROOT=/tmp/bundle_configs
-    - RELAY_MANAGED_DYNAMIC_CONFIG=true
   links:
     - cog
   entrypoint: /usr/local/bin/relay
--- a/lib/cog/models/relay.ex
+++ b/lib/cog/models/relay.ex
@@ -32,8 +32,10 @@ defmodule Cog.Models.Relay do
 
   def allow_user_defined_id_on_insert(changeset) do
     case {Map.get(changeset.data, :id), get_field(changeset, :id)} do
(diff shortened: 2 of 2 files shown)


lib/cog/models/relay.ex
```diff
@@ -32,8 +32,10 @@ defmodule Cog.Models.Relay do
 
   def allow_user_defined_id_on_insert(changeset) do
     case {Map.get(changeset.data, :id), get_field(changeset, :id)} do
-      {nil, _changed_id} ->
+      {nil, nil} ->
         changeset
+      {nil, user_defined_id} ->
+        put_change(changeset, :id, String.downcase(user_defined_id))
```
christophermaier: Are we doing a belt-and-suspenders kind of thing here, given that Relay now does this as well?

docker-compose.yml
```diff
@@ -68,7 +68,6 @@ relay:
     - RELAY_DOCKER_CLEAN_INTERVAL=1m
     - RELAY_COG_HOST=cog
     - RELAY_DYNAMIC_CONFIG_ROOT=/tmp/bundle_configs
-    - RELAY_MANAGED_DYNAMIC_CONFIG=true
```
christophermaier: We also need to update the Cog Book bits on this to reflect that this is now the default, as well as that a directory is now basically required for the default.

lib/cog/models/relay.ex
vanstee: Kinda yeah. Since we're downcasing in Relay and using the same env var in our docker setup, I didn't want Cog to have the upcased version while Relay had the downcased one.

christophermaier: 👍

docker-compose.yml
vanstee: Created a PR for cog-book https://github.com/operable/cog-book/pull/61

PR #1205, opened 2016-12-02
Command output docs

### LINKED ISSUE #953: Document command output structure
Commands can return JSON objects, but to enable the efficient construction of pipelines, users should be able to quickly determine the structure of those objects, so they can know exactly how to extract the data they care about.

A first pass at this could be providing an `OUTPUT` section in the structured documentation. Having some way to actually ensure that the documented output is the same as the generated output would be ideal, but that's likely a bigger task.

Related: #948

vanstee:
Bundle authors can now include an `output` attribute when defining a command which is printed in a new "Output" section for command help docs.

Example:

screen shot 2016-12-02 at 12 03 17 pm

Closes #953

### DIFF
--- a/lib/cog/models/command_version.ex
+++ b/lib/cog/models/command_version.ex
@@ -16,6 +16,7 @@ defmodule Cog.Models.CommandVersion do
     field :arguments, :string
     field :subcommands, :map
     field :documentation, :string
+    field :output, :string
     field :status, :string, virtual: true
 
     belongs_to :command, Command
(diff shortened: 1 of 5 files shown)


vanstee:
I just yolo commited the 2 line change to spanner. I'll do follow ups for embedded bundle and a variety of Operable bundles.

vanstee:
For context of why this hasn't been merged yet. I'm going to break output out into two attributes: the description portion we have now, and an example attribut. This will also include an `--output` flag for the help command which returns that output example as a command result for testing in a pipeline.

PR #1211, opened 2016-12-06
hipchat timeout fix

mpeck:


test/support/hipchat_client.ex
```diff
@@ -47,6 +58,15 @@ defmodule Cog.Test.Support.HipChatClient do
     end
   end
 
+  defp wait_for_join(full) do
+    receive do
+      {:stanza, %Stanza.Message{body: "", from: %Romeo.JID{full: ^full}}} ->
+        :ok
+    after
+        @default_timeout -> raise "Error: Timeout waiting for bot to join room"
```
christophermaier: Let's put the body of the clause on an indented line by itself (per usual Elixir formatting), and add the name of the room we didn't join to the exception message.

test/support/hipchat_client.ex
```diff
@@ -35,10 +35,21 @@ defmodule Cog.Test.Support.HipChatClient do
     [hipchat_org|_] = String.split(user_name, "_", parts: 2)
     case Connection.start_link(opts) do
       {:ok, conn} ->
-        Connection.send(conn, Stanza.presence)
-        Connection.send(conn, Stanza.join("#{hipchat_org}_ci_bot_testing@conf.hipchat.com", nickname))
-        Connection.send(conn, Stanza.join("#{hipchat_org}_ci_bot_redirect_tests@conf.hipchat.com", nickname))
-        Connection.send(conn, Stanza.join("#{hipchat_org}_private_ci_testing@conf.hipchat.com", nickname))
+        :ok = Connection.send(conn, Stanza.presence)
+
+        # Rooms that we want to join
```
christophermaier: It'd be nice to add some explanatory comments here about why we currently believe it necessary to wait for confirmation that we've joined the chat room (explaining the kinds of transient errors we've seen, etc). That way we won't lose track of it in future refactorings, or will realize when we can remove it if we end up solving it a better way.

christophermaier:
If tests pass, :shipit:

PR #1213, opened 2016-12-07
Use Regex.replace to format Slack messages

### LINKED ISSUE #1161: Parsing fails with unicode chars followed by a URL
The following example demonstrates (ignore the messages from slackbot):

image

While Cog executed the first command, it didn't execute the second.

Happens on 0.16.1 _and_ in the Cog running in http://cogbot.slack.com.

vanstee:
`Regex.run` was returning incorrect indices when unicode was present. Instead of fixing that up, we can just switch to `Regex.replace`.

Fixes #1161

### DIFF
--- a/lib/cog/chat/slack/formatter.ex
+++ b/lib/cog/chat/slack/formatter.ex
@@ -1,5 +1,4 @@
 defmodule Cog.Chat.Slack.Formatter do
-
   @channel ~r/<\#(C.*)(\|(.*))?>/U
   @user    ~r/<\@(U.*)(\|(.*))?>/U
   @command ~r/<\!(.*)(\|(.*))?>/U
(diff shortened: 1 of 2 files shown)


lib/cog/chat/slack/formatter.ex
```diff
@@ -62,11 +62,26 @@ defmodule Cog.Chat.Slack.Formatter do
       nil ->
         message
       [{start, length}|_] ->
-        original = String.slice(message, start, length)
+        # Regex.run returns indices of a char list which causes weirdness when
+        # unicode is present. We have to convert with erlang's older binary to
+        # list conversion (ignoring unicode codepoints), do the replace, and
+        # then convert back with the same older conversion function for the
+        # indices to match up correctly. This is possibly a bug in Elixir's
+        # `Regex.run`.
```
kevsmith: Is this because `{start, length}` are byte indexes instead of Unicode code points? If so, are there any guarantees they're aligned to Unicode character boundaries so we avoid generating invalid strings?

I was looking at this code last night and wondered why we're not using named capture groups to extract the desired value and `String.replace` to edit the string. It would require running regexes twice but avoid having to care about string indexing at all.

vanstee: @kevsmith Yes, that's the reason. I'm not actually sure if there's the potentially to generate an invalid string here. I'll have to look at the implementation a bit more.

I think I remember the reason we don't `String.replace` being that there could be multiple identical matches in a string and we need to know which one to replace with with string. I might need to think about that a bit more, but your solution might still be ok.

PR #1215, opened 2016-12-07
Command output docs

vanstee:
Splits output documentation into description and example. Both or printed in the help docs for a command. The help command also now supports an `--output` flag which when used with a command name, only returns the json in the output example to the next command in the pipeline.


vanstee:
Merging as hipchat tests failed due to timeouts.

PR #1216, opened 2016-12-08
Be intelligent about API throttling

christophermaier:


test/support/slack_client.ex
```diff
@@ -73,9 +73,39 @@ defmodule Cog.Test.Support.SlackClient do
       nil ->
         raise RuntimeError, message: "$SLACK_USER_API_TOKEN not set!"
       token ->
-        {:ok, client} = start_link(token)
-        :ok = initialize(client)
-        {:ok, client}
+        api_wait()
+        api_wait()
```
vanstee: It does seem weird that this is called twice in a row. A note or a `:timer.sleep(@api_wait_interval * 2)` might be more intention revealing.

christophermaier: heh, a little too hacky... I'll fix that.

test/support/slack_client.ex
```diff
@@ -73,9 +73,39 @@ defmodule Cog.Test.Support.SlackClient do
...
+        try do
+          {:ok, client} = start_link(token)
+          :ok = initialize(client)
+          {:ok, client}
+        rescue
+          e in RuntimeError ->
```
christophermaier: Alas, it's what we have, though I suppose I could pattern match on the message.

vanstee: Eh, I think we'll be ok since we're logging it at least. Just wanted to avoid hiding any actual bugs behind this retry code. I think we'll be ok though.

PR #1224, opened 2016-12-12
Output docs

vanstee:
Since breaking down the other subcommand commands is a larger task, I wanted to get some basic output docs merged for 0.17. A few follow up PRs will break down existing commands and include command output docs in those as I go.


lib/cog/commands/help.ex
```diff
@@ -29,6 +29,31 @@ defmodule Cog.Commands.Help do
 
   @arguments "[<bundle> | <bundle:command>]"
 
+  @output_description """
+  Returns the highest version of each enabled and disabled bundle, a
+  specifc bundle, or a specific command depending on the args passed to the
```
mpeck: typo, "specific"

PR #1230, opened 2016-12-12
Revise DataStoreTest setup

christophermaier:
Since Mix / ExUnit start up the Cog application, and the `DataStore`
process is a registered one, we need to "swap out" that process for
tests.

This should be done for each test, swapping back the original process at
the end of each one.

Tests are also no longer run asynchronously, to prevent odd race
conditions.

### DIFF
--- a/test/cog/command/service/data_store_test.exs
+++ b/test/cog/command/service/data_store_test.exs
@@ -1,5 +1,5 @@
 defmodule Cog.Command.Service.DataStoreTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
 
   alias Cog.Command.Service.DataStore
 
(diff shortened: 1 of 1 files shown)


test/cog/command/service/data_store_test.exs
```diff
@@ -1,34 +1,30 @@
 defmodule Cog.Command.Service.DataStoreTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
```
mpeck: I believe that if don't run ecto in shared mode you can run the tests async. Downside is that you have to explicitly allow the connection for secondary processes that need to access the db. But I don't think we have to worry about that for these tests.

christophermaier: I didn't want our process-swapping to get messed up (e.g. one test process replace another test process' `DataStore`, etc.)

mpeck: ah, good point, didn't think about that.

Issue #1206, opened 2016-12-02
Rework old subcommand arg switching style commands


vanstee:
Since we now have the git style subcommand support, we can rework existing subcommands to have their own first-class documentation. Here's a list of all the commands in the embedded bundle that need fixing:

* [x] alias
* [x] bundle
* [x] group
* [x] permission
* [x] relay
* [x] relay-group
* [ ] role
* [x] rule
* [x] trigger
* [x] user

PR #1231, opened 2016-12-12
git style user command

mpeck:
updates the user command to use git style subcommands

part of #1206


vanstee:
I've been breaking down the tests into a test per command. Makes using `new_req` a bit easier as well. Not necessary though.

mpeck:
@vanstee I thought about doing that. But then you would have to run the tests like `mix test --only commands:user-<command>` for every command, instead of `mix test --only commands:user`. I thought the later was a more useful case, so I put them all together.

vanstee:
Oh I didn't know about those tags. I just run `mix test test/commands/alias*`. I can move those back so the tags make sense I guess. Would be less work for the rest of the tests.

PR #1232, opened 2016-12-12
git style trigger command

mpeck:
part of #1206


lib/cog/commands/helpers.ex
```diff
@@ -133,6 +133,8 @@ defmodule Cog.Commands.Helpers do
     do: "Could not find '#{resource_type}' with the name '#{resource_name}'"
   def error(:invalid_handle),
     do: "Invalid chat handle"
+  def error({:trigger_invalid, %Ecto.Changeset{}=changeset}),
+    do: changeset_errors(changeset)
```
vanstee: This doesn't have to be part of this PR, but it looks like we're including this `:trigger_invalid` symbol in the return value when we could probably just return the `{:error, changeset}`.

mpeck: I was pretty much just copying the bits that were already there. Though it may not be a terrible idea to clean some of these helpers up a bit. Or should we just create an issue for later?

vanstee: Whatever is good. Not really a big deal.

christophermaier: My $0.02 would be to tackle any changing of return values or anything else later; breaking up the commands is a big enough change as it is. It'll also be easier to review if there's only one thing you're dealing with.

test/commands/trigger_test.exs
```diff
@@ -1,13 +1,15 @@
 defmodule Cog.Test.Commands.TriggerTest do
-  use Cog.CommandCase, command_module: Cog.Commands.Trigger
+  use Cog.CommandCase, command_tag: :trigger
```
christophermaier: Is there any benefit to splitting out tests into separate files, one per subcommand? It feels like there's a bit more magic in the setup than we might need.

mpeck: @christophermaier we put them together to make the tests easier to run with tags. `mix test --only commands:trigger` vs `mix test --only commands:trigger-list --only commands:trigger-create ...`

christophermaier: @mpeck you can always add the `commands: :trigger` module tag to each file. If you wanted to run a single subcommand, you'd just execute that file directly.

mpeck: Yeah, that would also work. We can reorganize those bits. Is that something that should happen in these PRs or as separate work?

christophermaier: That kind of change seems like it'd go with the current work; the tests are changing to reflect the change in the commands.

PR #1233, opened 2016-12-13
Rework alias subcommands and write output docs

vanstee:


lib/cog/commands/alias.ex
```diff
@@ -37,32 +37,11 @@ defmodule Cog.Commands.Alias do
...
-               other ->
-                 {:error, {:unknown_subcommand, other}}
-             end
-
-    case result do
-      {:ok, template, data} ->
-        {:reply, req.reply_to, template, data, state}
-      {:error, err} ->
-        {:error, req.reply_to, Helpers.error(err), state}
+    if Helpers.flag?(req.options, "help") do
```
mpeck: Shouldn't we remove the usage bits? I've been taking this out in favor of the built in help bits.

lib/cog/commands/alias.ex
```diff
@@ -37,32 +37,11 @@ defmodule Cog.Commands.Alias do
...
+      {:ok, template, data} = show_usage
+      {:reply, req.reply_to, template, data, state}
+    else
+      Alias.List.handle_message(req, state)
```
mpeck: hrm, I just dropped the core command. I was thinking that it might be confusing to see it listed with everything else in the command list. Should I add this back in?

lib/cog/commands/alias.ex
christophermaier: It seems like the `-h` option was to get around the fact that with monolithic commands, the only way you could get help was from the top-level (e.g. `help alias` instead of `help alias delete`). If subcommands are now actually "real" commands, the need for `-h` disappears.

Are there other use cases I'm not considering?

vanstee: Agreed. Gonna remove it.

lib/cog/commands/alias.ex
christophermaier: Of course, I suppose bundle-level docs address that. Maybe we don't need a core command (particularly if we're end up being explicit and moving away from default commands)

vanstee: As mentioned in chat, I'm going to remove the base commands. I think the best solution will be to add the notion of command groups to chat which will generate help docs for those based on commands that share the same prefix.

PR #1236, opened 2016-12-13
Rework permission subcommands and write output docs

vanstee:


test/commands/permission_test.exs
```diff
@@ -1,6 +1,7 @@
 defmodule Cog.Test.Commands.PermissionTest do
```
mpeck: Still need to break these tests up.

vanstee:
Will merge once the tests have passed

PR #1237, opened 2016-12-13
Rework relay subcommands and write output docs

vanstee:


lib/cog/commands/relay.ex
```diff
@@ -19,46 +19,17 @@ defmodule Cog.Commands.Relay do
...
-             end
-
-    case result do
-      {:ok, template, data} ->
-        {:reply, req.reply_to, template, data, state}
-      {:ok, message} ->
-        {:reply, req.reply_to, message, state}
-      {:error, err} ->
-        {:error, req.reply_to, error(err), state}
+    if Helpers.flag?(req.options, "help") do
```
mpeck: I think we still need to drop this bit.

vanstee: derp

vanstee: Oh wait I just need to delete this entire command. My bad.

PR #1242, opened 2016-12-13
git style user command

mpeck:


test/commands/user/list_handles_test.exs
```diff
@@ -0,0 +1,32 @@
...
+  #### Setup Functions ####
+
+  defp with_users(_) do
+    [users: [user("admin"), user("tester")]]
+  end
+end
+
+
+
+
```
vanstee: Got some extra new lines in here

PR #1244, opened 2016-12-14
Rework relay group subcommands and write output docs

vanstee:


test/commands/relay_group/member/remove_test.exs
```diff
@@ -0,0 +1,26 @@
+defmodule Cog.Test.Commands.RelayGroupTest do
```
mpeck: Should this be `Cog.Test.Commands.Member.RemoveTest`

vanstee: Yep, thanks

PR #1250, opened 2016-12-14
rework rule command

mpeck:


lib/cog/commands/rule.ex
```diff
@@ -1,85 +1,31 @@
 defmodule Cog.Commands.Rule do
-  use Cog.Command.GenCommand.Base,
-    bundle: Cog.Util.Misc.embedded_bundle
 
-  alias Cog.Commands.Rule.{Info, List, Create, Delete}
   require Cog.Commands.Helpers, as: Helpers
```
vanstee: I think we can change this to an `alias` now as well FWIW. Not necessary though.

PR #1252, opened 2016-12-14
Remove support for Mustache templates

### LINKED ISSUE #1248: Remove Mustache templating support
With the addition of [Greenbar](https://github.com/operable/greenbar) templates, we can finally remove support for the old Mustache templates.

Among other things, this means the removal of the following:
1. Our dependency on [FuManchu](https://github.com/operable/fumanchu): https://github.com/operable/cog/blob/c0041e439dee3c2213f28444d95f995cfec75a1d/mix.exs#L65
2. The "old template logic" function head in the executor: https://github.com/operable/cog/blob/c0041e439dee3c2213f28444d95f995cfec75a1d/lib/cog/command/pipeline/executor.ex#L325-L351
3. Removal of the extracted "old template" module: https://github.com/operable/cog/tree/c0041e439dee3c2213f28444d95f995cfec75a1d/lib/cog/template/old, and renaming and reconsolidating the "new template" logic: https://github.com/operable/cog/tree/c0041e439dee3c2213f28444d95f995cfec75a1d/lib/cog/template
4. Removal of all "old template" logic in our TemplateRepository: https://github.com/operable/cog/blob/ed228d522814eab01761bc00c38d4490a2d637ce/lib/cog/repository/templates.ex

There's probably some other cleanup as well (database entries for old common templates?), but these are just from the top of my head.

christophermaier:
It's time to yank this; everybody should be using Greenbar!

Buried in this is also effectively a revert of
2f4a707

Fixes #1248


christophermaier:
Dropping in favor of #1261

PR #1254, opened 2016-12-15
Record higher resolution timestamps in the audit log

### LINKED ISSUE #1253: Increase timestamp resolution for events
Currently we only output to second resolution. When you import a log aggregator, you'd like to have more resolution to easily sort events.

Probably amounts to tweaking this format string: https://github.com/operable/cog/blob/c0041e439dee3c2213f28444d95f995cfec75a1d/lib/cog/events/util.ex#L7

imbriaco:
This updates the `timestamp` in audit logs to include microseconds as a decimal after the seconds part of the IS08601 formatted time. 

Example:
```json
"timestamp": "2016-12-15T16:28:34.332811Z"
```

Resolves #1253

### DIFF
--- a/lib/cog/events/util.ex
+++ b/lib/cog/events/util.ex
@@ -4,7 +4,7 @@ defmodule Cog.Events.Util do
   """
 
   # ISO-8601 UTC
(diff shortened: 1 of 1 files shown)


imbriaco:
Closing this in favor of a slightly larger rework that @christophermaier is working on separately.

PR #1258, opened 2016-12-16
command helpers cleanup

mpeck:


lib/cog/commands/alias.ex
```diff
@@ -0,0 +1,38 @@
...
+    do: Repo.get_by(UserCommandAlias, name: user_alias, user_id: user_id)
+  def get_command_alias(_, "site:" <> site_alias),
+    do: Repo.get_by(SiteCommandAlias, name: site_alias)
+  def get_command_alias(user_id, alias) do
+    case get_command_alias(user_id, "user:#{alias}") do
+      nil ->
+        get_command_alias(user_id, "site:#{alias}")
+      src_alias ->
+        src_alias
+    end
```
vanstee: I approved the PR so this is just a style question, but is the following less favored in functional programming? I just see it all the time in Ruby and liked the pattern. Not advocating a change here, just curious.

```elixir
get_command_alias(user_id, "user:#{alias}") ||
  get_command_alias(user_id, "site:#{alias}")
```

mpeck: idk to be honest. I generally stay away from that syntax because it can be confusing, but I'm not really passionate about it.

PR #1268, opened 2016-12-20
Refactor pipeline output disposition

### LINKED ISSUE #1243: Aborting commands is confusing with triggers
A command that is aborted currently does not behave as a failed command, so they behave counterintuitively when used with triggers. A triggered pipeline that aborts appears as a successful execution to a trigger, which can cause trouble when the success / failure status matters (when you're implementing a step of a CI pipeline using a trigger, for instance).

We should revisit how aborts are handled internally and find a way to expose aborts meaningfully to triggers. Returning an HTTP 422 seems like a good candidate for a final disposition.

christophermaier:
Now we formally classify redirection destinations according to how the
output needs to be rendered. We have three types of output currently:

* chat: a chat adapter destination. Output is rendered through
  Greenbar templates.
* trigger: output destined for trigger destinations. Output is *not*
  rendered through Greenbar, but embeded in a map with additional
  information. The keys are as follows:

  status: Either "success", "abort", or "error"
  pipeline_output: the output of the pipeline.
  message: explanatory message for "abort" or "error" responses.

  The only trigger adapter currently is the "http" adapter. The
  response statuses map to the following HTTP codes:

  success: 200
  error: 500
  abort: 422

* status_only: an abbreviated version of trigger-style output,
  containing only "status" and (optionally) "message" keys. The output
  of the pipeline is *not* included.

  Chat destinations cannot currently be "status-only"; this is reserved
  for trigger-initiated pipelines where the output is not explicitly
  redirected back to the requestor. HTTP requests must have a response,
  and this is how we ensure that the success or failure of the pipeline
  is reported back to the requestor.

By classifying our redirection destinations this way, we can express the
logic for rendering the outputs more explicitly. Furthermore, we can
generalize the output generation process so it will be easier to add
additional types of output disposition strategies in the future, should
the need arise. It also allows us to consolidate the various code paths
leading to response output.

Some other points:

* Anytime a pipeline results in an error, the result is only returned to
  the source of the request, regardless of which redirect destinations
  may be present. This was the behavior for only trigger-initiated
  pipelines previously, but now applies to all pipeline requests.
* Common templates are now referred to referred to internaly using a
  tagged tuple: `{:common, template_name}`.

Fixes #1243

### DIFF
--- a/lib/cog/command/pipeline/destination.ex
+++ b/lib/cog/command/pipeline/destination.ex
@@ -1,28 +1,36 @@
 defmodule Cog.Command.Pipeline.Destination do
 
-  @type output_level :: :full | :status_only
-
+  @type classification :: :chat | :trigger | :status_only
   @type t :: %__MODULE__{raw: String.t,
-                         output_level: output_level,
                          adapter: String.t,
-                         room: %Cog.Chat.Room{}}
+                         room: %Cog.Chat.Room{},
+                         classification: classification}
   defstruct [raw: nil,
-             output_level: :full,
              adapter: nil,
-             room: nil]
+             room: nil,
+             classification: nil]
+
   require Logger
   alias Cog.Chat.Adapter
 
   @doc """
   Given a list of raw pipeline destinations, resolve them all to the
-  appropriate adapter-specific destinations, or return error tuples
-  for each that were invalid.
+  appropriate adapter-specific destinations, grouped by their output
+  disposition classification, or return error tuples for each that
+  were invalid.
   """
   def process(raw_destinations, sender, origin_room, origin_adapter) when is_binary(origin_adapter) do
-    raw_destinations
+    result = raw_destinations
     |> Enum.map(&make_destination/1)
     |> maybe_add_origin(origin_adapter)
     |> resolve(sender, origin_room, origin_adapter)
+
+    case result do
+      {:ok, destinations} ->
+        {:ok, Enum.group_by(destinations, &(&1.classification))}
+      {:error, _}=error ->
+        error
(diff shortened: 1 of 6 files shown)


lib/cog/command/pipeline/executor.ex
```diff
@@ -299,78 +304,82 @@ defmodule Cog.Command.Pipeline.Executor do
...
-    Enum.each(filtered_destinations, fn(d) ->
-      ReplyHelper.send("early-exit", %{}, d.room, d.adapter, state.mq_conn)
-    end)
-
+    # We don't want to send output to other destinations, because
+    # you'd end up with "mystery" messages saying "Pipeline succeeded
+    # but there was no output!" without indication of what's going
+    # on.
+    #
+    # TODO: If the "early-exit" template gets more contextual smarts, we
```
kevsmith: What kind of contextual smarts would the template need?

christophermaier: Additional information about who executed a pipeline, when, from where, what the pipeline was, which stage it got to before returning no output, etc.; something that would give the members of a chat room enough context to know what was going on if they got redirected to from such a pipeline.

lib/cog/command/pipeline/destination.ex
```diff
@@ -1,28 +1,36 @@
 defmodule Cog.Command.Pipeline.Destination do
 
-  @type output_level :: :full | :status_only
-
+  @type classification :: :chat | :non_chat | :status_only
```
christophermaier: Derp... I'll change it to `:trigger`

PR #1274, opened 2016-12-20
Link triggers and their users via FK constraint

### LINKED ISSUE #1246: Trigger users need to have referential integrity
Triggers need to be linked to their users. Users can be deleted from the system, but the triggers remain behind, even showing the name of now-deleted user. The trigger can no longer be executed, either. If another user with the same name is created, the trigger will be run as that user instead.

Triggers need to be linked with referential integrity constraints to their users. When a user is deleted, we should either a) prevent the deletion or b) set the trigger user to `NULL`.

christophermaier:
If you attempt to delete a user that is currently associated with any
triggers (i.e., that user is the one the triggers run as), the request
will be blocked.

This will prevent triggers from breaking when users are deleted,
allowing operators to re-wire their triggers as appropriate.

Fixes #1246

### DIFF
--- a/priv/repo/migrations/20161220162331_trigger_user_referential_integrity.exs
+++ b/priv/repo/migrations/20161220162331_trigger_user_referential_integrity.exs
@@ -1,0 +1,13 @@
+defmodule Cog.Repo.Migrations.TriggerUserReferentialIntegrity do
+  use Ecto.Migration
+
+  def change do
+    execute """
+    ALTER TABLE triggers
+    ADD CONSTRAINT triggers_as_user_fkey
+    FOREIGN KEY(as_user) REFERENCES users(username)
+    ON DELETE RESTRICT
+    ON UPDATE CASCADE
+    """
+  end
+end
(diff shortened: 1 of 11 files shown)


priv/repo/migrations/20161220162331_trigger_user_referential_integrity.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Cog.Repo.Migrations.TriggerUserReferentialIntegrity do
+  use Ecto.Migration
+
+  def change do
+    execute """
+    ALTER TABLE triggers
+    ADD CONSTRAINT triggers_as_user_fkey
+    FOREIGN KEY(as_user) REFERENCES users(username)
+    ON DELETE RESTRICT
+    ON UPDATE CASCADE
```
vanstee: This looks good, but wouldn't it be nicer to store the user id here instead? Then we wouldn't have to worry about updates.

christophermaier: We talked about that, but decided to keep the username for now since it was more compatible with the current schema, and using the ID would require a more complicated migration and additional code modifications. It's also as unique as the ID.

PR #1276, opened 2016-12-20
Simplify audit log events

### LINKED ISSUE #1255: Modify event log format to be friendlier to log aggregators
Instead of stuffing additional fields into a `data` field, just create a flat map.

Also, encode complex values (e.g. `cog_env`) as JSON, instead of as maps.

christophermaier:
Removes the `data` map field from each event type. Now, all events are
simply flat maps, which should make ingesting them into aggregation
software (e.g. Logstash, Splunk) more straightforward.

Internally, we're just using maps instead of structs, though we still
maintain centralized control of how the maps are generated, so they
should be self-consistent.

Note that the order of the fields in the printed JSON in the log file
will not be deterministic (these are maps, not structs), so you'll need
to take that into account if you're rolling your own log analysis
tooling. Treat each line like a JSON map, though, and you'll be fine.

Fixes #1255

### DIFF
--- a/lib/cog/events/api_event.ex
+++ b/lib/cog/events/api_event.ex
@@ -1,11 +1,53 @@
 defmodule Cog.Events.ApiEvent do
   @moduledoc """
-  Provides functions for generating REST API request processing
-  events.
+  Encapsulates information about REST API request processing
+  events. Each event is a map; all events share a core set of fields,
+  while each event sub-type will have an additional set of fields
+  particular to that sub-type.
 
   The functions in this module generate API events from `Plug.Conn`
   instances, and depend on various metadata having been added to the
   `Conn` beforehand; see `Cog.Plug.Util` and `Cog.Plug.Event`
+
+  # Common Fields
+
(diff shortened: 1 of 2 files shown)


christophermaier:
@imbriaco Once the dust settles in operable/cog-book, I can add a page documenting the event fields (basically taking the `moduledoc` from these modules and putting them in the book).

imbriaco:
Did you decide not to struct-ify it after all?

christophermaier:
Yeah, decided to go with plain maps. Couldn't really find a way to use structs without repeating all the common fields for each one. Also decided that the work to control the order that the fields printed in logs wasn't worth it, given the relative ease with which you could whip up some custom parsing code with Ruby, `jq`, etc if you really just had to do some kind of processing of the logs on the command line.

imbriaco:
Fair enough. I think flattening is valuable enough as is to go ahead and merge this. I may poke at ordering later if I feel the urge but :+1: to merge.

PR #1277, opened 2016-12-21
Improve list templates and fix failing tests due to bugfixes

vanstee:
Most of the changes here revolve around making list commands print scannable output. The only exception is `rule list` as other commands require an id, but we need to show the entire rule since that's really the only identifiable information available.

Tests were also fixed to remove newlines and join tags and extra break tags in hipchat output which were each related to recent greenbar bugfixes.

### DIFF
--- a/priv/templates/embedded/bundle-versions.greenbar
+++ b/priv/templates/embedded/bundle-versions.greenbar
@@ -1,15 +1,24 @@
 ~each var=$results~
-~$item.compatible~
+
 ~if cond=$item.incompatible not_bound?~
-~attachment color="gray"~
-**Version:** ~$item.version~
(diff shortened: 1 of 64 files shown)


priv/templates/embedded/bundle-versions.greenbar
```diff
@@ -1,15 +1,24 @@
 ~each var=$results~
-~$item.compatible~
```
vanstee: @mpeck Wasn't sure what this did. Let me know if we need to keep it.

vanstee: Confirmed with @mpeck that this can be removed.

Issue #1262, opened 2016-12-16
Explicitly refuse to run unsupported bundles


christophermaier:
Currently, Cog has an `n` and `n-1` policy for which bundle versions it will support. That is, if `v5` is the current bundle configuration version, then you can run `v5` and `v4` bundles, but nothing older.

You cannot install older bundles, but there isn't currently anything that prevents the execution of outdated bundles, should they already be installed on a system prior to the Cog upgrade that removed support. This can cause unexpected failures as Cog tries to execute commands that are built with assumptions of an execution environment that may no longer exist.

We should explicitly disallow the execution of such commands. Ideas include automatic disabling upon upgrade (with some kind of notification to the user / operator), refusal to enable outdated bundles, etc.

imbriaco:
Note that we'll also want to be able to call these out in `bundle list` and `help` much the same way we can disabled bundles.

mpeck:
Incompatible bundle versions are implied by querying the `cog_bundle_version` on the `config_file` column so there are no changes required to the db.

Visually we need to call out incompatible bundles in both chat commands and cogctl.

`help`
Bundles that have *NO* compatible versions installed will show in a new section with an "Incompatible Bundles" header.

`bundle versions`
Incompatible versions should be flagged as "incompatible"

`bundle info`
Incompatible versions should be flagged as "incompatible"

`cogctl bundle info`
Incompatible versions should be flagged as "incompatible"

`cogctl bundle versions`
Incompatible versions should be listed under "incompatible versions". There should be a flag `--incompatible` that only lists incompatible versions for the specified bundle.
Note: There currently is no way to list all bundle versions in cogctl. You must always specify the bundle to list versions on. If a sysadmin wanted to view all incompatible versions he or she would need to first get a list of all bundles and then list incompatible versions for each bundle. This isn't ideal. We should probably allow the listing of all bundle versions via `cogctl bundle versions` sans a bundle name, and then provide optional filtering via the `--incompatible` flag.

`cogctl bundle uninstall`
There should be a flag `--incompatible` that uninstalls just incompatible versions.
Note: It would be nice if we could just accept the bundle list on stdin and then do the right thing. An example workflow might be, `cogctl bundle | cogctl bundle uninstall --incompatible` to uninstall all incompatible bundles.

PR #1278, opened 2016-12-28
Disable and disallow enabling of incompatible bundle versions

mpeck:
This moves a lot of the post startup bits to a separate module and adds additional logic to disable old incompatible bundles. It also prevents incompatible bundles from being re-enabled.

part of #1262 
depends on https://github.com/operable/spanner/pull/88

### DIFF
--- a/lib/cog/repository/bundles.ex
+++ b/lib/cog/repository/bundles.ex
@@ -39,6 +39,8 @@ defmodule Cog.Repository.Bundles do
   @install_types [:normal, :force]
   @default_install_type :normal
 
+  @minimum_bundle_config_version Spanner.Config.minimum_config_version
+
   ########################################################################
   # CRUD Operations
 
@@ -200,9 +202,18 @@ defmodule Cog.Repository.Bundles do
   ########################################################################
   # Enabled/Disabled Status
 
-  def set_bundle_version_status(%BundleVersion{bundle: %Bundle{name: protected}}, _) when protected in @reserved_bundle_names,
+  # Users aren't allowed to change the status of bundles with a reserved name.
+  # If they attempt it, return an error.
+  def set_bundle_version_status(%BundleVersion{bundle: %Bundle{name: protected}}, _)
+  when protected in @reserved_bundle_names,
     do: {:error, {:protected_bundle, protected}}
-  def set_bundle_version_status(bundle_version, status) when status in [:enabled, :disabled],
+  # If an attempt is made to enable a bundle version with an incompatible bundle config
+  # then we return an error.
+  def set_bundle_version_status(%BundleVersion{config_file: %{"cog_bundle_version" => cbv}}=bv, status)
+  when cbv < @minimum_bundle_config_version and status in [:enabled],
+    do: {:error, {:incompatible_config_version, bv}}
+  def set_bundle_version_status(bundle_version, status)
+  when status in [:enabled, :disabled],
     do: __set_bundle_version_status(bundle_version, status)
 
   # Private implementation that we can use to ensure that the right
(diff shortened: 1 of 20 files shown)


lib/cog/repository/bundles.ex
```diff
@@ -292,6 +299,30 @@ defmodule Cog.Repository.Bundles do
   end
 
   @doc """
+  Disable old incompatible bundles
+  """
+  def disable_incompatible_bundles() do
+    query =
```
kevsmith: How many transactions does this function require assuming `n` incompatible bundles?

web/views/bundles_view.ex
```diff
@@ -35,6 +36,18 @@ defmodule Cog.V1.BundlesView do
 
   ########################################################################
 
+  defp ordered_compatible_versions(versions) do
```
kevsmith: These functions might be a better fit on the `Bundle` model itself, no?

test/support/util.ex
```diff
@@ -0,0 +1,24 @@
...
+
+  @doc """
+  unwraps :error tuples, returning just the error
+  If a tuple without an :error atom is passed, raises an error.
+  """
+  def unwrap_error({:error, error}),
+    do: error
+  def unwrap_error(value),
+    do: raise "Expected {:error, error}, but received #{inspect(value)} instead."
+
```
kevsmith: Logic can be written more succinctly as:

```
def unwrap({thing, value}, thing), do: value
def unwrap(value, thing), do
  raise "Expected 2-tuple starting with #{inspect thing} but received #{inspect value} instead."
end
```

```
iex(3)> Util.unwrap({:ok, "foo"}, :ok)
"foo"
iex(4)> Util.unwrap({:error, :timeout}, :error)
:timeout
iex(5)> Util.unwrap({:error, :timeout}, :error2)
** (RuntimeError) Expected 2-tuple starting with :error2 but received {:error, :timeout} instead.
iex:3: Util.unwrap/2
```

lib/cog/repository/bundles.ex
```diff
@@ -39,6 +39,8 @@ defmodule Cog.Repository.Bundles do
   @install_types [:normal, :force]
   @default_install_type :normal
 
+  @minimum_bundle_config_version Spanner.Config.minimum_config_version
```
mpeck: Generally I agree with you, but it gets used in a guard expression and I can't call normal functions there.

lib/cog/startup.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Cog.Startup do
+  require Logger
+
+  @moduledoc """
+  Tasks that should happen after the main Cog supervisor has started up.
+  Note: Don't start any long running or supervised processes here. This
+  module is purely for transient book keeping/setup functions.
+  """
```
mpeck: Yeah, I'll just drop the note. I was trying to say not to start up any supervised processes here because I thought we might end up chasing down some phantom process. But in hindsight I guess that's kinda obvious.

lib/cog/startup.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Cog.Startup do
```
mpeck: I was thinking that it would make organization a bit better. We are starting to get a good bit of logic in there so I thought it was time to break it up. I also thought that it might make it easier to refactor this bit in the future if we so desired such a thing. That said, I wouldn't be heartbroken to move it back if you think that's better. It's basically just copied and pasted from the top level module.

web/views/bundles_view.ex
mpeck: I agree, but I think they're ok for now. Aside from the ordering bit they're just used to format the data for presentation. We should be returning versions from the repository grouped into compatible and incompatible, and sorted. But because of the way we get bundles, that would take a bit of work that I didn't want to bite off since this PR was already getting pretty big.

I would like to do some refactoring on the bundles repository. It's already pretty huge and I'm pretty sure we can consolidate a good bit of the code in there.

lib/cog/startup.ex
mpeck: 10-4, I can move it back then.

lib/cog/repository/bundles.ex
kevsmith: I see.

I still think `set_bundle_version_status` could use some explanatory comments to aid understanding.

mpeck: I added some comments @kevsmith. Were those not clear enough? Or is Github just not showing the changes?

kevsmith: I think I was reading an outdated diff. My bad.

PR #1287, opened 2017-01-05
Share MQTT connections between processes

kevsmith:
Refactored `Carrier.Messaging.Connection` so multiple Erlang/Elixir processes can share
a single connection. The connection keeps track of subscribers and GenMqtt-style reply endpoints and ensures correct delivery of messages including support for `+` and `*` topic wildcards. `Connection` places monitors on all subscribing processes allowing it to clean up subscriptions after subscribers crash.

These changes are in preparation for the new executor design based on GenStage. Attempting to share a single connection between stages for the same pipeline didn't integrate well with GenMqtt causing Cog to use transient MQTT connections and incurring a small, but noticeable,
performance penalty.

Centralizing connection sharing logic in the connection itself resolves all previous problems and results in a more testable design compared to my earlier attempts at layering multiplexing on top of `Connection`.

### DIFF
--- a/lib/carrier/messaging/connection.ex
+++ b/lib/carrier/messaging/connection.ex
@@ -20,6 +20,7 @@ defmodule Carrier.Messaging.Connection do
 
   @internal_mq_username Cog.Util.Misc.internal_mq_username
   @default_connect_timeout 5000 # 5 seconds
+  @default_call_timeout 5000 # 5 seconds
   @default_log_level :error
 
   defstruct [:conn, :tracker, :owner]
@@ -101,7 +102,7 @@ defmodule Carrier.Messaging.Connection do
   @spec call(conn::connection(), topic::String.t, endpoint::String.t, message::Map.t, opts::call_opts()) :: Map.t | {:error, atom()}
   def call(conn, topic, endpoint, message, opts \\ []) do
     subscriber = Keyword.get(opts, :subscriber, self())
-    timeout = Keyword.get(opts, :timeout, 5000)
+    timeout = Keyword.get(opts, :timeout, @default_call_timeout)
     GenServer.call(conn, {:call, topic, endpoint, message, subscriber, timeout}, :infinity)
   end
 
--- a/lib/carrier/messaging/connection.ex
+++ b/lib/carrier/messaging/connection.ex
@@ -2,6 +2,9 @@ defmodule Carrier.Messaging.Connection do
 
   require Logger
 
+  use Cog.Util.Debug
+
+  alias Carrier.Messaging.Tracker
   alias Carrier.Messaging.Messages.MqttCall
   alias Carrier.Messaging.Messages.MqttCast
   alias Carrier.Messaging.Messages.MqttReply
@@ -9,30 +12,30 @@ defmodule Carrier.Messaging.Connection do
   require Record
   Record.defrecord :hostent, Record.extract(:hostent, from_lib: "kernel/include/inet.hrl")
 
+  use GenServer
+
   @moduledoc """
-  Interface for the message bus on which commands communicate.
+  General interface to Cog's message bus
   """
 
   @internal_mq_username Cog.Util.Misc.internal_mq_username
   @default_connect_timeout 5000 # 5 seconds
   @default_log_level :error
 
-  defstruct [:id, :conn, :call_reply]
+  defstruct [:conn, :tracker, :owner]
 
   # Note: This type is what we get from emqttc; if we change
   # underlying message buses, we can just change this
   # definition. Client code can just depend on this opaque type and
   # not need to know that we're using emqttc at all.
   @typedoc "The connection to the message bus."
-  @opaque connection :: %__MODULE__{}
+  @opaque connection :: pid()
 
-  @doc """
-  Starts up a message bus client process using only preconfigured parameters.
-  """
-  @spec connect() :: {:ok, connection()} | :ignore | {:error, term()}
-  def connect() do
-    connect([])
-  end
+  @type call_opts :: [] | [call_opt()]
+  @type call_opt :: {:timeout, integer()} | {:subscriber, pid()}
+
+  @type publish_opts :: [] | [publish_opt()]
+  @type publish_opt :: {:routed_by, String.t} | {:threshold, integer()}
 
   @doc """
   Starts up a message bus client process.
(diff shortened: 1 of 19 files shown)


lib/cog/command/pipeline/executor.ex
```diff
@@ -101,17 +102,18 @@ defmodule Cog.Command.Pipeline.Executor do
 
   require Logger
 
-  def start_link(request),
-    do: :gen_fsm.start_link(__MODULE__, [request], [])
+  def start_link(request) do
+    result = :gen_fsm.start_link(__MODULE__, [request], [])
+    result
+  end
```
christophermaier: What's the difference here?

lib/cog/command/pipeline/executor.ex
```diff
@@ -626,13 +628,12 @@ defmodule Cog.Command.Pipeline.Executor do
 
   defp fetch_user_from_request(%Cog.Messages.AdapterRequest{}=request) do
     # TODO: This should happen when we validate the request
-    if ChatAdapter.is_chat_provider?(request.adapter) do
+    result = ChatAdapter.is_chat_provider?(request.adapter)
+    if result do
```
christophermaier: Why this change?

test/support/snoop.ex
```diff
@@ -108,7 +108,7 @@ defmodule Cog.Snoop do
   # GenServer implementation details
 
   def init(topic) do
-    {:ok, mq_conn} = Connection.connect
+    {:ok, mq_conn} = Carrier.Messaging.ConnectionSup.connect()
```
christophermaier: yay, parentheses 👍

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+  defstruct [reply_endpoints: %{}, subscriptions: %{}, monitors: %{}, unused_topics: []]
+
+  @spec add_reply_endpoint(Tracker.t, pid()) :: {Tracker.t, String.t}
+  def add_reply_endpoint(%__MODULE__{reply_endpoints: reps}=tracker, subscriber) do
+    case Map.get(reps, subscriber) do
+      nil ->
+        topic = make_reply_endpoint_topic()
+        reps = reps
+        |> Map.put(subscriber, topic)
+        |> Map.put(topic, subscriber)
```
christophermaier: So `reply_endpoints` is actually intended to serve as a bi-directional lookup linking each subscriber process to their unique message queue topic (and vice versa), correct?

If so, is there a better name that could capture that? `reply_endpoints` sounds like it'd be a list of topics.

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+        {maybe_monitor_subscriber(%{tracker | reply_endpoints: reps}, subscriber), topic}
+      topic ->
+        {tracker, topic}
```
christophermaier: Is it a valid scenario to call this function more than once per subscriber, or is this a defensive, just-in-case thing?

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+  end
+
+  @spec add_subscription(Tracker.t, String.t, pid()) :: Tracker.t
+  def add_subscription(%__MODULE__{subscriptions: subs}=tracker, topic, subscriber) do
+    subs = case Map.get(subs, topic) do
+             nil ->
+               Map.put(subs, topic, {subscription_matcher(topic), [subscriber]})
+             {matcher, subscribed} ->
+               Map.put(subs, topic, {matcher, Enum.uniq([subscriber|subscribed])})
+           end
```
christophermaier:
```elixir
subs = Map.update(subs, topic, 
                 {subscription_matcher(topic), [subscriber]},  
                 fn({m,subs}) -> {m, Enum.uniq[subscriber|subs] end)
```
Wish they had an `update_lazy`, though 😦

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+    maybe_monitor_subscriber(%{tracker | subscriptions: subs}, subscriber)
+  end
+
+  @spec del_subscription(Tracker.t, String.t, pid()) :: {Tracker.t, boolean()}
+  def del_subscription(%__MODULE__{subscriptions: subs, unused_topics: ut}=tracker, topic, subscriber) do
+    case Map.get(subs, topic) do
+      {_matcher, [^subscriber]} ->
+        subs = Map.delete(subs, topic)
+        {maybe_unmonitor_subscriber(%{tracker | subscriptions: subs, unused_topics: Enum.uniq([topic|ut])}, subscriber), true}
```
christophermaier: Since we're doing the unsubscribing outside of this module, could we just pass out any unused topics in the return value from this function, instead of tracking them internally?

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+
+  @spec del_subscriber(Tracker.t, pid()) :: Tracker.t
+  def del_subscriber(tracker, subscriber) do
+    tracker
+    |> del_reply_endpoint(subscriber)
+    |> del_all_subscriptions(subscriber)
+    |> unmonitor_subscriber(subscriber)
+  end
+
+  @spec unused?(Tracker.t) :: boolean()
```
christophermaier: It doesn't look like this is used in non-testing code... could this become a testing helper instead?

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+  end
+
+  @spec find_reply_subscriber(Tracker.t, String.t) :: pid() | nil
+  def find_reply_subscriber(%__MODULE__{reply_endpoints: reps}, topic) do
+    Map.get(reps, topic)
+  end
+
+  @spec find_subscribers(Tracker.t, String.t) :: [] | [pid()]
+  def find_subscribers(%__MODULE__{subscriptions: subs}, topic) do
+    Enum.reduce(subs, [], &(find_matching_subscriptions(&1, topic, &2)))
```
christophermaier: We could simplify `find_matching_subscriptions` (and this) by rewriting in terms of `Enum.flat_map` instead.

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+  end
+
+  def unmonitor_subscriber(%__MODULE__{monitors: monitors}=tracker, subscriber) do
+    mref = Map.fetch!(monitors, subscriber)
+    :erlang.demonitor(mref, [:flush])
+    %{tracker | monitors: Map.delete(monitors, subscriber)}
+  end
+
+  defp has_subscriptions?(tracker, subscriber) do
+    if Enum.reduce_while(tracker.reply_endpoints, false, &(has_endpoint(&1, subscriber, &2))) do
```
christophermaier: Since `reply_endpoints` is bidirectional, could this be replaced with `Map.has_key(reply_endpoints, subscriber)`?

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+      true
+    else
+      Enum.reduce_while(tracker.subscriptions, false, &(has_any_subscription(&1, subscriber, &2)))
```
christophermaier: I think this could be simplified if recast in terms of `Enum.any?`

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+  end
+
+  defp del_all_subscriptions(%__MODULE__{subscriptions: subs}=tracker, subscriber) do
+    subs = Enum.reduce(Map.keys(subs), subs, &(delete_subscription(&1, subscriber, &2)))
+    %{tracker | subscriptions: subs}
+  end
+
+  defp delete_subscription(topic, subscriber, subs) do
+    case Map.get(subs, topic) do
+      {_, []} ->
```
christophermaier: If nobody is subscribed, why do we keep the entry?

lib/carrier/messaging/tracker.ex
kevsmith: This is to make subscribing to a topic idempotent.

lib/carrier/messaging/tracker.ex
kevsmith: This will be used with GenStage code. I want connections to close themselves when they aren't being used. I tried enabling the behavior now but it broke tests since the connection would close itself when all the subscriptions were removed during test teardown.

lib/carrier/messaging/connection.ex
```diff
@@ -47,14 +50,85 @@ defmodule Carrier.Messaging.Connection do
...
+    timeout = Keyword.get(opts, :timeout, 5000)
+    GenServer.call(conn, {:call, topic, endpoint, message, subscriber, timeout}, :infinity)
+  end
+
+  @doc """
+  Sends a GenMqtt cast message
+  """
+  @spec cast(conn::connection(), topic::String.t, endpoint::String.t, message::Map.t) :: :ok | {:error, atom()}
+  def cast(conn, topic, endpoint, message) do
+    GenServer.call(conn, {:cast, topic, endpoint, message}, :infinity)
```
kevsmith: 1 hour is effectively infinity :)

The design predates this PR. Let's tackle it separately.

christophermaier: Sounds good... this was just the first time I noticed infinity = 1 hour.

Some infinities are larger than other infinities, though, so it's cool 👅

lib/carrier/messaging/tracker.ex
kevsmith: I couldn't think of a better name. And really this is an internal implementation detail of the Tracker. External callers should stick to the API.

lib/carrier/messaging/tracker.ex
kevsmith: I tried that but felt it bloated the function's return value. `{tracker, boolean()}` vs. `{tracker, boolean(), [String.t]}`

lib/carrier/messaging/tracker.ex
```diff
@@ -0,0 +1,179 @@
...
+
+  @spec unused?(Tracker.t) :: boolean()
+  def unused?(%__MODULE__{reply_endpoints: reps, subscriptions: subs, monitors: monitors}) do
+    Enum.empty?(reps) and Enum.empty?(subs) and Enum.empty?(monitors)
+  end
+
+  @spec get_and_reset_unused_topics(Tracker.t) :: {Tracker.t, [String.t]}
+  def get_and_reset_unused_topics(tracker) do
+    {%{tracker | unused_topics: []}, tracker.unused_topics}
+  end
```
kevsmith: Which would require external callers to directly access struct fields making it harder to change behavior in the future.

lib/carrier/messaging/tracker.ex
christophermaier: I get that it's internal; I was thinking more from a readability perspective than an API perspective.

lib/carrier/messaging/tracker.ex
kevsmith: Ah yeah. Whoops.

lib/cog/command/pipeline/executor.ex
kevsmith: Leftover from debugging. Will remove.

lib/cog/command/pipeline/executor.ex
kevsmith: Debugging. Will fix.

lib/carrier/messaging/tracker.ex
christophermaier: It does bloat it a bit, but it might be worth it to reduce the surface area of code in this module.

lib/carrier/messaging/tracker.ex
christophermaier: I think I ended up changing my mind on this halfway through the review and forgot to revise the comment. If we just returned unused topics in the response from an unsubscription, we wouldn't need this at all.

lib/carrier/messaging/tracker.ex
kevsmith: After thinking about this some I'm not sure the simplification is worth loss of efficiency. I'm assuming `find_subscribers/2` will be called very frequently.

iiuc, I think using an empty list to represent an empty result is potentially inefficient by performing unnecessary allocations and collections:

```elixir
Enum.flat_map(subs, &(find_matching_subscriptions(&1, topic)))

defp find_matching_subscriptions({_, {matcher, subscribed}}, topic) do
  if Regex.match?(matcher, topic) do
    subscribed
  else
    # This is the inefficient bit as it allocates an empty list here
    # and a new list in the caller. The caller's list allocation is required
    #  to handle appending the returned from here to the end of the accumulator
    []
  end
end
```

In contract `Enum.reduce` incurs the allocation and subsequent collection overhead only when a topic is matched.

lib/carrier/messaging/tracker.ex
kevsmith: Derp. I had this in the initial version.

Issue #1187, opened 2016-11-29
Ignore bare `!` messages


agis-:
Typing just `!` in Slack means most likely a regular user message, not a command invocation. So I think it would make sense for Cog to completely skip such lines.

Example:

![image](https://cloud.githubusercontent.com/assets/827224/20715487/08c22e56-b658-11e6-8a17-34f3608c7bc9.png)

In the same manner, maybe things like `!!!` and `!!!!` should be ignored too.

PR #1293, opened 2017-01-06
Ignore chat message that are only a command prefix

vanstee:
Fixes https://github.com/operable/cog/issues/1187


vanstee:
Still have an open question about how to handle `!!!` (command prefix + previous command token). Will be fixed in another PR if needed.

PR #1294, opened 2017-01-07
Tidy up pipeline code

kevsmith:
* Migrates all usage of `adapter` to the newer `provider` model
* Reorganizes output helpers and formatters into a common namespace


vanstee:
The `Output` namespace module is nice :+1:

lib/cog/command/permissions_cache.ex
```diff
@@ -3,14 +3,16 @@ defmodule Cog.Command.PermissionsCache do
   use GenServer
   require Logger
 
+  alias Cog.Util.TimeHelpers
```
vanstee: Wonder if we can remove most of this in favor of Elixir's `DateTime` module.

lib/cog/bundle/warehouse.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Cog.BundleRegistry do
+defmodule Cog.Bundle.Warehouse do
```
vanstee: :+1:

PR #1305, opened 2017-01-17
Remodel pipelines w/GenStage

kevsmith:
This PR introduces a new pipeline execution subsystem based on `GenStage`. The new design intends to be much more modular than `Cog.Command.Pipeline.Executor` and better separate various pipeline execution concerns like command execution, output generation, and error handling.


lib/carrier/messaging/connection.ex
```diff
@@ -122,39 +122,41 @@ defmodule Carrier.Messaging.Connection do
...
+      # hasn't yet established a network connection to the message bus. By
+      # ensuring that we only return after the process is actually connected,
+      # we can simplify startup of processes that require a message bus
+      # connection.
+      #
+      # It also means that those clients don't have to know details about
+      # emqttc (like the structure of the "connected" message), so fewer
+      # implementation details about our choice of message bus don't leak out.
+      #
+      # If we don't connect after a specified timeout, we just fail.
```
christophermaier: Does this comment really apply anymore, given the restructuring?

lib/carrier/messaging/connection.ex
```diff
@@ -122,39 +122,41 @@ defmodule Carrier.Messaging.Connection do
...
+      receive do
+        {:mqttc, ^conn, :connected} ->
+          # Detect when the connection owner has exited before the connection
+          # is fully setup and exit. Prevents long stack traces from polluting
+          # log output.
```
christophermaier: How does the code in this branch do that?

lib/carrier/messaging/connection.ex
```diff
@@ -248,6 +250,10 @@ defmodule Carrier.Messaging.Connection do
     {:noreply, state}
   end
 
+  def terminate(_reason, _state) do
+    Logger.debug("MQTT connection #{inspect self()} closed")
```
christophermaier: Would it be useful to incorporate the reason into this message?

lib/cog/messages.ex
```diff
@@ -84,6 +86,18 @@ defmodule Cog.Messages.Command do
...
   field :user, :map, required: true
+
+  @doc "Creates a new Command from an invocation, options, and args"
+  defmacro create(invocation, options, args) do
+    quote bind_quoted: [invocation: invocation, options: options, args: args] do
+      %Cog.Messages.Command{command: invocation.meta.full_command_name,
+                            options: options,
+                            args: args,
+                            invocation_id: invocation.id,
+                            services_root: ServiceEndpoint.url()}
```
christophermaier: Is this a macro just to avoid repeatedly calling `ServiceEndpoint.url()`?

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,337 @@
...
+  Tells a prepared pipeline to begin executing
+  """
+  def run(pipeline) do
+    GenServer.call(pipeline, :run, :infinity)
+  end
+
+  @doc """
+  Notifies pipeline when processing is finished
+  """
+  def notify(pipeline) do
```
christophermaier: "notify" seems like a general concept, but "teardown" is specific. Should this function be renamed to "teardown" instead?

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,337 @@
...
+    end
+  end
+
+  def handle_call(:run, _from, state) do
+    state = %{state | started: DateTime.utc_now()}
+    case fetch_user(state) do
+      {:ok, user} ->
+        with {:ok, parsed, destinations} <- parse(user, state),
+             {:ok, perms} <- PermissionsCache.fetch(user),
+               user_json <- EctoJson.render(user) do
```
christophermaier: Whoa, we're still using `EctoJson`?

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,337 @@
...
+          start_pipeline(parsed, destinations, user_json, perms, state)
+        else
+          {:error, {:parse_error, message}} -> start_error_pipeline({:error, :parse_error, message}, user, state)
```
christophermaier: Can we put these clauses on a separate line from the conditions for readability?

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,337 @@
...
+      error ->
+        start_error_pipeline(error, nil, state)
+    end
+  end
+  def handle_call(_msg, _from, state) do
+    {:reply, :ignored, state}
+  end
+
+  def handle_cast(:teardown, state) do
+    duration = DateTime.to_unix(DateTime.utc_now, :milliseconds) - DateTime.to_unix(state.started, :milliseconds)
```
christophermaier: Any way we could use `Cog.Events.Util.elapsed/2` here instead? As coded, though, that would turn this from milliseconds to microseconds; not sure if that's better or not :man_shrugging:

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,337 @@
...
+    Enum.each(state.stages, &(Process.send(&1, {:pipeline_complete, self()}, [])))
+    {:noreply, %{state | status: :done}}
+  end
+  def handle_cast(_, state) do
+    {:noreply, state}
+  end
+
+  # Close bus connection and exit after last stage exits
+  # This avoids closing the connection while one of the sinks is still using it
+  def handle_info({:DOWN, _, _, stage, _}, %__MODULE__{status: :done, stages: [stage]}=state) do
```
christophermaier: Ah, cool 👍

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,337 @@
...
+    Logger.debug("Pipeline #{state.request.id} terminated")
+    Connection.disconnect(state.conn)
+    {:stop, :normal, %{state | stages: []}}
+  end
+  def handle_info({:DOWN, _, _, stage, _}, %__MODULE__{status: :done, stages: stages}=state) do
+    {:noreply, %{state | stages: List.delete(stages, stage)}}
```
christophermaier: Would there ever be a situation where the `stage` isn't a member of `stages`? Doesn't look like that would really break things here, but if it could happen, it may be worth logging at least.

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,337 @@
...
+  end
+  # Change pipeline state to :done if a stage crashes
+  def handle_info({:DOWN, _, _, stage, reason}, %__MODULE__{status: :running}=state) do
+    if stage in state.stages do
+      message = case reason do
+                  {_, stacktrace} ->
+                    "Pipeline #{state.request.id} crashed: #{inspect stacktrace, pretty: true}"
+                  _ ->
+                    "Pipeline #{state.request.id} crashed"
```
christophermaier: Can we add the reason to this log message? Or can it be a potentially large blob of data?

lib/cog/pipeline/relay_selector.ex
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  @doc "Returns the name of the selected Relay"
+  @spec relay(RelaySelector.t) :: String.t | nil
+  def relay(%__MODULE__{relay: relay}), do: relay
+
+  @doc "Constructs a MQTT topic for the named command"
+  @spec relay_topic(RelaySelector.t, String.t) :: String.t
+  def relay_topic(%__MODULE__{relay: nil}, _command_name), do: ""
```
christophermaier: How does this get called?

lib/cog/pipeline/signals.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Cog.Pipeline.DoneSignal do
+  defstruct [:error, :invocation, :bundle_version_id, :template]
+
+  def error?(%__MODULE__{error: nil}), do: false
+  def error?(%__MODULE__{}), do: true
+
+  def done?(%__MODULE__{}), do: true
+  def done?(_), do: false
```
christophermaier: This is a catch-all for "other signals", right?

If so, does `error?/1` need to have a similar catch-all function head?

lib/cog/pipeline/util.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Cog.Pipeline.Util do
+
+  alias Cog.Pipeline.Destination
+
+  def here_destination(request) do
```
christophermaier: We don't need the output classification logic anymore with the new pipeline structure, correct?

The name is slightly misleading, given that we're actually returning a list (even though it only has one item).

Could this just be a function in `Cog.Pipeline.Destination` directly?

lib/carrier/messaging/connection.ex
kevsmith: Probably, yeah.

lib/cog/pipeline.ex
kevsmith: Yup. The code was lifted as-is from the Executor modulo recasting it into a `with` statement.

lib/cog/pipeline.ex
kevsmith: I couldn't convince myself that would ever happen so I didn't add that defensive code originally. The only monitors `Pipeline` creates are on the stages it manages so it should never receive "rouge" `:DOWN` messages.

lib/cog/pipeline.ex
kevsmith: That's the problem. It's huge. At least 2 screens worth of text. Could be handled by adding `String.Chars` impls for a few structs. I could tackle that in a separate PR.

lib/cog/pipeline/relay_selector.ex
kevsmith: The sequence of using `RelaySelector` goes like this:

```
# Create a selector for a given bundle and version
selector = RelaySelector.new("my_bundle", "1.3.1")
# Ask the selector to select a relay
{:ok, selector} = RelaySelector.select(selector)
# Build the topic for the selected relay
topic = RelaySelector.relay_topic(selector)
```
If the caller creates a selector and calls `relay_topic/2` without first calling `select/1` then there's no relay from which to build a topic string.

lib/cog/pipeline/util.ex
kevsmith: We still need it so we send correctly formatted output. `:chat`, `:status`, and `:status_only` outputs are all different to varying degrees.

Code was originally moved to `Util` as part of a previous refactoring. I'll see if it can be moved into `Destination` directly.

lib/cog/pipeline.ex
kevsmith: I thought microseconds was overkill when reporting on pipeline execution. Maybe we need to teach `elapsed/2` how to handle caller-specified units?

lib/carrier/messaging/connection.ex
kevsmith: Yes. If the `after` clause of the `receive` block (starting on line 142) executes then creating the connection process will fail and return `:econnrefused`.

lib/cog/pipeline.ex
christophermaier: Sounds good

lib/cog/messages.ex
kevsmith: It was to clean up the code a bit without obscuring intent. It was originally a function but due to how `Conduit` does code generation it had to be a macro.

lib/cog/pipeline/relay_selector.ex
christophermaier: How does the caller behave when they get an empty topic string?

lib/cog/pipeline.ex
```diff
@@ -0,0 +1,339 @@
...
+              e in PrepareError ->
+                start_error_pipeline({:error, Exception.message(e)}, user, state)
+            end
+          error ->
+            start_error_pipeline(error, user, state)
+        end
+      error ->
+        start_error_pipeline(error, user, state)
+    end
+  end
```
vanstee: Guessing this was moved over form the old executor, but could use some `with` magic.

PR #1308, opened 2017-01-17
Dump and load schema instead of applying migrations

vanstee:
We now store the database as a schema dump in the repo. This accomplishes a couple of things:

* We can always easily get the schema into the current state by loading a schema instead of applying a directory full of migrations. It's also a bit faster and quieter in the logs. Oh and no need to drop the db when reseting.
* When we commit a migration the schema is updated, which means we can see the resulting state of the schema as part of a PR.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,7 +1,7 @@
 FROM operable/elixir:1.3.4-r0
 
 # Greenbar-only compilation dependencies
-RUN apk -U add expat-dev gcc g++ libstdc++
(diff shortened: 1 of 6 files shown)


vanstee:
@christophermaier Fixed that in `scripts/docker-start` where we only load the schema if the table doesn't exist.

vanstee:
Just realized this wont work due to the data we populate in our migrations like option types and such. We'd need to add a seeds file as well if we wanted to continue down this path. Gonna close for now and revisit if needed later on.

PR #1309, opened 2017-01-18
Protect the last user in the cog-admin group

vanstee:
Now that the admin user can have any username, change our cog-admin
group protection trigger to prohibit removing the last member rather
than specifically checking for the "admin" username.

### DIFF
--- a/lib/cog/models/user.ex
+++ b/lib/cog/models/user.ex
@@ -136,8 +136,8 @@ defimpl Groupable, for: Cog.Models.User do
         case error do
           # Handle exception from protected_admin_group_memberships trigger
           %Postgrex.Error{postgres: %{code: :raise_exception,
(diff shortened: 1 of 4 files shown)


lib/cog/models/user.ex
```diff
@@ -136,8 +136,8 @@ defimpl Groupable, for: Cog.Models.User do
         case error do
           # Handle exception from protected_admin_group_memberships trigger
           %Postgrex.Error{postgres: %{code: :raise_exception,
-                                      message: "cannot remove admin user from cog-admin group"}} ->
-            {:error, :cannot_remove_admin_user}
+                                      message: "cannot remove last user from cog-admin group"}} ->
+            {:error, :cannot_remove_last_admin_user}
```
kevsmith: Does this case  have any tests?

vanstee: Added a test. Thanks.

PR #1312, opened 2017-01-19
Threads

vanstee:
Probably not going to end up merging this, but figured I'd open a PR for some discussion.

This PR makes Cog respond to the message it was sent with a new thread response. All we need to do is set `thread_ts` with the value of `ts` from the original message. There are still some issues on Slack's end to work out, like attachments missing from responses and such. Also on our end we need to disable this when redirecting to another room since threads can only be created and responded to in a single room. Either way, this was a cool experiment and was really only a minimal amount of work.

screen shot 2017-01-18 at 8 45 31 pm

### DIFF
--- a/lib/cog/chat/adapter.ex
+++ b/lib/cog/chat/adapter.ex
@@ -228,6 +228,13 @@ defmodule Cog.Chat.Adapter do
                                                     "message" => message,
                                                     "provider" => provider,
                                                     "metadata" => metadata}, state) do
+    metadata = case Cog.Chat.MessageMetadata.from_map(metadata) do
+      {:ok, metadata} ->
+        metadata
+      _error ->
+        %Cog.Chat.MessageMetadata{}
+    end
+
     case with_provider(provider, state, :send_message, [target, message, metadata]) do
       :ok ->
         :ok
@@ -236,6 +243,7 @@ defmodule Cog.Chat.Adapter do
       {:error, reason} ->
         Logger.error("Failed to send message to provider #{provider}: #{inspect reason, pretty: true}")
     end
+
     {:noreply, state}
   end
   def handle_cast(_conn, @incoming_topic, "event", event, state) do
--- a/lib/cog/chat/adapter.ex
+++ b/lib/cog/chat/adapter.ex
@@ -228,10 +228,15 @@ defmodule Cog.Chat.Adapter do
                                                     "message" => message,
                                                     "provider" => provider,
                                                     "metadata" => metadata}, state) do
-    metadata = case Cog.Chat.MessageMetadata.from_map(metadata) do
-      {:ok, metadata} ->
-        metadata
-      _error ->
+    metadata = case metadata do
+      map when is_map(map) ->
+        case Cog.Chat.MessageMetadata.from_map(metadata) do
+          {:ok, metadata} ->
+            metadata
+          _error ->
+          %Cog.Chat.MessageMetadata{}
+        end
+      _ ->
         %Cog.Chat.MessageMetadata{}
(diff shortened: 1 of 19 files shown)


lib/cog/chat/adapter.ex
```diff
@@ -139,19 +139,19 @@ defmodule Cog.Chat.Adapter do
     result
   end
 
-  def send(provider, target, message) do
+  def send(provider, target, message, thread_id) do
```
kevsmith: What would you think about replacing `thread_id` with an options keyword list?

`def send(provider, target, message, opts \\ [])`

lib/cog/chat/message.ex
```diff
@@ -7,6 +7,7 @@ defmodule Cog.Chat.Message do
   field :user, [object: Cog.Chat.User], required: true
   field :text, :string, required: true
   field :provider, :string, required: true
+  field :thread_id, :string, required: false
```
kevsmith: Similar to above I'm thinking we should replace `:thread_id` with a `metadata` module and field like so:

`field :metadata, [object: Cog.Chat.MessageMetadata], required: false`

lib/cog/chat/adapter.ex
vanstee: Good idea. To keep our api surface area down it makes sense to always send our metadata field here whether or not it's empty. We can ignore it in the non-Slack adapters for now.

vanstee:
* Moved thread_id to metadata included in each message
* Added an env var `SLACK_ENABLE_THREADED_RESPONSES` to enable thread responses (disabled by default) as there are still some issues on Slack's end (rendering attachments are broken, dealing with `thread_ts` that can't be found, scrolling down and acking notifications related to thread responses).
* Fixed a bug where thread responses broke when redirecting to other rooms

lib/cog/chat/message_metadata.ex
```diff
@@ -0,0 +1,6 @@
+defmodule Cog.Chat.MessageMetadata do
```
christophermaier: Some documentation on this module in particular would be nice, emphasizing that (for now, at least) it's a bit of a grab-bag of "stuff" that only really applies to Slack, and not more broadly.

On further reflection, I wonder if we're perhaps over-architecting things by making this a Conduit-validated module, particularly given its grab-bag nature. Maybe it should just be a keyword list until a broader and more general use case becomes apparent?

lib/cog/chat/adapter.ex
```diff
@@ -226,15 +226,29 @@ defmodule Cog.Chat.Adapter do
...
-                                                    "provider" => provider}, state) do
-    case with_provider(provider, state, :send_message, [target, message]) do
+                                                    "provider" => provider,
+                                                    "metadata" => metadata}, state) do
+    metadata = case metadata do
+      map when is_map(map) ->
+        case Cog.Chat.MessageMetadata.from_map(metadata) do
+          {:ok, metadata} ->
+            metadata
+          _error ->
```
kevsmith: Conversion error should be logged.

lib/cog/chat/adapter.ex
```diff
@@ -226,15 +226,29 @@ defmodule Cog.Chat.Adapter do
...
+          %Cog.Chat.MessageMetadata{}
+        end
+      _ ->
+        %Cog.Chat.MessageMetadata{}
```
kevsmith: Should log that we're ignoring non-map metadata so future us doesn't go insane trying to figure out why metadata isn't getting included :)

vanstee:
Added logging and comments about the metadata struct.

PR #1313, opened 2017-01-19
group/role api refactor

mpeck:
This change just does some preliminary cleanup on the role and group controllers.

Removes the `groups` field from the groups json response. Since groups should no longer have other groups as members.

Moves the query logic from the role controller and into the role repository to more closely match the other resources.


lib/cog/repository/groups.ex
```diff
@@ -116,11 +116,36 @@ defmodule Cog.Repository.Groups do
...
+          {:ok, updated} ->
+            updated
+          {:error, error} ->
+            Repo.rollback(error)
+        end
+      else
+        result
+      end
+
+    end)
```
kevsmith: I think moving the logic to a named function would improve readability.

web/controllers/v1/group_controller.ex
```diff
@@ -61,6 +66,18 @@ defmodule Cog.V1.GroupController do
         conn
         |> put_status(:bad_request)
         |> json(%{errors: "Bad ID format"})
+      {:error, {:not_found, {key, names}}} ->
+        conn
+        |> put_status(:unprocessable_entity)
```
kevsmith: What entity does the unknown key represent here?

web/controllers/v1/role_controller.ex
```diff
@@ -2,32 +2,28 @@ defmodule Cog.V1.RoleController do
...
+  alias Cog.Repository.Roles
 
   plug Cog.Plug.Authentication
   plug Cog.Plug.Authorization, permission: "#{Cog.Util.Misc.embedded_bundle}:manage_roles"
 
   # Search by name only for now
   def index(conn, %{"name" => name}) do
-    role = Repo.get_by!(Role, name: name)
-    |> Repo.preload([[permissions: :bundle], [group_grants: :group]])
+    role = Roles.by_name(name)
```
kevsmith: Yay for repos! 👍

web/router.ex
```diff
@@ -25,8 +25,8 @@ defmodule Cog.Router do
     resources "/v1/users/reset-password", V1.PasswordResetController, only: [:create, :update]
 
     resources "/v1/groups", V1.GroupController
-    get "/v1/groups/:id/users", V1.GroupMembershipController, :index
-    post "/v1/groups/:id/users", V1.GroupMembershipController, :manage_group_users
+    get "/v1/groups/:id/users", V1.GroupController, :show
+    post "/v1/groups/:id/users", V1.GroupController, :manage_group_users
```
kevsmith: Do these route changes impact cogctl or flywheel?

mpeck: They shouldn't. The index function in the membership controller was an exact duplicate of the show function in the group controller and the manage_group_users function is just routing the logic through the update function. The api should remain unchanged.

My thinking with routing the group users bits through the update function was that we could eventually drop the extra routes with a future update. But for now everything should function the same as before.

PR #1319, opened 2017-01-20
remove group perms

mpeck:
Removes support for direct permissions on groups


priv/repo/migrations/20170120211447_drop_group_permissions.exs
```diff
@@ -0,0 +1,149 @@
...
+      WITH RECURSIVE
+      role_grants AS (
+        SELECT gr.group_id
+        FROM group_roles AS gr
+        JOIN role_permissions AS rp
+          USING (role_id)
+        WHERE rp.permission_id = p_permission
+      )
+      SELECT group_id
+       FROM role_grants;
```
vanstee: Does this still need to be `RECURSIVE`? I think we might be able to simplify this to the following as we don't need to reselect group id's:

```
SELECT group_id FROM group_roles
  JOIN role_permissions USING (role_id)
  WHERE role_permissions.permission_id = p_permission
```

PR #1321, opened 2017-01-23
group many-to-many associations

mpeck:
This just changes the user and role associations to many-to-many, which reduces some of the extra unnecessary bits on the model.

### DIFF
--- a/lib/cog/models/group.ex
+++ b/lib/cog/models/group.ex
@@ -4,17 +4,16 @@ defmodule Cog.Models.Group do
 
   alias Cog.Models.GroupRole
   alias Cog.Models.UserGroupMembership
+  alias Cog.Models.User
+  alias Cog.Models.Role
 
   alias Ecto.Changeset
 
   schema "groups" do
     field :name, :string
 
-    has_many :user_membership, UserGroupMembership, foreign_key: :group_id
-    has_many :direct_user_members, through: [:user_membership, :member]
-
-    has_many :role_grants, GroupRole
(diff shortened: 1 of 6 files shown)


lib/cog/models/group.ex
```diff
@@ -4,17 +4,16 @@ defmodule Cog.Models.Group do
...
 
   schema "groups" do
     field :name, :string
 
-    has_many :user_membership, UserGroupMembership, foreign_key: :group_id
-    has_many :direct_user_members, through: [:user_membership, :member]
-
-    has_many :role_grants, GroupRole
-    has_many :roles, through: [:role_grants, :role]
+    many_to_many :users, User, join_through: UserGroupMembership, join_keys: [group_id: :id, member_id: :id]
```
vanstee: Wonder if `s/member_id/user_id/` would make things even nicer here and throughout the codebase.

mpeck: It would require a migration which is why I didn't do it. But maybe that shouldn't stop me.

vanstee: Meh, whatever is good. Not a huge deal.

mpeck: I don't think the `member_id` is really exposed anywhere other than here, but for consistency with the db I think you're right. We should just go ahead and change it.

mpeck: ugh, so because of the way the groupable bits are implemented I think changing this to `user_id` would involve a bigger refactor. I'm gonna leave it for now and we can revisit if it ends up being too confusing for folks.

Issue #794, opened 2016-06-23
Guard against cascading deletes


christophermaier:
When a user deletes a role that is currently assigned to one or more groups, we would like to prevent the deletion, as this would end up potentially removing access to different parts of the system for many users.

Currently our database constraints are all set to `CASCADE`; we could switch these to `RESTRICT` to have the database do this for us.

We could add "force" functionality to remove referencing rows first, or users could craft pipelines to do the same (conversation in internal chat leans toward the latter, currently).

Though the example given is for roles assigned to groups, similar logic should hold for all similar relationships.

PR #1322, opened 2017-01-24
Restrict deletion on group and relay group relations

vanstee:
Instead of restricting on delete for all foreign key relations, it made sense to only restrict relations that can be manually modified via the api. The main exception is the tree of bundle relations, so when a bundle is deleted we still cascade and delete all associated child records. To cleanly delete bundles, permissions are unlinked from any roles they had been granted to first. All exceptions raised are returned as errors on the id attribute since there's no other key that would really make sense to hang it off of.

Closes https://github.com/operable/cog/issues/794

### DIFF
--- a/test/cog/bootstrap_test.exs
+++ b/test/cog/bootstrap_test.exs
@@ -100,7 +100,8 @@ defmodule Cog.Bootstrap.Test do
 
   test "the embedded bundle cannot be deleted" do
     try do
-      Cog.Models.Bundle |> Cog.Repo.delete_all
+      embedded = Repo.get_by!(Bundle, name: Cog.Util.Misc.embedded_bundle)
+      Cog.Repository.Bundles.delete(embedded)
     rescue
       exception ->
         assert exception.postgres.message == "cannot modify embedded bundle"
--- a/test/cog/bootstrap_test.exs
+++ b/test/cog/bootstrap_test.exs
@@ -1,10 +1,6 @@
 defmodule Cog.Bootstrap.Test do
   use Cog.ModelCase
-
-  alias Cog.Models.Bundle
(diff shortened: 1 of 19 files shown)


lib/cog/models/relay_group_assignment.ex
```diff
@@ -10,12 +10,10 @@ defmodule Cog.Models.RelayGroupAssignment do
   end
 
   @required_fields ~w(bundle_id group_id)
-  @optional_fields ~w()
 
   def changeset(model, params) do
     model
-    |> cast(params, @required_fields, @optional_fields)
-    |> foreign_key_constraint(:bundle_id, name: :relay_group_assignments_bundle_id_fkey)
+    |> cast(params, @required_fields)
```
vanstee: Removed this by mistake. Let me revert these changes.

test/cog/bootstrap_test.exs
```diff
@@ -78,31 +74,34 @@ defmodule Cog.Bootstrap.Test do
...
     Cog.Bootstrap.bootstrap(@admin_user)
-    assert Cog.Models.Group |> Cog.Repo.one! |> Map.get(:name) == Cog.Util.Misc.admin_group
+    assert Group |> Repo.one! |> Map.get(:name) == Cog.Util.Misc.admin_group
     assert Cog.Bootstrap.is_bootstrapped?
   end
 
   test "the admin group cannot be deleted" do
     try do
-      Cog.Models.Group |> Cog.Repo.delete_all
+      admin_group = Repo.get_by!(Group, name: "cog-admin")
```
christophermaier: Should use `Cog.Util.Misc.admin_group` here for consistency.

PR #1324, opened 2017-01-25
users can no longer have direct perms or roles

mpeck:


lib/cog/support/model_utilities.ex
```diff
@@ -140,6 +140,14 @@ defmodule Cog.Support.ModelUtilities do
   Grant a permission to an implementor of `Permittable`, and return
   the implementor.
   """
```
christophermaier: Should probably change the docstring, since it's now outdated.

PR #1331, opened 2017-01-27
fix group chat command

### LINKED ISSUE #1332: group command does not return associated users

mpeck:
Replaces the users key with a members key when encoding a group to json. Fixes the bug with group info not returning associated members.

resolves #1332

### DIFF
--- a/lib/cog/models/group.ex
+++ b/lib/cog/models/group.ex
@@ -91,11 +91,10 @@ end
 # Currently we have one scheme for the chat command and another for the api.
 defimpl Poison.Encoder, for: Cog.Models.Group do
   def encode(struct, options) do
-    map = struct
+    struct
     |> Map.from_struct
     |> Map.take([:id, :name, :roles])
     |> Map.put(:members, struct.users)
-
-    Poison.Encoder.Map.encode(map, options)
+    |> Poison.Encoder.Map.encode(options)
   end
 end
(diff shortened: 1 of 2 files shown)


lib/cog/models/group.ex
```diff
@@ -86,11 +86,15 @@ defimpl Permittable, for: Cog.Models.Group do
...
+# Currently we have one scheme for the chat command and another for the api.
 defimpl Poison.Encoder, for: Cog.Models.Group do
   def encode(struct, options) do
     map = struct
     |> Map.from_struct
-    |> Map.take([:id, :name, :roles, :users])
+    |> Map.take([:id, :name, :roles])
+    |> Map.put(:members, struct.users)
 
     Poison.Encoder.Map.encode(map, options)
```
vanstee: Might as well just tack this on to the pipeline.

```elixir
struct
|> Map.from_struct
|> Map.take([:id, :name, :roles])
|> Map.put(:members, struct.users)
|> Poison.Encoder.Map.encode(options)
```

test/commands/group/info_test.exs
```diff
@@ -11,6 +11,8 @@ defmodule Cog.Test.Commands.Group.InfoTest do
       new_req(args: ["info_group"])
       |> send_req(Info)
 
-    assert(%{name: "info_group"} = response)
+    assert(%{name: "info_group",
+             roles: [],
+             members: []} = response)
```
vanstee: Might be nice to check this rendered with some actual members.

PR #1338, opened 2017-01-30
Fixed pipeline history range parsing errors

### LINKED ISSUE #1337: Parsing history index ranges fails
Running the `history` command with a single argument crashes with `FunctionClauseError`.

kevsmith:
Fixes #1337

### DIFF
--- a/lib/cog/commands/history.ex
+++ b/lib/cog/commands/history.ex
@@ -35,25 +35,26 @@ defmodule Cog.Commands.History do
   end
 
   defp parse_args([]), do: {:ok, {nil, nil}}
-  defp parse_args([hist_start]) do
+  defp parse_args([hist_start]) when is_binary(hist_start) do
     case String.split(hist_start, "-") do
       [hist_start] ->
         with {:ok, parsed} <- parse_int(hist_start),
           do: {:ok, {parsed, nil}}
-      [hist_start, "-"] ->
-        with {:ok, parsed} <- parse_int(hist_start),
-          do: {:ok, parsed, nil}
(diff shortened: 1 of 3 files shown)


lib/cog/repository/pipeline_history.ex
```diff
@@ -77,6 +77,7 @@ defmodule Cog.Repository.PipelineHistory do
     query  = build_user_history_query(user_id, hist_start, hist_end, limit)
     query = from q in query,
             where: q.state == "finished"
+    IO.inspect query
```
vanstee: Guessing this is left over debugging stuffs.

lib/cog/commands/history.ex
```diff
@@ -35,25 +35,26 @@ defmodule Cog.Commands.History do
   end
 
   defp parse_args([]), do: {:ok, {nil, nil}}
-  defp parse_args([hist_start]) do
+  defp parse_args([hist_start]) when is_binary(hist_start) do
```
vanstee: Totally not a big deal, but it probably makes sense to call this parameter a `hist_range`

lib/cog/commands/history.ex
```diff
@@ -65,6 +66,9 @@ defmodule Cog.Commands.History do
     with {:ok, hist_end} <- parse_int(hist_end),
       do: {:ok, {nil, hist_end}}
   end
```
vanstee: Are these function heads that include `"-"` still used? It doesn't look like the parser will parse `-` as a command arg and i couldn't figure out how else this was called.

PR #1348, opened 2017-02-13
Replace escaped double quotes with double quotes

### LINKED ISSUE #1323: Parsing escaped quotes in json fails
`@cog seed "[{\"test\": \"test\"}]"` fails, when it should succeed.

kevsmith:
Fixes `Cog.Command.Seed` crash triggered when given a string containing escaped double quotes.

Fixes #1323


lib/cog/commands/seed.ex
```diff
@@ -30,6 +30,7 @@ defmodule Cog.Command.Seed do
   def handle_message(%Command{args: [input]}=req, state) when not(is_binary(input)),
     do: {:error, req.reply_to, "Argument must be a string", state}
   def handle_message(%Command{args: [input]}=req, state) do
+    input = String.replace(input, ~r/\\\"/, "\"")
```
vanstee: Do we also need to do this for single quotes?

PR #1349, opened 2017-02-13
Apply constraints when deleting relay groups

christophermaier:


christophermaier:
@mpeck I think we wanted to avoid making big changes to the API so close to 1.0. In any event, cogctl now has some commands to help with this (e.g. `cogctl relay group remove $GROUP --all` will remove all relays from a group, and `cogctl relay group unassign $GROUP --all` unassigns all bundles from a group. If we wanted to, we could easily add a `--force` flag to `cogctl relay group delete $GROUP` without touching the API.

This change was to ensure that the API didn't throw a HTTP 500 when you tried to delete a group.

PR #1359, opened 2017-02-17
add config to let it run on heroku

cullenmcdermott:
I was trying to get Cog to run on Heroku and hit a few roadblocks which this should fix.

mix.lock was still stuck on an old version of greenbar_markdown which wouldn't build on Heroku.

The Procfile and elixir_buildpack.config should let it run after everything compiles. You'll also need to attach Heroku Postgres to the dyno and set your chat environment variables.

The change to config.exs is because Heroku will bind your app to a random port and sets a $PORT environment variable. There's probably a better way to do this as this might cause confusion since it adds yet another port environment variable. Without letting it bind to $PORT Heroku will think it did not come up successfully and it'll kill the Dyno with error R10.


cullenmcdermott:
After thinking about the port thing a little more, it might be best to run a reverse proxy in front of this as Heroku will only let you listen on 80 and 443 as far as I know. Would Cog work behind a reverse proxy?

davejlong:
@cullenmcdermott Yes, Cog works against a remote proxy. Here is the Nginx config map that we use in our Kubernetes Helm chart.

What makes Cog difficult to run on something like Heroku is that the application at port 4000 is just the base API. There's also the Triggers API which typically runs on port 4001 and the Service API which runs on port 4002.

PR #1360, opened 2017-02-17
Refactor the Dockerfile

christophermaier:
Streamline the Dockerfile, getting rid of things we never used, removing some unnecessary and costly operations, and leverage the Docker cache for quicker builds.

It's also smaller!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ COPY . /home/operable/cog/
 RUN chown -R operable /home/operable
 
 RUN apk update -U && \
-    apk add expat-dev gcc g++ libstdc++ make && \
+    apk add expat-dev gcc g++ libstdc++ && \
     mix clean && mix deps.get && mix compile && \
     apk del gcc g++ && \
     rm -f /var/cache/apk/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,15 +2,8 @@ FROM operable/elixir:1.3.4-r0
 
 ENV MIX_ENV prod
 
-# Setup Operable user. UID/GID default to 60000 but can be overriden.
-ARG OPERABLE_UID
-ENV OPERABLE_UID ${OPERABLE_UID:-60000}
-
-ARG OPERABLE_GID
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,32 +1,41 @@
...
-    apk del gcc g++ && \
-    rm -f /var/cache/apk/*
+# Really, we only need the cog directory to be owned by operable,
+# because (by default) that's where we write log files. None of the
+# actual scripts or library files need to be owned by operable.
+RUN chown -R operable /home/operable/cog
+
+COPY mix.exs mix.lock /home/operable/cog/
+COPY config/ /home/operable/cog/config/
+RUN mix deps.get
```
vanstee: You may want to add `--only prod` here. We can also skip the archive check if you want, but not as big of a deal.

Dockerfile
```diff
@@ -1,32 +1,41 @@
...
+        libstdc++ && \
+    mix deps.compile && \
+    apk del .build_deps
+
+COPY emqttd_plugins/ /home/operable/cog/emqttd_plugins/
+COPY priv/ /home/operable/cog/priv/
+COPY web/ /home/operable/cog/web/
+COPY lib/ /home/operable/cog/lib/
+
+RUN mix compile
```
vanstee: To speed things up you could add `--no-deps-check` and `--no-archive-check` here.

Makefile
```diff
@@ -13,7 +13,6 @@ reset-db:
 docker:
 	scripts/get_cogctl.sh
 	docker build \
-		   --build-arg MIX_ENV=prod \
 		   -t $(DOCKER_IMAGE) .
```
vanstee: Totally just a minor style thing, but this might be easier to read on a single line.

Dockerfile
christophermaier: 👍

Dockerfile
christophermaier: 👍

Makefile
christophermaier: 👍

PR #1365, opened 2017-02-21
Fix history command bugs

### LINKED ISSUE #1364: `history -l 1` crashes
We need to make sure we can return a single value from history. Also we should check the limit to make sure it isn't a negative number.

screen shot 2017-02-21 at 11 26 15 am

vanstee:
* Limit 1 returns the expected result
* Negative limits cause an abort with error message

Fixes #1364

### DIFF
--- a/lib/cog/commands/history.ex
+++ b/lib/cog/commands/history.ex
@@ -84,7 +84,6 @@ defmodule Cog.Commands.History do
   end
 
   defp format_entries([]), do: []
-  defp format_entries([entry]), do: format_entry(entry, 2)
   defp format_entries(entries) do
     max_width = find_max_width(entries)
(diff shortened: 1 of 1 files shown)


lib/cog/commands/history.ex
```diff
@@ -83,8 +83,16 @@ defmodule Cog.Commands.History do
     end
   end
 
+  defp ensure_positive_limit(limit) do
```
kevsmith: I think it's more concise to use multiple function heads here

```
defp ensure_positive_limit(n) when n < 0, 
  do: {:error, "Limit must be a positive integer"}
defp ensure_positive_limit(n) when n >= 0, do: {:ok, n}
```

vanstee: Ah good call.

PR #1370, opened 2017-02-22
secure pipeline command

### LINKED ISSUE #1327: Revisit pipeline command-related permissions
This was a left-over nice-to-have from #1320: 

```
Revisit permissions - Currently any user can view the pipeline history
for any other user. LIkewise any user can kill any other user's executing
pipelines. I'm not sure this is the desired behavior. I can make convincing 
arguments for both sides.
```

mpeck:
Adds a hardcoded permission to the pipeline command limiting who can manage pipelines to the owning user and users with the `operable:manage_user_pipeline` permission.

resolves #1327

### DIFF
--- a/lib/cog/commands/pipeline/list.ex
+++ b/lib/cog/commands/pipeline/list.ex
@@ -10,7 +10,10 @@ defmodule Cog.Commands.Pipeline.List do
   alias Cog.Repository.Permissions
   alias Cog.Models.User
 
-  @description "Display command pipeline statistics"
+  @description """
+  Display command pipeline statistics. Note: You can only list your own \
+  pipelines unless you have the operable:manage_user_pipeline permission.
+  """
   @valid_state_names ["R","W","F","running","waiting","finished"]
 
   # Allow any user to run ps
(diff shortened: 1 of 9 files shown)


lib/cog/commands/pipeline/list.ex
```diff
@@ -68,11 +75,32 @@ defmodule Cog.Commands.Pipeline.List do
...
+            # user, or the requesting user must have the manage_user_pipeline
+            # permission.
+            {:ok, req_user} = UserRepo.by_username(req.user["username"])
+            if app_user.id == req_user.id do
+              {:ok, HistoryRepo.pipelines_for_user(app_user.id, Map.get(opts, "last", 20))}
+            else
+              perm = Permissions.by_name("operable:manage_user_pipeline")
+              if User.has_permission(req_user, perm) do
+                {:ok, HistoryRepo.pipelines_for_user(app_user.id, Map.get(opts, "last", 20))}
+              else
```
christophermaier: Could we consolidate some of this? Seems like it's doing the same kind of checks as in other places in this PR (e.g. `if has_perm? || entry.user.id == req_user.id do...`)

PR #1381, opened 2017-03-19
Get Slack tests working again

### LINKED ISSUE #1377: Update Slack Integration Tests
Slack tests appear to be broken at the moment, with repeated sporadic failures across the suite. It appears as though Slack is re-sending older messages when our test client connects for a new test. Not quite sure yet if this is some change on Slack's side, or some kind of race condition in our test code. The functionality of Cog itself with respect to Slack appears sound, though; this appears to be   just a problem with the tests themselves.

christophermaier:
It appears that Slack is now sending recently-sent messages to a client
when it connects. That is, it's possible to receive messages sent before
you actually connect. This was causing lots of random failures in the
tests, because we would end up making assertions on messages sent in
response to previous tests! The actual functionality of the Slack
provider was fine; this was purely a testing issue.

Many changes were made to clean and tighten up the tests (detailed
below), but the big take-home is that we now pay attention to message
timestamps to determine whether to process a message or not.

An overview of what was done:

* Run Slack tests in Travis CI

* Create a Slack client once per Slack suite, as opposed to once per
  Slack *test*.

  Though not strictly necessary, this reduces the chances of running
  afoul of Slack's throttling, which results in shorter overall test
  runs.

* Update Slack tests to send test-specific messages.

  Previously, we'd send a lot of "echo blah" or "echo $TIMESTAMP"
  messages. Once Slack started apparently sending recently sent messages
  upon connection, this made for some hard-to-track-down false postives
  and timeouts. Now that every test sends unique (and traceable)
  messages, it's much easier to figure out what messages are being
  matched.

* Clean up and refactor the testing Slack client

  In the course of debugging the tests, I discovered extra code in the
  testing Slack client (and its `SlackClientState`) that wasn't actually
  being used. For instance, we would store messages, but not actually do
  anything with them. We also never join a specific channel when
  connecting our test client (that is taken care of when setting up the
  Slack accounts, anyway), so that code is out. Finally, we don't
  currently test anything with edited messages, so that functionality is
  out, too.

  Additionally, we now explicitly pay attention to the timestamp of
  messages. When we send a message, we note its timestamp and only
  process responses with a greater timestamp.

Fixes #1377

### DIFF
--- a/test/support/slack_client.ex
+++ b/test/support/slack_client.ex
@@ -1,54 +1,61 @@
 defmodule Cog.Test.Support.SlackClientState do
+  @moduledoc """
+  Additional state to track for a `Cog.Test.Support.SlackClient`
+  process. Ideally, we'd add this data to the client's state directly,
+  but the Elixir-Slack library doesn't allow arbitrary data in its
+  state. If it did, then this module could go away.
+  """
+  def start_link(),
+    do: Agent.start_link(&fresh_state/0, name: __MODULE__)
 
-  def start_link() do
-    Agent.start_link(fn -> fresh_state end, name: __MODULE__)
-  end
+  @doc """
+  Record a timestamp for a message. The client should only process
+  messages that have a timestamp greater than this one.
+  """
+  def set_cutoff_ts(ts),
+    do: Agent.update(__MODULE__, &Map.put(&1, :cutoff_ts, ts))
 
-  def store_message(sender, message) do
+  @doc """
+  Retrieve the current "cutoff" timestamp. The client should only
+  process messages that have a timestamp greater than this one.
+  """
+  def get_cutoff_ts(),
+    do: Agent.get(__MODULE__, &Map.get(&1, :cutoff_ts))
+
+  @doc """
+  Record a test process as waiting for a response from Slack.
+
+  `reply_from`: a Slack user. Only messages sent from this user are
+                considered.
+  `caller`: a tuple consisting of a unique ref and the test process
+            waiting for a response.
+  """
+  def add_waiter(reply_from, {ref, pid}=caller)
+  when is_reference(ref) and is_pid(pid) do
     Agent.update(__MODULE__,
-      fn(state) ->
-        messages = state
-        |> Map.get(:messages)
-        |> Map.update(sender, [message],
-          fn(previous) ->
-            Enum.sort([message|previous], &by_ts/2)
-          end)
-        Map.put(state, :messages, messages)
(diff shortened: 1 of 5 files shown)


test/support/slack_client.ex
```diff
@@ -112,93 +119,65 @@ defmodule Cog.Test.Support.SlackClient do
...
-        callers ->
-          reply = {:ok, msg}
-          Enum.each(callers, fn({ref, caller}) -> send(caller, {ref, reply}) end)
+      # The Elixir-Slack library converts Slack messages to atom-keyed
+      # maps instead of string-keyed ones :/
+      if String.to_float(message[:ts]) > SlackClientState.get_cutoff_ts() do
+        msg = parse_message(message, state)
+        waiters = SlackClientState.get_waiters(message.user)
+        if waiters do
+          Enum.each(waiters, fn({ref, caller}) ->
```
christophermaier: Function body is nested too deep (max depth is 2, was 4).

test/support/slack_client.ex
```diff
@@ -1,54 +1,61 @@
...
+  Record a test process as waiting for a response from Slack.
 
-      Map.put(state, :waiters, updated)
-    end)
+  `reply_from`: a Slack user. Only messages sent from this user are
+                considered.
+  `caller`: a tuple consisting of a unique ref and the test process
+            waiting for a response.
+  """
+  def add_waiter(reply_from, {ref, pid}=caller)
```
christophermaier: There are spaces around operators most of the time, but not here.

test/support/slack_client.ex
```diff
@@ -112,93 +119,65 @@ defmodule Cog.Test.Support.SlackClient do
...
-        send(sender, {ref, {:error, result["error"]}})
-        {:noreply, state}
-      true ->
-        user_id = find_user_id!(reply_from, state)
-        SlackClientState.add_waiter(user_id, caller)
-        {:noreply, state}
-    end
-  end
-  def handle_info({{ref, sender}=caller, {:chat_and_wait, room, message, reply_from, edited}}, state) do
+  def handle_info({{ref, sender}=caller, {:chat_and_wait, room, message, reply_from}}, state) do
```
christophermaier: There are spaces around operators most of the time, but not here.

christophermaier:
Ebert has finished reviewing this Pull Request and has found:

* 4 fixed issues! 🎉

You can see more details about this review at https://ebertapp.io/github/operable/cog/pulls/1381.

PR #1383, opened 2017-03-30
Push builds to cog-preview

nmohoric:
This will fix cog-preview at dockerhub.

This would need corresponding DOCKER_EMAIL, DOCKER_USER, and DOCKER_PASSWORD settings setup in travis. You'd also need to turn off automatic builds at dockerhub.

Just an idea, feel free to close if you'd like to go about this differently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ env:
 script:
   - docker-compose -f docker-compose.ci.yml run test ./scripts/wait-for-it.sh postgres:5432 -s -t 30 -- make test-$TEST
 
+after_success:
+  - if [ "$TRAVIS_BRANCH" == "master" ]; then
+    docker login -e="$DOCKER_EMAIL" -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD";
+    docker push operable/cog-preview;
+    fi
+
 notifications:
   slack:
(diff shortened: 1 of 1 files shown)


nmohoric:
I hadn't considered this is running a matrix, so it would upload a build for each run in the matrix.
It would also be uploading the test container which is probably not what we want.

There's really not a good way to only do this after all have completed.

This is probably closable and I can go back to the drawing board.

christophermaier:
@nmohoric Hrmm, yeah, that does complicate things... One thing I had considered is reworking the Docker build process such that it was again possible to use just a `docker build` to get an image. I think the main sticking point is the presence of the new binary `cogctl` that we inject. I think that's used internally in a setup script, but we might be able to rework things so that's no longer required. Historically, we've also had the `cogctl` binary inside the `cog` image as a way to distribute it an make it easier for people to run (since installing it natively used to require having an Erlang VM in place, the setup of which isn't necessarily trivial). Now, however, `cogctl` is written in Python and distributed as a stand-alone binary, so that motivating force is pretty much gone now.

I think simplifying the image build process is probably worthwhile now. It makes it easier to have external automated services (like Docker Hub!) run builds, and makes it easier for community members to also build. I'll close this PR and create a new issue tracking that idea. 

Thank you so much for taking a run at this, though, @nmohoric ! It's much appreciated 👍

PR #1389, opened 2017-04-28
Change Cog-Book URLs in README to unsecure

### LINKED ISSUE #1386: The "Cog Book" shows a security warning in Chrome and Firefox
Clicking through to the Cog Book displays an insecure connection warning for me in both Chrome and Firefox. It seems that this URL is forwarding to servers provided by github, and the certificate provided is only valid for github and related domains.

URL affected: https://cog-book.operable.io/index.html (anything under cog-book.operable.io)
Chrome version: Version 57.0.2987.133 (64-bit)
Firefox version: 52.0.2 (32-bit)
Error from Firefox:
```
cog-book.operable.io uses an invalid security certificate. The certificate is only valid for the following names:
www.github.com, *.github.com, github.com, *.github.io, github.io, *.githubusercontent.com, githubusercontent.com

Error code: SSL_ERROR_BAD_CERT_DOMAIN
```

I've been seeing this for at least the last week or so. I'm not sure if this is the right place to report this, but I'm surprised it's been this way for so long, so thought I should let someone know.

ohaiwalt:
This prevents the security warning, and fixes #1386 until a certificate solution is implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@ Cog brings the power of the command line to the place you collaborate with your 
 
 Powerful access control means you can collaborate around even the most sensitive tasks with confidence. A focus on extensibility and adaptability means that you can respond quickly to the unexpected, without your team losing visibility.
 
-* [Installation Guide](https://cog-book.operable.io/#_installation_guide) - You want Cog, now go here to start installing it.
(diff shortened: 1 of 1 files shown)


ohaiwalt:
This was a README change, and the build failed :1st_place_medal:

PR #1408, opened 2017-12-14
Temporarily point warehouse to static site

nmohoric:
This will allow us to temporarily point the warehouse at this gh-pages static site until we can figure out what to do with the warehouse.


nmohoric:
Sadly this doesn't work correctly, they're in the wrong format. Will continue working on it.

nmohoric:
Finally got around to checking into this, it was actually just a problem with the exact two bundles i was trying to install. Going to merge this and cut a patch release unless anyone has sees a problem with that.

jesselang:
LGTM. 😃

PR #1413, opened 2018-04-04
Allow overriding the reply_broadcast value when using threading in slack

nmohoric:
This allows you to turn reply broadcasting off if you'd prefer when using threading.
This means that messages will only go to the thread and not also to the channel.

It defaults to broadcasting ON, since that's what it was previously doing.


rebyn:
This needs integration tests added to [slack_threads_test.exs](https://github.com/operable/cog/blob/master/test/integration/slack_threads_test.exs).

PR #1414, opened 2018-05-12
Chat service POC

nmohoric:
I've taken what Mark started and just added some tests for it.
This adds a chat service endpoint so you can post message back to the user/channel (or a specific user/channel if desired).

Will add docs on its usage, but it's just a POST to /v1/services/chat/1.0.0/send_message  with a destination channel/user and a message. 

Right now if the user wants to respond back directly they'd need to find out if the COG_ROOM was "direct", if so set the destination as @${COG_CHAT_HANDLE}, otherwise use #${COG_ROOM}. In the future i'd like to update the relay to set a COG_INITIATOR or something which decides this for you.

### DIFF
--- a/lib/cog/command/service/chat.ex
+++ b/lib/cog/command/service/chat.ex
@@ -15,7 +15,7 @@ defmodule Cog.Command.Service.Chat do
     end
   end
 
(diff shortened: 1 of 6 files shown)


jesselang:
LGTM

